Issue #6, opened 2022-08-01
Improve Haskell interface


WhatisRT:
The interface that's currently provided is only a proof of concept, and is pretty bad in practice. We should have proper data types and a good general story for the interface.

PR #15, opened 2022-09-13
Various extensions to the model

### LINKED ISSUE #11: Make the model more modular
Currently, the actual model lives entirely in `Ledger.lagda`, which has become a pretty large file. This should be split up into modules for certain topics.

WhatisRT:
This PR adds some (abstract) cryptographic primitives, addresses, scripts and protocol parameters, as well as a proper UTXOW rule, and improves the Haskell interface. Also closes #11.


WhatisRT:
Turns out that this also closes #6.

Ali-Hill (approved):
It looks good to me. We will need to integrate the different changes to membership for finsets on the next merge.

PR #17, opened 2022-09-27
Alasdair/specifications

Ali-Hill:
Changed definitions of membership and subsets for FinSet and FinMap data types


WhatisRT:
This is deprecated in favor of #20

PR #18, opened 2022-10-03
Nix build for midnight and some nix improvements

teodanciu:


default.nix
```diff
@@ -1,76 +1,79 @@
...
+let
+  customAgda = (import (builtins.fetchTarball
     https://github.com/nixos/nixpkgs/tarball/a7ecde854aee5c4c7cd6177f54a99d2c1ff28a31) ({ # 21.11
       overlays = [ (import (fetchFromGitHub {
         repo = "agda";
         owner = "input-output-hk";
         rev = "d5ea03b96328f38741efef4535197076ff0e05d5";
         sha256 = "WiadZWZvPWcR49JkiXPiMKW3plRjBlR94wyg/aoEoG8=";
       })).overlay ];
-    } // (if pkgs.system == "aarch64-darwin" then { system = "x86_64-darwin"; } else {}))); let
```
teodanciu, line 11: I removed the nested `let` - it seemed more difficult to follow and require a few passing on the variables between the scopes

src/MidnightExample/Main.hs
```diff
@@ -1,4 +1,4 @@
-module MidnightExample.Main (module MidnightExample.Main, module MAlonzo.Code.MidnightExample.HSLedger) where
+module Main (module Main, module MAlonzo.Code.MidnightExample.HSLedger) where
```
teodanciu, line 1: I removed the `MidnightExample` prefix from the module name in order to make it consistent with the main module from `Ledger` and make it easier to extract this in a function. 
We could also go the other way, and change `HSLedgerTest` module to `Ledger.HSLedgerTest` - I just chose one option arbitrarily, but happy to do the other one too

default.nix
WhatisRT: Your version is definitely clearer than mine!

x.nix
```diff
@@ -0,0 +1,151 @@
+{ nixpkgs ? import <nixpkgs> {} }:
```
WhatisRT, line 1: Is this file a leftover?

default.nix
```diff
@@ -1,116 +1,150 @@
...
+    dir = "MidnightExample";
+    agdaLedgerPath = "MidnightExample/HSLedger.agda";
+    hsMainPath = "MidnightExample/Main.hs";
+  };
+
+  ledgerExecutableSpecMidnight = haskellPackages.callCabal2nix "Agda-ledger-executable-spec-midnight" "${ledgerHsSrcMidnight}" { };
+
+  docs = { dir }:
+    stdenv.mkDerivation {
+      pname = "Simple-utxo-ledger-${dir}-docs";
```
WhatisRT, line 115: That name is pretty old, I think this would be better
```suggestion
      pname = "${dir}-docs";
```

default.nix
```diff
@@ -1,116 +1,150 @@
...
+        mkdir -p $out
+        agda -c --ghc-dont-call-ghc --compile-dir $out ${agdaLedgerPath}
+        cp ${hsMainPath} $out
+        cp ${dir}/agda-ledger-executable-spec.cabal $out
+        # Append all the modules generated by MAlonzo to the cabal file
+        find $out/MAlonzo -name "*.hs" -print | sed "s#^$out/#        #;s#\.hs##;s#/#.#g" >> $out/agda-ledger-executable-spec.cabal
+      '';
+    };
+
+  ledgerHsSrc = hsSrc {
```
WhatisRT, line 97: There's a bunch of repetition here. What do you think about making a single function that accepts this set and generates all three derivations at once?

teodanciu: Indeed! Let's see how that looks like, trying it now

default.nix
```diff
@@ -1,116 +1,148 @@
...
-  agdaWithPkgs = agdaDeps.agda;
 
-  ledger-hs-src = stdenv.mkDerivation {
-    pname = "Agda-ledger-hs-src";
+  deps = [ agdaStdlib agdaFinset agdaStdlibMeta ];
+  agdaWithPkgs = customAgda.agda.withPackages { pkgs = deps; ghc = pkgs.ghc; };
+
+in
+rec {
+
```
WhatisRT, line 71: We need to make agda itself available as well
```suggestion
  agda = agdaWithPkgs;
```

PR #19, opened 2022-10-07
flakify repository to prepare for CI integration

disassembler:
This is preparation for cicero integration.


WhatisRT:
Can we disable the three errors where there is no builder available somehow? Or will there be a builder in the future?

Also, the darwin build fails because of this:
```MidnightExample-docs> error: path '/nix/store/011r8qs3ggg40mp58gmnhyyygnhyfybf-texlive-datetime2-irish-1.1' is not valid```
That error looks to me like garbage collection happened too quickly? `Ledger-docs` also uses this package, and it's there at that time.

**Comment by dermetfan** [comment]
https://github.com/input-output-hk/plutus/pull/4988#issuecomment-1335839787

**Comment by WhatisRT** [comment]
Looks like something is broken about the darwin builders more generally? This time it fails with
```
Agda-ledger-MidnightExample-hs-src> client_loop: send disconnect: Broken pipe
error: unexpected end-of-file
```

dermetfan (approved):

WhatisRT (changes requested):
Glad that the tests now all succeed! The only thing that I'd still want before this is merged is to not use nix flakes for building the code. I don't know enough about them to comfortably maintain that myself, and I also don't want to tell people to enable experimental features to work with this repo.

dermetfan:
@WhatisRT Would it suffice to add [flake-compat](https://github.com/edolstra/flake-compat) so that people can build without enabling flakes? We _could_ revert everything back to the pre-flake way and then import it back into the flake. Not sure that's worth it though as pretty much every other repo is now flakified which would make this one an exception.

WhatisRT:
The bigger issue for me is really the ability to maintain it myself. In my experience, whenever I need to ask for help with anything nix related, it takes forever to get something done.

By looking at that flake, I get the impression that most of the things I don't understand in there are related to CI, so maybe separating out CI related things into its own file and adding `flake-compat` might work for me as well. And I assume that if it doesn't then actually reverting to the old `default.nix` would be a small change to that.

jmchapman:
close this?

WhatisRT:
Yeah, since we now have a working CI I'll close this for now. If we ever want to add Cicero we can revisit this.

Issue #10, opened 2022-08-01
Inference with extensional set equality


WhatisRT:
Implicit arguments can't currently be inferred from extensional set equality arguments. I.e., if we pass an argument of type `s ≡ᵉ s'` to a function, we still have to pass `s` and `s'` explicitly, which is pretty annoying. One way to fix this is by redefining `_≡ᵉ_` as a `data` type, but this means that a bunch of things have to be rewritten. Maybe there is some better solution.

WhatisRT:
This also applies to extensional equality for finite maps

PR #20, opened 2022-10-10
Add axiomatic set theory

### LINKED ISSUE #12: Axiomatic interface for finite sets & maps
There is a bunch of complexity coming from implementation details of the set theory leaking elsewhere. Usually set theory is axiomatic, so we can probably state the axioms that we need and have a cleaner interface.

WhatisRT:
Still TODO:
- [x] Fix POV proof
- [x] Some cleanup

Closes #12 and #10


Soupstraw (approved):
Looks good to me!

src/Ledger/Utxo.lagda
```diff
@@ -200,7 +214,7 @@ For all $\var{env}\in\UTxOEnv$, $\var{utxo},\var{utxo'}\in\UTxO$, $\var{fee},\va
 pov :
 \end{code}
 \begin{code}[inline*]
-  utxo ∩ᵖ outs tx ≡ ∅ᵐ
+  txid tx ∉ map proj₁ (dom (utxo ˢ))
```
Soupstraw, line 217: It would be great if it was clearer what the `proj₁` applies to. Sometimes it gets something from an existential, but sometimes it's used to get the first element of a pair. I had to look deeper here to understand what the `proj₁` applies to.

Ali-Hill (approved):
Looks good to me. I like the axiomatic set theory.

src/Ledger/Utxo.lagda
WhatisRT: Let's merge the PR then and turn this into an issue :)

PR #22, opened 2022-10-24
Add rules related to governance

WhatisRT:


WhatisRT:
This PR doesn't intend to match the rules in the Shelley spec exactly, it takes some shortcuts and applies some refactorings that we wanted to do as part of the next ledger era, most notably moving PPUP out of UTXO into LEDGER.

Ali-Hill (approved):
Looks good to me. It also builds and type checks on my system.

PR #35, opened 2022-11-29
Prove POV and computational for LEDGERS

WhatisRT:


Ali-Hill (commented):
This all looks good to me. We will need to look at how best to merge multi-asset with the notion of HasCoin in this pull request but I don't think it will be too difficult. I will approve this once I have gotten it to type check.

Ali-Hill:
> This all looks good to me. We will need to look at how best to merge multi-asset with the notion of HasCoin in this pull request but I don't think it will be too difficult. I will approve this once I have gotten it to type check.

Ali-Hill (approved):
This now builds and type checks for me.

Ali-Hill (changes requested):
I have two changes: 
1. Ledger/PPUp/Properties.agda does not type check for me 
2. I didn't catch this originally as this file was not checked in the Everything.agda file

WhatisRT:
Good find! I changed notation at some point (from 1# + 1# to 2 via literal overloading) and those two weren't definitionally equal and I didn't adjust it everywhere. It's fixed now.

Ali-Hill (approved):

Ali-Hill:
Everything type-checks for me now. The only files that are not type-checked by everything are the ones that we fixed in the master branch already.

williamdemeo:
Everything type-checks for me except the file `src/Interface/Show.agda`.  There's an ambiguity. (I suspect you added a module called `Show` to stdlib-meta and perhaps we don't need `src/Interface/Show.agda` anymore...?)

Here's the error:
```
/formal-ledger-specifications/src/Interface/Show.agda:3,8-22
Ambiguous module name. The module name Interface.Show could refer
to any of the following files:
  /nix/store/qnbpy249pdvqnr152sd2ffqw946qipyw-agda-stdlib-meta-0.1/Interface/Show.agda
  /formal-ledger-specifications/src/Interface/Show.agda
```

WhatisRT:
Yeah, this was just a leftover and already has been removed from `master`, and it'll disappear with the rebase.

PR #49, opened 2023-02-10
Mostly cosmetic

williamdemeo:
although fewer imports should be safer

### DIFF
--- a/src/MidnightExample/HSLedger.agda
+++ b/src/MidnightExample/HSLedger.agda
@@ -1,18 +1,16 @@
 {-# OPTIONS --overlapping-instances #-}
 module MidnightExample.HSLedger where
 
-open import Prelude hiding (_++_)
+open import Data.Integer        using (ℤ)
+open import Data.String         using (_++_)
+open import Interface.DecEq     using (DecEq)
+open import Interface.Hashable  using (Hashable; hash; Hashable₁; Hashable₂)
+open import Prelude             using (ℕ; String; List; _$_; foldr; _×_; _,_; Maybe)
 
-open import Interface.Hashable
-open import Interface.DecEq
-
-open import Data.Integer hiding (show)
 import Data.Integer.Show as Z
+import Data.List as L
+import Data.Maybe as M
 import Data.Nat.Show as N
-import Data.List
-import Data.Maybe
-open import Data.String using (_++_)
-
 import MidnightExample.Types as F
 
 Hash = ℕ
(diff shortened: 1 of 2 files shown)


src/MidnightExample/Ledger.lagda
```diff
@@ -214,7 +215,7 @@ non-zero, the proofs are identical.
 private variable s s' : LedgerState
                  b : Block
 
-open import Algebra.Properties.AbelianGroup +-0-abelianGroup
+open import Algebra.Properties.AbelianGroup +-0-abelianGroup using (identityʳ-unique)
```
WhatisRT, line 218: It's been a while since I wrote this, is there a reason why this import isn't at the top? If not, I think that'd be a good addition here

src/MidnightExample/HSLedger.agda
```diff
@@ -1,18 +1,16 @@
 {-# OPTIONS --overlapping-instances #-}
 module MidnightExample.HSLedger where
 
-open import Prelude hiding (_++_)
+open import Data.Integer        using (ℤ)
+open import Data.String         using (_++_)
+open import Interface.DecEq     using (DecEq)
+open import Interface.Hashable  using (Hashable; hash; Hashable₁; Hashable₂)
+open import Prelude             using (ℕ; String; List; _$_; foldr; _×_; _,_; Maybe)
```
WhatisRT, line 8: I think that generally it's best to keep modifications on module imports (`using/hiding/renaming`) to a minimum. Maybe my policy is bad, but it is currently to only use them to avoid name clashes and to rename things to a name that makes more sense in the current context (like renaming `×` to `∧`). I view the former as a bad thing (maybe this needs a type class) and the latter as a good thing. But making all imported names explicit hides both of those under a bunch of noise, and I'm not sure the noise adds anything.

I'm happy to discuss this further if you disagree. We should also really make some notes on code style somewhere.

src/MidnightExample/HSLedger.agda
```diff
@@ -30,10 +28,11 @@ instance
...
   Hashable-× .hash (a , b) = hash (N.show (hash a) ++ "," ++ N.show (hash b))
 
-open import MidnightExample.Ledger Hash
+open import MidnightExample.Ledger Hash using (Point; Tx; inc; dec; Header; Block; LedgerState; LEDGER-step; addBlockHash)
+
+open import Foreign.Convertible using (Convertible; to; from)
 
-open import Foreign.Convertible
-open import Foreign.Haskell.Coerce
+-- open import Foreign.Haskell.Coerce -- (It seems we're not using this...?)
```
WhatisRT, line 35: In that case just go ahead and remove it. I don't think there's ever a reason not to remove an unused import.

WhatisRT (commented):
Do you want to merge this for the Midnight example, or do you also want to do the same for Ledger as part of this PR? In the former case, I think you can convert it from a draft to a regular PR.

src/MidnightExample/HSLedger.agda
williamdemeo: I think these are valid points.  I got in the habit of making explicit everything I'm using in every module, mainly for clarity and safety.  I think it's safer to name everything that's used in a module because, for example, you are being explicit about which `refl` you're using. I could imagine a case in which you use some term in a proof that inhabits a type from a different module than you thought, and consequently the proof is at best misleading, possibly just plain wrong, even though it type-checks.

I'll continue going through the `Ledger` module and explicitly importing stuff if only because it's a good exercise for me as I learn what's involved in each module as I go... but I won't make a pull request for those edits, unless you want to see them.

WhatisRT: I'm not sure that helps with safety, since you'd have to be aware of the entire import list while writing the module to be really sure what each name means. `M-.` is much faster and requires less context to do the same thing.

williamdemeo:
Got it. You don't want to be explicit about what each module uses.  Closing.

PR #65, opened 2023-03-15
william/setup-mods-squashed

williamdemeo:
This provides a slight improvement to setup instructions for end users.


williamdemeo:
Trying again... I want to preserve Andre's changes, which are only in the original PR.

PR #67, opened 2023-03-15
minor improvement to setup instructions

williamdemeo:
This is a cleaner (squashed) version of my earlier PR with slightly improved instructions for end users.


CONTRIBUTING.md
```diff
@@ -2,13 +2,13 @@
...
+We are currently aspiring to follow the [Agda standard library style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md) as much as reasonable. Since some of our code will be rendered into a PDF, the formatting of the PDF takes priority over formatting of the code, so deviations are to be expected.
 
 ## Setup with emacs
 
-We rely on a patched Agda and stdlib, which makes setup more difficult. You can also install it using `nix-env -iA agda -f .`, but this is a global install, which you may not want if you also have other Agda projects.
+We rely on standard Agda version 2.6.3 and a patched version of the Agda Standard Library; this makes setup more difficult. You can install the correct version of Agda and the stdlib using `nix-env -iA agda -f .`, but this is a global install which you may not want if you also have other Agda projects.
```
williamdemeo, line 9: Change "We rely on standard..." to "We use..."

CONTRIBUTING.md
```diff
@@ -2,32 +2,45 @@
...
-            agda2-program-name (concat my/ledger-agda-name "/bin/agda"))
-    (setq agda2-version      "2.6.2.2"
-          agda2-program-name "agda"))
-  (agda2-restart))
-(with-eval-after-load 'agda2-mode (define-key agda2-mode-map (kbd "C-c C-x C-t") 'my/toggle-ledger-agda))
-```
+-  Build `agda` and `agda-mode` binaries and put symlinks to them somewhere in your path (e.g., ~/IOHK/ledger-agda`):
+
+   ```
+   mkdir -p ~/IOHK
```
williamdemeo, line 16: @WhatisRT Don't we need to make sure the `~/IOHK` directory exists (or else create it) in order for the `nix-build -A agda -o ~/IOHK/ledger-agda` command to succeed?

CONTRIBUTING.md
```diff
@@ -2,32 +2,45 @@
...
+   nix-build -A agda -o ~/IOHK/ledger-agda
+   export PATH = ~/IOHK/ledger-agda/bin:$PATH
+   ```
+
+   *Note*. You need not have built/installed Agda prior to invoking this `nix-build` command (though it's okay if you did).
```
williamdemeo, line 21: maybe "okay if you have" (instead of "okay if you did")

CONTRIBUTING.md
```diff
@@ -50,3 +63,7 @@ For example:
 niv update nixpkgs -r 4e329926df7ee5fa49929a83d31ee7d541f8b45c
 niv update nixpkgs -v 21.11.337905.902d91def1e
 ```
+
+
+[Agda]: https://wiki.portal.chalmers.se/agda/pmwiki.php
```
williamdemeo, line 68: I just put the links at the bottom to make the plain text version of this file easier for us to read/edit.

CONTRIBUTING.md
```diff
@@ -50,3 +63,7 @@ For example:
...
+[Agda standard library style guide]: https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md
```
williamdemeo, line 69: What happened to the `nix-build -A agdaLedger` command?  What was that for?

Tbh, I don't yet fully grasp what the nix-build commands do in each case... but I'll have a look at the *.nix files and figure it out.  (Still learning nix/nixos so this is probably good exercise for me.)

williamdemeo (commented):
minor comments/questions/doubts

CONTRIBUTING.md
williamdemeo: Done.

CONTRIBUTING.md
williamdemeo: Also, I think `export PATH = ~/IOHK/ledger-agda/bin:$PATH`, or something like it, is necessary, yes?

CONTRIBUTING.md
williamdemeo: Fixed.

CONTRIBUTING.md
WhatisRT: That `export` shouldn't be necessary, especially since we point emacs to the path manually. It also doesn't do anything outside the shell it's currently run from.

For making sure the directory exists, if we're that explicit about it maybe we should make it clear that it's a 'put it wherever you like' situation. Something like `my/favourite/directory/ledger-agda`

CONTRIBUTING.md
WhatisRT: `nix-build -A X` looks into `default.nix` and builds derivation `X`. If you look at the bottom of ours, it has the two lines `ledger = specsDerivations ...` and `midnight = specsDerivations ...`. `specsDerivations` is defined as `{ ... }: let ... in { executableSpecSrc = ...; executableSpec = ...; docs = ...; }` so it returns a set with those three elements. Each of those exist for `ledger` and `midnight`, so there are six derivations in total. So you can do something like `nix-build -A ledger.docs` to build that specific derivation, but also `nix-build -A ledger` to build all three for ledger.

Additionally, above `specsDerivation` there are three other derivations, `agda`, `agda2` (that one is a hack for working on stdlib-meta) and `agdaLedger`. The two Agdas just build Agda with libraries and `agdaLedger` is the Agda package of this repo. Building that type checks all of our Agda code and generates a package that is used by the other derivations. If you just run `nix-build`, it'll build all derivations at the top level, so exactly those three. So you never need to run `nix-build -A agdaLedger`. In the past, some of the other derivations also used to be at the top level which was slow, so it was useful to single it out, but the two Agdas should always be cached.

CONTRIBUTING.md
WhatisRT: In particular, we don't want them in the PATH, since then they will collide with a global install

CONTRIBUTING.md
williamdemeo: Nice explanation.  Thank you.  I don't suppose we should add any of that information to `Contributing.md`, should we?  Or should we assume those who intend to contribute will know enough about nix that they can deduce all that from `default.nix`?

WhatisRT: I don't think we should try to be a documentation for nix (or the parts of it we use), but I also wouldn't assume that people know nix. We could link to the nix documentation, but that's not actually useful for beginners I think. I have no idea what's best here.

WhatisRT (approved):
LGTM

PR #72, opened 2023-03-20
Update the nixpkgs used for Agda & move them into nix/sources.nix

WhatisRT:
This should make CI a lot faster

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -36,13 +36,15 @@ let
   };
 
   deps = [ agdaStdlib agdaStdlibMeta ];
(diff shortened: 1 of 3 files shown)


williamdemeo:
All the `nix-build` commands on the main README.md still work fine.  I can't confirm the CI speedup, but have no reason to doubt this will be faster. 👍🏼

WhatisRT:
You can see the CI speedup in the Actions tab. A successful run on other branches takes a little over 1h, while this branch ran twice in ~30 and ~40 minutes respectively.

PR #73, opened 2023-03-20
Add new nix substituters

teodanciu:
to improve cache use and build time


WhatisRT:
I assume this is superseded by #82?

teodanciu:
Let's close it for now, because it doesn't do anything as things are at the moment. 
It would still be good to have a better cache, so I might come back to it if there will be a good way to populate this cache

PR #74, opened 2023-03-21
WIP: examples of token algebras

williamdemeo:
This PR will implement some examples of token algebras for the multi-asset ledger.


williamdemeo:
The output of `nix-build -A ledger`

```
❯ nix-build -A ledger                                    
copying path '/nix/store/ch13s2s8lndcfgqanr01342x31w02ba7-glibc-locales-2.33-123' from 'https://cache.nixos.org'...
copying path '/nix/store/0pbpfglkgv23h6vy68vw7f1fv2f8wxap-distribution-nixpkgs-1.6.2-data' from 'https://cache.nixos.org'...
copying path '/nix/store/cznzsql8cs7nv6g6gflf1724r7d4s9lz-gawk-5.1.1' from 'https://cache.nixos.org'...
copying path '/nix/store/dykv06vzcn128bmwi19fm640xkdp747f-findutils-4.8.0' from 'https://cache.nixos.org'...
copying path '/nix/store/dwg2bk188x6axq8xk7ikzc8w2v7gddnd-bzip2-1.0.6.0.2' from 'https://cache.nixos.org'...
copying path '/nix/store/p30hamqm7fn0i7p8fq81d0ijv0gg0b9g-gnumake-4.3' from 'https://cache.nixos.org'...
copying path '/nix/store/75qbaxxl99sjzp6jpr2y6263jhlpkw6k-ed-1.17' from 'https://cache.nixos.org'...
copying path '/nix/store/58yya3k69fkk5x0gfazw1h3qjf1plgii-diffutils-3.8' from 'https://cache.nixos.org'...
copying path '/nix/store/gapgxqfyny37b2i6xbcmdzb9vnqar6y0-gnutar-1.34' from 'https://cache.nixos.org'...
copying path '/nix/store/mbj7jqg5cgzsc8nml56n3x4ymq3gca5q-gnused-4.8' from 'https://cache.nixos.org'...
copying path '/nix/store/296s5ywm9hw8ia9hr9j80z47rv51v6y9-gzip-1.12' from 'https://cache.nixos.org'...
copying path '/nix/store/dmhq727zgz5zixq1vz22kqji5y1bgm5i-openssl-1.1.1o' from 'https://cache.nixos.org'...
copying path '/nix/store/fdjqirgff4vyv92vqa2b59fr1zppj45n-patchelf-0.13' from 'https://cache.nixos.org'...
copying path '/nix/store/ga29fi1zxjyir4znwmrkp9y94r8mncba-xz-5.2.5' from 'https://cache.nixos.org'...
building '/nix/store/xmrfmqwx8xvgggvnkc9hq80sczv726s9-formal-ledger-0.1.drv'...
unpacking sources
unpacking source archive /nix/store/2a8xi4abm4g03h8kil9ymy47vwsggklx-src
source root is src
patching sources
copying path '/nix/store/k51ba3zyi52q66yyvc4lba96ndc3fvx0-patch-2.7.6' from 'https://cache.nixos.org'...
configuring
no configure script, doing nothing
building
copying path '/nix/store/kwk9gwaly79cwkmv4msd8rj8hyva299m-bzip2-1.0.6.0.2-bin' from 'https://cache.nixos.org'...
Checking Everything (/build/src/Everything.agda).
 Checking Ledger.PDF (/build/src/Ledger/PDF.lagda).
copying path '/nix/store/9wvdg6ayndlmhan3ccz938l0k9hvgxg9-xz-5.2.5-bin' from 'https://cache.nixos.org'...
copying path '/nix/store/b6ak2wii6k6j7m606c67rwyng1m7whnx-stdenv-linux' from 'https://cache.nixos.org'...
copying path '/nix/store/pgfaql94nvwcs6xddyzhfb2yg8crmhlr-cabal2nix-2.18.0' from 'https://cache.nixos.org'...
  Checking Ledger.Utxo (/build/src/Ledger/Utxo.lagda).
```

(It hangs when checking `Utxo.lagda` 😞)

WhatisRT:
I didn't have time to give it a proper review yet, but here are two things I've spotted:

- The module parameters `PolicyID` etc. should be moved into the definition of `TransactionStructure`. The whole point of that record is to bundle all the types & functions that we need for talking about transactions. I also have the suspicion that those additional module parameters are what worsens the type-checking performance.
- There are a bunch of 'permanent comments' (things like explanations). Those should go into the LaTeX part instead of Agda comments.

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,502 @@
...
+    PQ : q ≡ 0ℕ ⊎ ¬(q ≡ 0ℕ)
+    PQ = dichlem
+
+    taut : (q ≡ 0ℕ) → (q ≡ 0ℕ)
+    taut = λ z → z
+
+    contr : ¬ (q ≡ 0ℕ) → (q ≡ 0ℕ)
+    contr np = ι-left-unique{AdaPolicy , AdaName} (proj₂ (AdaForAll I)) _≡_.refl
+
+    goal : q ≡ 0ℕ
```
WhatisRT, line 383: This can be simplified a lot by inlining & switching to `Dec`. Here's one option:
```
    goal : I hasHowMuchAda ≡ 0ℕ
    goal with I hasHowMuchAda ≟ 0ℕ
    ... | yes p = p
    ... | no _  = ι-left-unique{AdaPolicy , AdaName} (proj₂ (AdaForAll I)) _≡_.refl
```
Since the neutral element is also used multiple times, you could move that into a global `private` definition. Then you can even do
```
   IsMonoidMorphism.ε-homo mh with I hasHowMuchAda ≟ 0ℕ
   ... | yes p = p
   ... | no _  = ι-left-unique{AdaPolicy , AdaName} (proj₂ (AdaForAll I)) _≡_.refl
```

src/Ledger/TokenAlgebra.lagda
```diff
@@ -1,29 +1,43 @@
...
+  MemoryEstimate : Type
   MemoryEstimate = ℕ
 
   open CommutativeMonoid Value-CommutativeMonoid using (_≈_; ε)
        renaming (Carrier to Value; refl to reflᵛ; _∙_ to _+ᵛ_) public
 
   field coin     : Value → Coin
         inject   : Coin → Value
-        policies : Value → ℙ PolicyId
+        policies : Value → Pred PolicyID 0ℓ
```
WhatisRT, line 36: This won't work for the UTXOW rule, since we need decidable membership of `policies`.

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,502 @@
...
+
+_⊕_ : Op₂ (AssetID ⇀ Quantity)
+
+-- the summed relation
+proj₁ ((Ru , _) ⊕ (Rv , _)) = Ru ⊕₁ Rv
+
+-- proof that the summed relation is a left-unique relation
+proj₂ ((Ru , luu) ⊕ (Rv , luv)) {aid} {q} {q'} = Goal
+ where
+ Goal : (Ru ⊕₁ Rv) aid q → (Ru ⊕₁ Rv) aid q' → q ≡ q'
```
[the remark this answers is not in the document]
williamdemeo: Fixed.

src/Ledger/TokenAlgebra.lagda
```diff
@@ -32,4 +46,58 @@ record TokenAlgebra : Set₁ where
 
   sumᵛ : List Value → Value
   sumᵛ = foldr _+ᵛ_ (inject 0)
+-- Derived types (See Fig 3 of Mary spec [1].) --------------------------
+
+record TokenAlgebraRel {ℓ : Level} : Type (lsuc ℓ) where
```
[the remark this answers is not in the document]
williamdemeo: Okay, I'll try that.  I had previous found another way to get rid of the lifts... I can't remember right now why I had to revert back to using lifts.  When I try the above suggestion, I'll probably remember.

williamdemeo:
> I didn't have time to give it a proper review yet, but here are two things I've spotted:
> 
> * The module parameters `PolicyID` etc. should be moved into the definition of `TransactionStructure`. The whole point of that record is to bundle all the types & functions that we need for talking about transactions. I also have the suspicion that those additional module parameters are what worsens the type-checking performance.
> * There are a bunch of 'permanent comments' (things like explanations). Those should go into the LaTeX part instead of Agda comments.

Fixed.

src/Ledger/TokenAlgebra.lagda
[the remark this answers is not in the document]
williamdemeo: You're right, it works!  Thanks for the suggestion.

src/Ledger/TokenAlgebra/ValueRelation.lagda
williamdemeo: Yes, thank you... good.  (I must have been a bit tired when I wrote that proof.)

src/Ledger/TokenAlgebra.lagda
williamdemeo: Yes, well, `Pred` is indeed moving toward a more general, Agda standard library-ish direction.  It's true it doesn't presuppose decidability. But should we not view that a feature, not a bug, since then decidability must be explicitly postulated (e.g., with a module parameter) if and when (and only if and when) it is needed?

WhatisRT:
`ByteString` and `AdaName` should also be part of `TransactionStructure`. `GlobalConstants` are a well-defined thing in the spec, so we shouldn't put anything in there that isn't actually part of that.

As for faster type-checking, I'm all for improving that but that should go into a separate PR. All these additions make this PR more difficult to review.

src/Ledger/TokenAlgebra.lagda
WhatisRT: To be more specific, here's where `policies` will be used:
```
scriptsNeeded : UTxO → TxBody → ℙ ScriptHash
scriptsNeeded utxo txb =
  mapPartial ((λ { (inj₂ sh) → just sh ; _ → nothing }) ∘ payCred ∘ proj₁) ((utxo ˢ) ⟪$⟫ txins txb)
  ∪ mapPartial (λ { record { credential = inj₂ sh } → just sh ; _ → nothing }) (setFromList $ txvote txb)
  ∪ map policyIdToScriptHash $ policies (mint txb)
```
The only reason it's not there yet is that we don't have a function `policyIdToScriptHash` yet.

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,471 @@
...
+Ru ⊕₁ Rv = λ aid q → ∃[ qu ] (∃[ qv ] (Ru aid qu × Rv aid qv × (q ≡ qu + qv)))
+
+_⊕_ : Op₂ (AssetID ⇀ Quantity)
+proj₁ ((Ru , _) ⊕ (Rv , _)) = Ru ⊕₁ Rv
+proj₂ ((Ru , luu) ⊕ (Rv , luv)) {aid} {q} {q'} = Goal
+ where
+ Goal : (Ru ⊕₁ Rv) aid q → (Ru ⊕₁ Rv) aid q' → q ≡ q'
+ Goal (qu , qv , Ruqu , Rvqv , q≡quqv) (qu' , qv' , Ruqu' , Rvqv' , q'≡qu'qv') = begin
+   q           ≡⟨ q≡quqv ⟩
+   qu + qv    ≡⟨ trans (cong (λ b → qu + b) (luv Rvqv Rvqv')) (cong (λ a → a + qv') (luu Ruqu Ruqu')) ⟩
```
WhatisRT, line 145: You can use `cong₂` here instead

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,471 @@
...
+
+\subsection{The identity of the value monoid}
+
+We'll use \AgdaBound{ι} to denote the identity for the Value monoid.
+
+\begin{code}
+ι : REL AssetID Quantity 0ℓ
+ι _ q = q ≡ 0
+
+ι-left-unique : left-unique-rel ι
```
WhatisRT, line 51: It's probably a good idea to put all those into a hidden code block (\begin{code}[hide]) and in a `private` block. We won't be using those anywhere, and I don't think it's worth showing those proofs in the PDF.

default.nix
```diff
@@ -35,6 +35,18 @@ let
     buildInputs = [ agdaStdlib ];
   };
 
+  # agdaPrelude = customAgda.agdaPackages.mkDerivation {
```
WhatisRT, line 38: Please revert this change

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,471 @@
...
+
+R⇒ι-right : ∀{R} → R ⇒ ι-right R
+R⇒ι-right {R} {aid} {q} Rh = q , 0 , Rh , refl , sym (+-identityʳ q)
+
+\end{code}
+
+\subsection{Equivalence on the value monoid}
+
+\begin{code}
+-- the following probably belongs somewhere else
```
WhatisRT, line 108: Are you sure this isn't in the stdlib yet? If not, I'd move it to `Relation.Binary.Properties.Ext`. I'm using `Ext` to denote modules that contain things that should have been in the stdlib (and would like to merge upstream some time in the future).

WhatisRT (changes requested):

WhatisRT:
I did some investigation, and it seems to me that one major source of slowdown was simply the line `open import Tactic.Defaults` in `Ledger.Utxo`. I'd be highly surprised if restricting the imports to fewer names had any effect that's even measurable in any reasonable benchmarks. Name lookup should always be `O(length n)` even if literally all names up to the same length have been defined, so I don't think it's worth any effort (show me some data to convince me otherwise, agda has some `--profile` options).

williamdemeo:
> `ByteString` and `AdaName` should also be part of `TransactionStructure`. `GlobalConstants` are a well-defined thing in the spec, so we shouldn't put anything in there that isn't actually part of that.
> 
> As for faster type-checking, I'm all for improving that but that should go into a separate PR. All these additions make this PR more difficult to review.

Okay, will do.

williamdemeo:
> I did some investigation, and it seems to me that one major source of slowdown was simply the line `open import Tactic.Defaults` in `Ledger.Utxo`. I'd be highly surprised if restricting the imports to fewer names had any effect that's even measurable in any reasonable benchmarks. Name lookup should always be `O(length n)` even if literally all names up to the same length have been defined, so I don't think it's worth any effort (show me some data to convince me otherwise, agda has some `--profile` options).

Yes, however, what helped me to finally identify `Tactic.Defaults` as the import which causes the problem was a process of elimination, removing unnecessary imports one-by-one until I found the culprit.  (How did you figure out that `Tactic.Defaults` was the culprit?)

So, even if it doesn't noticeably effect performance, I think it's good hygiene to only import what we use.  Moreover, since (for now) we are relying on the Agda Standard Library style guide, if we only using a few things from a module, then we should name those things.

Finally, with fewer imports comes greater freedom to use more sensible names for new definitions and avoid primes and such.

Some of my improvements in this PR are aimed at adhering to the stdlib style guide.  (I should probably separate these into a separate PR, but that is more time consuming... nonetheless, since it would make this easier to review, I will split those improvements off into a separate PR.)

Here are a couple of principles from the style guide I've been gradually incorporating.

- "The list of imports should be declared in alphabetical order."
- "When using only a few items (i.e. < 5) from a module, it is a good practice to enumerate the items that will be used by declaring the import statement with the directive using. This makes the dependencies clearer"
- "If multiple import modifiers are used, then they should occur in the following order: `public`, `using`, `renaming`, and if `public` is used then the `using` and `renaming` modifiers should occur on a separate line. For example:
   ```agda
   open Monoid monoid public
     using (ε) renaming (_∙_ to _+_)
   ```
- "where blocks are preferred rather than the let construction."

williamdemeo:
This PR is superseded by #103 (closing).

PR #77, opened 2023-03-22
Add Info gov action & script support for voting

WhatisRT:


src/Ledger/Foreign/LedgerTypes.agda
```diff
@@ -85,6 +85,10 @@ record PParams : Set where
         poolDeposit   : Coin
         Emax          : Epoch
         pv            : Pair ℕ ℕ
+        ccTermLimit   : ℕ
+        govExpiration : ℕ
```
Soupstraw, line 89: What does this expiration apply to? It's not very clear from the name.

src/Ledger/Tally.lagda
```diff
@@ -46,32 +47,33 @@ private
...
     aid ∈ dom (s ˢ)
-    → (kh , role) ∈ roles ˢ
     ────────────────────────────────
-    Γ ⊢ s ⇀⦇ vote aid role kh v md ,TALLY'⦈ addVote s aid role kh v
+    Γ ⊢ s ⇀⦇ vote aid role cred v x ,TALLY'⦈ addVote s aid role cred v
 
-  TallyPropose : let open TallyEnv Γ in
-    Γ ⊢ s ⇀⦇ propose c addr a md ,TALLY'⦈ addAction s epoch (txid , length l) c addr a
+  TallyPropose : ∀ {x} → let open TallyEnv Γ; open PParams pparams using (govExpiration; govDeposit) in
+    c ≡ govDeposit
```
Soupstraw, line 74: This also needs a check for `expiresIn`

Soupstraw (approved):
Looks good!

src/Ledger/Foreign/LedgerTypes.agda
WhatisRT: It's how many epochs a governance action has before it expires. The name is from the CIP and there'll be some explanation in the prose later.

src/Ledger/Tally.lagda
WhatisRT: `expiresIn` is set to `current epoch + govExpiration` in `addAction`. So we don't need to check it, because the transaction doesn't get a say in what it is.

Soupstraw: Ah, I see

PR #80, opened 2023-03-23
Nix cleanup & improvements

WhatisRT:
This improves consistency of package names, provides a single `Agda` package that should work for all purposes we had separate packages for previously and renames the generated PDFs. The produced files were called `PDF.pdf`, this renames them to `cardano-ledger.pdf` and `midnight-example.pdf` respectively.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -98,12 +98,12 @@ rec {
         ];
         buildPhase = ''
           find ${dir} -name \*.lagda -exec agda --latex {} \;
-          cd latex && latexmk -xelatex ${dir}/${doc}.tex && cd ..
+          cd latex && latexmk -xelatex ${dir}/PDF.tex && cd ..
         '';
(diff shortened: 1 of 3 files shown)


lemastero (approved):
LGTM 👍

default.nix
```diff
@@ -41,6 +41,7 @@ let
 in
 rec {
 
+  agdaWithStdLibMeta = agdaWithPkgs deps;
```
teodanciu, line 44: I think it's better to make this dependency explicit,  and it's better to cache this one, since the other one (with ledger) is used externally and doesn't have any dependencies in this file , so the dependency to the cache is less obvious

teodanciu (approved):
Well we're a bit stuck with approving now, since I also added commits to it. 
I will approve anyway! 

approve

PR #83, opened 2023-04-03
Publish specs on github pages

teodanciu:
This PR will create a static html file listing the pdf files generated in the build.

A current limitation is that the pdfs need to have unique names, otherwise only one of them is kept. 
At the moment,  in master, this is what is happening, because both ledger and midnight pdfs are called PDF.pdf , however [this PR ](https://github.com/input-output-hk/formal-ledger-specifications/pull/80) addresses this, so, after it is merged,  the page will show to different pdfs called "cardano-ledger.pdf" and "midnight-example.pdf". 

An example of how this page looks like: https://bienpulenta.github.io/formal-ledger-specifications

In order for the github page to be created, we need to change the Settings in this repo, and set the deploy-source to branch "gh-pages", like in this example screenshot:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 name: Formal Ledger Specs
 on: [push]
 
+permissions:
+  contents: write
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


teodanciu:
Indeed, force-pushed to rebase

WhatisRT (approved):
LGTM!

PR #84, opened 2023-04-11
change notation to ⇀ and ≤ᵗ

williamdemeo:
+ rename _↛_ to _⇀_
+ rename _≥ᵗ_ to _≤ᵗ_

### DIFF
--- a/src/Ledger/Prelude.agda
+++ b/src/Ledger/Prelude.agda
@@ -59,7 +59,7 @@ abstract
   ≟-∅ = Decˡ.≟-∅
 
 open import Axiom.Set.Rel th hiding (_∣'_; _∣^'_) public
-open import Axiom.Set.Map th renaming (Map to _↛_) public
+open import Axiom.Set.Map th renaming (Map to _⇀_) public
 open import Axiom.Set.TotalMap th public
 open L.Decˡ hiding (_∈?_; ≟-∅) public
 open import Axiom.Set.Sum th public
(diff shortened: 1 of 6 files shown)


src/Ledger/Prelude.agda
```diff
@@ -85,5 +85,5 @@ module Properties where
   module _ {A : Set} ⦃ _ : DecEq A ⦄ where
     open Intersectionᵖ {A} ∈-sp public
 
-_ᶠᵐ : {A B : Set} → A ↛ B → FinMap A B
+_ᶠᵐ : {A B : Set} → A ⇀ B → FinMap A B  -- unicode: `\rightharpoonup`
```
WhatisRT, line 88: I don't think we should leave notes on how to type something in random comments

WhatisRT (changes requested):
This is nice, but there are a bunch of new places with maps in the Voltaire branch. I'd rather change the arrow once that's merged. So we can either leave this open until then, or you can just change the inequality symbol and we merge it now.

src/Ledger/Prelude.agda
williamdemeo: Okay, let's wait.

williamdemeo:
Okay, let's wait.

> This is nice, but there are a bunch of new places with maps in the Voltaire branch. I'd rather change the arrow once that's merged. So we can either leave this open until then, or you can just change the inequality symbol and we merge it now.

williamdemeo:
The change to harpoon notation for partial functions is already part of #105.  The leq notation is too minor for a PR and should be incorporated in a later cosmetic/style PR.

PR #97, opened 2023-04-26
Split `txgov` into two fields

### LINKED ISSUE #41: Conway transaction body
We already have all the features in the transaction body, but we should split the governance into two fields, one for proposals and one for voting. The reasoning behind this is that while the ordering of those two are relevant, the ordering of mixed things has no semantic significance.

This is the companion issue to https://github.com/input-output-hk/cardano-ledger/issues/3145

WhatisRT:
Closes #41

### DIFF
--- a/src/Ledger/Tally.lagda
+++ b/src/Ledger/Tally.lagda
@@ -49,7 +49,6 @@ private
            role : GovRole
            cred : Credential
            v : Vote
-           l : List GovProcedure
            c : Coin
            addr : RwdAddr
            a : GovAction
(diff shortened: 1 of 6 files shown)


src/Ledger/Tally.lagda
```diff
@@ -64,19 +63,21 @@ addAction s e aid c addr a = insert s aid record
   { votes = ∅ᵐ ; deposit = c ; returnAddr = addr ; expiresIn = e ; action = a }
 
 -- x is the anchor in those two cases, which we don't do anything with
-data _⊢_⇀⦇_,TALLY'⦈_ : TallyEnv → TallyState → GovProcedure → TallyState → Set where
-  TallyVote : ∀ {x} → let open TallyEnv Γ in
+data _⊢_⇀⦇_,TALLY'⦈_ : TallyEnv × ℕ → TallyState → GovVote ⊎ GovProposal → TallyState → Set where
```
Soupstraw, line 66: Could we add `k` to `TallyEnv`?

Soupstraw (approved):
LGTM

PR #98, opened 2023-04-27
Added autovoters

### LINKED ISSUE #91: Pre-defined DReps
We need two pre-defined DReps, who always count as having voted on every action:

- Abstain: abstains on every action
- No Confidence: counts as `No` on every action except `No Confidence`, where it counts as `Yes`

Soupstraw:
This PR adds the `Voter` datatype. A voter can be either a credential voter that votes using a credential (such as DReps), an abstainer that always abstains, or a no confidence voter that always votes no to everything except the NoConfidence action. The autovoter votes are added as soon as a new action is proposed and these autovoters can be delegated to just like you would delegate to DReps.

resolves #91

### DIFF
--- a/src/Ledger/Deleg.lagda
+++ b/src/Ledger/Deleg.lagda
@@ -6,11 +6,14 @@
 open import Ledger.Prelude
 open import Ledger.Epoch
 open import Ledger.Crypto
+import Ledger.PParams as PP
 
 module Ledger.Deleg
   (crypto : Crypto)
-  (Network : Set)
+  (TxId Network DocHash : Set)
   (epochStructure : EpochStructure)
+  (ppd : PP.PParamsDiff epochStructure)
+  (ppHashable : isHashableSet (PP.PParams epochStructure))
   ⦃ _ : DecEq Network ⦄
   where
 
@@ -18,6 +21,8 @@ open Crypto crypto
 
 open import Ledger.Address Network KeyHash ScriptHash
 open import Ledger.PParams epochStructure using (PParams)
+open import Ledger.GovernanceActions TxId Network DocHash epochStructure ppd ppHashable crypto
+  using (VDeleg)
 
 open EpochStructure epochStructure
 
@@ -29,7 +34,7 @@ record PoolParams : Set where
         deposit     : Coin
 
 data DCert : Set where
-  delegate   : Credential → Maybe Credential → Maybe Credential → Coin → DCert
+  delegate   : Credential → Maybe VDeleg → Maybe Credential → Coin → DCert
   -- ^ TODO change `nothing` to leaving things unchanged & figure out how to undelegate best
   regpool    : Credential → PoolParams → DCert
   retirepool : Credential → Epoch → DCert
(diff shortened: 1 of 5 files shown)


src/Ledger/Deleg.lagda
```diff
@@ -98,10 +109,10 @@ requiredDeposit pp nothing = 0
 \begin{code}
 data _⊢_⇀⦇_,DELEG⦈_ : DelegEnv → DState → DCert → DState → Set where
   DELEG-delegate :
-    d ≡ requiredDeposit pp mc ⊔ requiredDeposit pp mc'
+    d ≡ requiredVDelegDeposit pp mv ⊔ requiredDeposit pp mc
```
WhatisRT, line 112: Since we need to switch the semantics of `nothing` to 'do nothing' for the delegations, we also need to change how the deposits work. Instead, we charge a deposit if `c` isn't already in `vDelegs` or `sDelegs`. You don't have to do this as part of this PR, I just wanted to point it out since your changes here will be removed then.

src/Ledger/GovernanceActions.lagda
```diff
@@ -38,9 +38,9 @@ GovActionID : Set
 GovActionID = TxId × ℕ
 
 data GovRole : Set where
-  CC    : GovRole
-  DRep  : GovRole
-  SPO   : GovRole
+  CC        : GovRole
```
WhatisRT, line 41: Any reason for the change in indentation?

src/Ledger/Ratify.lagda
```diff
@@ -29,8 +29,10 @@ instance
 \begin{figure*}[h]
 \begin{code}
 record StakeDistrs : Set where
-  field poolStakeDistr  : Credential ↛ Coin
-        drepStakeDistr  : Credential ↛ Coin
+  field drepStakeDistr    : Credential ↛ Coin
```
WhatisRT, line 32: We should probably change `drepStakeDistr` to `VDeleg ↛ Coin`. We'll compute that by combining the `voteDelegs` with the per credential delegation, and that's the type it'll give us. That also means we don't need the `autoVoterStakeDist` special casing.

src/Ledger/Ratify.lagda
```diff
@@ -108,25 +110,34 @@ restrictedDists coins rank dists = dists -- record dists { drepStakeDistr = rest
...
 votedAbstainHashes = votedHashes Vote.abstain
 
-getStakeDist : GovRole → StakeDistrs → Credential ↛ Coin
+autoVoterStakeDist : StakeDistrs → Voter ↛ Coin
+autoVoterStakeDist record
+  { noConfidenceStake = ncs
+  ; abstainerStake = as
+  } = fromListᵐ $ (noConfidence , ncs) ∷ (abstainer , as) ∷ []
+
+getStakeDist : GovRole → StakeDistrs → Voter ↛ Coin
```
WhatisRT, line 128: I think this should return something of type `VDeleg ↛ Coin`. It's strange that there's a `GovRole` in there.

src/Ledger/Tally.lagda
```diff
@@ -24,8 +27,16 @@ open import Ledger.Address Network KeyHash ScriptHash
 \end{code}
 \begin{figure*}[h]
 \begin{code}
+data Voter : Set where
```
WhatisRT, line 30: This is a strange type. I think if we have a type like this at all, it should be isomorphic to `VDeleg × GovRole`.

Soupstraw: I was considering that, but `noConfidenceRep , CC` sounds a bit odd.

WhatisRT (approved):

Soupstraw (commented):
LGTM

PR #103, opened 2023-05-06
Example of a token algebra

williamdemeo:
-  A new directory called `TokenAlgebra` was added inside of which is the `ValueRelation.lagda` file with the `ValueRelation` module.  It's called `ValueRelation` because it's based on `REL`; we expect to add to this directory a `ValueSet.lagda` file/module with an example based on `Set`.

-  Here a partial function is constructed as an inhabitant of the `REL` type from the Agda Standard Library along with a proof of left-uniqueness (i.e., if `(a, b)` and `(a, b')` belong to the relation, then `b ≡ b'`).

-  A couple of modifications to the original `TokenAlgebra` module were necessary. Some of these were done in a separate new `TokenAlgebraPoly` type.

   - [X] `PolicyID` is now a parameter of the `TokenAlgebra` module instead of a field in the `TokenAlgebra` record type.
   - [X] `policies : Value → ℙ PolicyID` is changed to `policies : Value → Pred PolicyID 0ℓ`, but @WhatisRT says we need to change that back.

**ISSUES/TODO**

- [ ] It's not possible to keep the `TokenAlgebraPoly` type at level `Type₁` Therefore, the current implementation of `TransactionStructure`, which lives at level `Type₁`, must use the `TokenAlgebra` type based on `Set` (though we could simply change the universe level in the definition of `TransactionStructure` to make it compatible with `TokenAlgebraPoly` if desired).

- [ ] We need to figure out how to construct inhabitants of `Set` type so that we can make use of `policies : Value → ℙ PolicyID` instead of `policies : Value → Pred PolicyID 0ℓ` in the `TokenAlgebraPoly` type.

### DIFF
--- a/src/Ledger/TokenAlgebra.lagda
+++ b/src/Ledger/TokenAlgebra.lagda
@@ -1,19 +1,25 @@
 \section{Token Algebra}
+\label{sec:token-algebra}
+
 \begin{code}[hide]
 {-# OPTIONS --safe #-}
-module Ledger.TokenAlgebra where
 
-open import Ledger.Prelude
+open import Agda.Primitive  using() renaming (Set to Type)
 
-open import Algebra using (CommutativeMonoid)
-open import Algebra.Morphism
-open import Data.Integer
-open import Data.Nat.Properties using (+-0-commutativeMonoid)
+module Ledger.TokenAlgebra (PolicyId : Type) where
 
-record TokenAlgebra : Set₁ where
-  field PolicyId : Set
-        Value-CommutativeMonoid : CommutativeMonoid 0ℓ 0ℓ
+open import Ledger.Prelude hiding (Rel)
 
+open import Algebra              using (CommutativeMonoid)
+open import Algebra.Morphism     using (IsCommutativeMonoidMorphism-syntax)
+open import Data.Nat.Properties  using (+-0-commutativeMonoid)
+open import Relation.Binary      using (Rel)
+open import Relation.Unary       using (Pred)
+
+record TokenAlgebra : Type₁ where
+  field Value-CommutativeMonoid : CommutativeMonoid 0ℓ 0ℓ
+
+  MemoryEstimate : Type
   MemoryEstimate = ℕ
 
   open CommutativeMonoid Value-CommutativeMonoid using (_≈_; ε)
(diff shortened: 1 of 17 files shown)


src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,464 @@
...
+   - `Quantity = ℤ ⊎ {⊥}` and `Val aid = ⊥` for all but finitely many `aid : AssetID` OR
+   - `Quantity = Maybe ℤ` and `Val aid = nothing` for all but finitely many `aid : AssetID`
+   (The two options are equivalent.)
+3. Use the fact that the set theory, as Andre has defined it, is inherently finite.
+
+In my view the first option, with a finite domain, seems the easiest to express and work with.
+
+Here are two alternative approaches that might be better:
+
+---------------------
```
WhatisRT, line 375: Anything below here is very implementation-y. If you want to keep this, you could make an appendix or something for this.

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,464 @@
...
+\begin{code}
+open TokenAlgebraPoly
+open CommutativeMonoid renaming (_∙_ to _⋆_) hiding (trans ; sym)
+open Algebra
+
+module _
+  {AdaPolicy   : PolicyID}
+  {DecEqValue  : DecidableEquality (AssetID ⇀ Quantity) }
+  {Size        : AssetID ⇀ Quantity → ℕ}
+  {AdaForAll   : (R : AssetID ⇀ Quantity) → ∃[ q ] (proj₁ R) (AdaPolicy , AdaName) q}
```
WhatisRT, line 237: Note that `AdaForAll` implies `⊥`:
```
contr : ⊥
contr = proj₂ $ AdaForAll ((λ _ _ → ⊥) , λ where () ())
```
To not have a contradiction, you'd need to instead define a subset of relations that satisfy this property. However, everything in this file becomes way easier if you instead require all relations to be total. Then, the definition of `⊕₁` wouldn't require the two existentials for example. This also fixes the issue that there are tons of projections all over the place.

src/Ledger/TokenAlgebra/ValueRelation.lagda
```diff
@@ -0,0 +1,464 @@
...
+An inhabitant of `Value` is a map denoting a finite collection of quantities of assets.
+
+\begin{code}
+open TokenAlgebraPoly
+open CommutativeMonoid renaming (_∙_ to _⋆_) hiding (trans ; sym)
+open Algebra
+
+module _
+  {AdaPolicy   : PolicyID}
+  {DecEqValue  : DecidableEquality (AssetID ⇀ Quantity) }
```
WhatisRT, line 235: This is never inhabitable, even though I don't think that's provable. Think of the map that contains `1 Ada` if the Riemann hypothesis is true, and `0 Ada` otherwise. This is fixed by switching to a decidable set theory.

WhatisRT (changes requested):
In addition to the comments, I made some edits to show a bunch more simplifications. It was easier just to do them than to explain what to do.

src/Ledger/TokenAlgebra/ValueRelation.lagda
williamdemeo: I don't see why assuming a decidable set theory is better than assuming decidable equality here.  We won't ever construct an inhabitant/proof of `DecidableEquality (AssetID ⇀ Quantity)`.  That's why it's a module parameter; i.e., it's an assumption or "axiom."  In particular, it precludes having a map that contains 1 Ada if the Riemann hypothesis is actually a theorem and 0 if RH is false.

WhatisRT: It's an executable spec, and in particular we use that decidable equality in `consumed pp s tx ≡ produced pp s tx` in line 253 in `Ledger.Utxo` (or rather when we provide a `Computational` instance for that). Also, it's not true that this axiom precludes that. For example, here's a proof that all types have decidable equality using this axiom:
```
  Dec-all : {A : Set} → DecidableEquality A
  Dec-all {A} a b = Dmap
    (mk⇔ (λ h → helper (subst (λ M → proj₁ M ID 0) h (Prelude.refl , Prelude.refl , Prelude.refl)))
          (λ where refl → Prelude.refl)) (DecEqValue Amap Bmap)
    where open import Relation.Nullary.Decidable renaming (map to Dmap)
          ID : AssetID
          ID = AdaPolicy , AdaName

          genMap : A → AssetID ⇀ Quantity
          genMap X = (λ p x → p ≡ ID × x ≡ 0 × (a ≡ X))
                   , λ where (_ , refl , _) (_ , refl , _) → Prelude.refl

          Amap : AssetID ⇀ Quantity
          Amap = genMap a

          Bmap : AssetID ⇀ Quantity
          Bmap = genMap b

          helper : proj₁ Bmap ID 0 → a ≡ b
          helper (_ , _ , h) = h
```

src/Ledger/TokenAlgebra.lagda
```diff
@@ -32,4 +45,79 @@ record TokenAlgebra : Set₁ where
...
+
+In the formal ledger specification \AgdaBound{AssetID} is sometimes viewed as a direct sum type,
+the inhabitants of which belong to either \AgdaBound{AdaIDType} or the product
+\AgdaBound{PolicyId}~\AgdaBound{×}~\AgdaBound{AssetName}; if we were adhering to that point of view,
+then we would have defined
+\AgdaBound{AssetID}
+  = \AgdaBound{AdaIDType}~\AgdaBound{⊎}~(\AgdaBound{PolicyId}~\AgdaBound{×}~\AgdaBound{AssetName}).
+
+\begin{code}
+AssetName : Type
```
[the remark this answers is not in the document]
williamdemeo: Okay, sounds good.  Although, I think PolicyID does need to remain in TokenAlgebra.lagda.

src/Ledger/TokenAlgebra/ValueRelation.lagda
williamdemeo: I'll move that stuff to a separate module in a file under `TokenAlgebra`, called `DataStructures` or something, and leave it out of this PR, maybe working on it at some point in the future if/when we want to consider alternative implementations, optimizations, and the like.

PR #106, opened 2023-05-19
notation and formatting mods

williamdemeo:
- [X] make formalization more accessible by using more conventional notation like
   - [X] change `_↛_` to `_⇀_`
   - [X] change `_|^_` to `_↾_`
   - [X] change _|^'_ to _↾'_
- [X] make the style more even/consistent by adhering to the [agda standard library style guide](https://github.com/agda/agda-stdlib/blob/master/notes/style-guide.md)


src/Ledger/Utxo/Properties.lagda
```diff
@@ -28,7 +28,7 @@ open import Tactic.MonoidSolver
 
 open TransactionStructure txs
 
-open import Ledger.Crypto
+open import Ledger.Crypto using ()
```
WhatisRT, line 31: Seems like this is unused, so you might as well remove it.

WhatisRT (changes requested):
Thank you! I like the `↾` notation, great idea! The only thing I'd change here is that double spaces to align things should only be in places where they'll affect the LaTeX output. Otherwise, I think the alignments would become inconsistent over time.

WhatisRT:
Also, do you intend to do any further work on this PR? If not, you can mark it as ready.

williamdemeo:
> Thank you! I like the `↾` notation, great idea! The only thing I'd change here is that double spaces to align things should only be in places where they'll affect the LaTeX output. Otherwise, I think the alignments would become inconsistent over time.

Hmmm... I think having two different indentation conventions---one for `.agda` files and another for `.lagda` files---might be harder to maintain.  As a practical matter, I would prefer to not have to look at the file name or the type of code block to decide which indentation style to use.

williamdemeo:
I'll a rewrite of one complicated proof I was working on, and then rebase on the latest master and resolve the conflicts before marking it "ready for review."

WhatisRT:
I agree that having two different conventions for indentation is annoying, but they do have different meaning. The issue is that we may use the additional space to align all sorts of things that don't really fit the pattern. Your point about it being annoying to need to know where you are to format things also applies to all sorts of other things too - in the places where code is being displayed in the PDF, different formatting rules apply. Lines need to be shorter, we should be using less Agda specific syntax, etc. But I don't think all the other code should suffer from those stricter rules.

williamdemeo:
> I agree that having two different conventions for indentation is annoying, but they do have different meaning. The issue is that we may use the additional space to align all sorts of things that don't really fit the pattern. Your point about it being annoying to need to know where you are to format things also applies to all sorts of other things too - in the places where code is being displayed in the PDF, different formatting rules apply. Lines need to be shorter, we should be using less Agda specific syntax, etc. But I don't think all the other code should suffer from those stricter rules.

Fair enough.  I'll try to remove any/all indentation modifications I made to `.agda` files before marking "ready for review."

williamdemeo:
> > I agree that having two different conventions for indentation is annoying, but they do have different meaning. The issue is that we may use the additional space to align all sorts of things that don't really fit the pattern. Your point about it being annoying to need to know where you are to format things also applies to all sorts of other things too - in the places where code is being displayed in the PDF, different formatting rules apply. Lines need to be shorter, we should be using less Agda specific syntax, etc. But I don't think all the other code should suffer from those stricter rules.
> 
> Fair enough. I'll try to remove any/all indentation modifications I made to `.agda` files before marking "ready for review."

Although, if we later decide to document these, and change them to `.lagda` files that will be used to generate pdfs and html pages, then we will have to reinsert these extra spaces... so I wonder if I should just leave them in.  After all, it's just a single space! :)

williamdemeo:
Finally got this major rebase against master to give the desired result (so I'm just one commit ahead of master right now).

I hope we can merge this soon because this is the second time I've had to do a major rebase.  

However, if @WhatisRT wants to merge his PR first, no problem.  In that case, I'll rebase this one once more after his merge.  And hopefully @WhatisRT will approve this PR soon thereafter 😅

WhatisRT:
Thank you! I'll rebase & merge it for you, but I'll exclude the changes to `Ledger.Utxo.Properties` for now. It's not that I don't like them, but that file is already a major headache & @Soupstraw has some changes lined up that I want to prioritize. I'll put the unmerged changes back onto this branch then.

WhatisRT (approved):

PR #118, opened 2023-05-31
Treasury donations

### LINKED ISSUE #109: Treasury donations
We want the ability to donate money to the treasury. After some discussions, the easiest option seems to be to add a new field to the transaction body that contains the size of the donation & include that field in the transaction balancing equation. If we expose that field to plutus we can also do some extra logic with it, for example having some epoch where the unspent funds can only be moved back into the treasury.

Soupstraw:
This PR adds the donations field to TxBody and updates the `pov` proof to account for that. I also extracted some lemmas from the pov proof and moved them to the `DepositHelpers` module.

resolves #109

### DIFF
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -166,20 +166,21 @@ instance
 
       from' : F.TxBody → TxBody
       from' txb = let open F.TxBody txb in record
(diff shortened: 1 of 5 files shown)


WhatisRT (approved):
LGTM! I made some small cosmetic changes & removed a few things that un-did some previous changes, but nothing major.

PR #120, opened 2023-06-02
Fixed a mistake in the VDEL rule

Soupstraw:
The VDEL-deregdrep transition seemed to be adding the credential, not removing it.


WhatisRT (approved):
This will need to be changed again later to compute `dReps'` from `dReps`, but for now it's an improvement so I'll merge it.

PR #123, opened 2023-06-06
Add DRep expiration

### LINKED ISSUE #113: DRep expiration
DReps should expire if they didn't vote for a sufficient amount of time. Here's the current design:

Every time a DRep votes, we remember the number `currentEpoch + drepActivity`. Then, when ratifying, we only count DReps in the denominator whose epoch we remembered is lower than the current epoch.

There are two optional additions to the design that we could consider:
- allow to publish a `regdrep` cert with no deposit to refresh the number without voting
- increment the number for every epoch that had no governance actions to vote on

WhatisRT:
Closes #113


WhatisRT:
This was included in #127

PR #127, opened 2023-06-07
Many fixes to enactment semantics

### LINKED ISSUE #93: Fix details of proposal semantics
- [ ] We need to check for enactment in the order of proposal
- [ ] Enacting `NoConfidence` removes all proposed actions (Note: I still prefer delay, as other actions below)
- [ ] Enacting `NewCommittee`, `NewConstitution` and `TriggerHF` delays all other ratification
- [ ] Enacting `TriggerHF` and `ChangePParams` need to be delayed for consensus

WhatisRT:
This closes #93. I suggest reviewing this commit by commit, since each commit is dedicated to one change, but the overall PR is probably quite confusing with all changes at once.


src/Ledger/Deleg.lagda
```diff
@@ -133,41 +140,50 @@ data _⊢_⇀⦇_,POOL⦈_ : PoolEnv → PState → DCert → PState → Set whe
...
-    pp ⊢ st ⇀⦇ dCert ,CERT⦈ record st { vState = stᵛ' }
+    Γ ⊢ st ⇀⦇ dCert ,CERT⦈ record st { vState = stᵛ' }
 
   CERT-pool :
     pp ⊢ stᵖ ⇀⦇ dCert ,POOL⦈ stᵖ'
     ────────────────────────────────
-    pp ⊢ st ⇀⦇ dCert ,CERT⦈ record st { pState = stᵖ' }
+    ⟦ e , pp , vs ⟧ᶜ ⊢ st ⇀⦇ dCert ,CERT⦈ record st { pState = stᵖ' }
+
+data _⊢_⇀⦇_,CERTBASE⦈_ : CertEnv → CertState → ⊤ → CertState → Set where
```
Soupstraw, line 177: Where is this going to be used?

src/Ledger/Ratify.lagda
```diff
@@ -207,6 +219,27 @@ accepted Γ es@record { cc = cc ; pparams = record { votingThresholds = drepThre
...
+delayingAction NoConfidence         = true
+delayingAction (NewCommittee _ _ _) = true
+delayingAction (NewConstitution _)  = true
+delayingAction (TriggerHF _)        = true
+delayingAction (ChangePParams _ _)  = false
+delayingAction (TreasuryWdrl _)     = false
+delayingAction Info                 = false
+
+delayed : (a : GovAction) → NeedsHash a → EnactState → Bool → Set
+delayed a h es d = ¬ verifyPrev a h es ⊎ d ≡ true
```
Soupstraw, line 242:
```suggestion
delayed a h es true = ⊤
delayed a h es _ = ¬ verifyPrev a h es
```
Maybe this would look a bit nicer

src/Ledger/Tally.lagda
```diff
@@ -52,27 +56,32 @@ private
            c : Coin
            addr : RwdAddr
            a : GovAction
+           prev : NeedsHash a
 \end{code}
 \begin{code}
+modifyMatch : ∀ {a} {A : Set a} → (A → Bool) → (A → A) → List A → List A
```
Soupstraw, line 62: I think a function of type `∀ {a} {A : Set a} → (A → Maybe A) → List A → List A` would be more convenient.

Soupstraw (approved):
LGTM

src/Ledger/Ratify.lagda
WhatisRT: I actually prefer the current form, but maybe that's something to revisit once we're polishing the PDF

src/Ledger/Deleg.lagda
WhatisRT: Good point! This should be the base case of `CERT`, but the `SS⇒BS` transformer doesn't current support that.

WhatisRT:
I've added some comments for the things you've pointed out & merge it like that

PR #128, opened 2023-06-08
TokenAlgebra examples

williamdemeo:
-  Make new directory `src/Ledger/TokenAlgebra` containing two files so far:

   -  `ValueRelation.lagda` contains examples of token algebras which avoid the special set theory implemented in `src/Axiom/Set.lagda` and the submodules of `Axiom.Set`.

   -  `ValueSet.lagda` contains an (incompleted) example of a token algebra based on @WhatisRT's set theory formalization in `src/Axiom/Set.lagda` and submodules of `Axiom.Set`.

-  In `ValueRelation` a partial function is constructed as an inhabitant of the `REL` type from the Agda Standard Library along with a proof of left-uniqueness (i.e., if `(a, b)` and `(a, b')` belong to the relation, then `b ≡ b'`).

-  A couple of modifications to the original `TokenAlgebra` module were necessary. Some of these were done in a separate new `TokenAlgebraPoly` type. In particular,

   ```agda policies : Value → ℙ PolicyID ```

   was changed to

   ```agda policies : Value → Pred PolicyID 0ℓ ```

   but @WhatisRT noted that this won't work.

   We have commented out the `policies` field in the original `TokenAlgebra` record type, which is the one used in the example in `ValueSet.lagda`.

-  If we want to use the `REL` type of the standard library, then it's not possible to keep the `TokenAlgebraPoly` type at level `Type₁` Therefore, the current implementation of `TransactionStructure`, which lives at level `Type₁`, must use the `TokenAlgebra` type based on `Set` (though we could simply change the universe level in the definition of `TransactionStructure` to make it compatible with `TokenAlgebraPoly` if desired).

-  We have left some holes that need filling which we're having trouble with because of the quirkiness of the special set theory.  We're still learning how to construct inhabitants of the `Set` type and functions/maps over that type.
 
-  `PolicyID` is no longer a parameter of `TokenAlgebra` module because we've added it to the `TransactionStructure` record type.

### DIFF
--- a/src/Axiom/Set/Rel.agda
+++ b/src/Axiom/Set/Rel.agda
@@ -1,33 +1,33 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 {-# OPTIONS -v allTactics:100 #-}
 
-open import Agda.Primitive renaming (Set to Type)
+open import Agda.Primitive using(lzero) renaming (Set to Type)
 open import Axiom.Set
-import Relation.Binary.Reasoning.Setoid as SetoidReasoning
-import Function.Related.Propositional as R
 
 module Axiom.Set.Rel (th : Theory {lzero}) where
-open Theory th
-open import Axiom.Set.Properties th
 
+open import Axiom.Set.Properties th
+open import Data.List.Ext.Properties
+open import Data.Maybe.Base using () renaming (map to map?)
+open import Data.Product.Properties
+open import Data.These hiding (map)
+open import Interface.DecEq
 open import Prelude hiding (filter)
+open import Relation.Binary hiding (Rel)
+open import Relation.Nullary
+open import Relation.Unary using () renaming (Decidable to Dec₁)
+open import Tactic.AnyOf
+open import Tactic.Defaults
 
 import Data.Product
 import Data.Sum
-open import Data.These hiding (map)
-open import Data.List.Ext.Properties
-open import Data.Product.Properties
-open import Data.Maybe.Base using () renaming (map to map?)
-open import Interface.DecEq
-open import Relation.Unary using () renaming (Decidable to Dec₁)
-open import Relation.Nullary
-open import Relation.Binary hiding (Rel)
+import Function.Related.Propositional as R
 import Relation.Binary.PropositionalEquality as I
+import Relation.Binary.Reasoning.Setoid as SetoidReasoning
 
+open Theory th
 open Equivalence
 
-open import Tactic.AnyOf
-open import Tactic.Defaults
 
 -- Because of missing macro hygiene, we have to copy&paste this. https://github.com/agda/agda/issues/3819
 private macro
--- a/src/Axiom/Set/Sum.agda
+++ b/src/Axiom/Set/Sum.agda
@@ -1,6 +1,6 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 
-open import Agda.Primitive renaming (Set to Type)
+open import Agda.Primitive using (lzero) renaming (Set to Type)
 open import Axiom.Set
 open import Algebra using (CommutativeMonoid)
 
--- a/src/Axiom/Set/TotalMap.agda
+++ b/src/Axiom/Set/TotalMap.agda
@@ -1,19 +1,19 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 {-# OPTIONS -v allTactics:100 #-}
 
-open import Agda.Primitive renaming (Set to Type)
 open import Axiom.Set
 
 module Axiom.Set.TotalMap (th : Theory) where
 
-open import Prelude
-
-open Theory th
+open import Agda.Primitive using () renaming (Set to Type)
 open import Axiom.Set.Map th
 open import Axiom.Set.Properties th
 open import Axiom.Set.Rel th
+open import Prelude
 open import Tactic.AnyOf
 open import Tactic.Defaults
+
+open Theory th
 
 private variable A B : Type
 
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -1,9 +1,8 @@
 {-# OPTIONS --overlapping-instances #-}
 
-open import Agda.Primitive using () renaming (Set to Type)
-
 module Ledger.Foreign.HSLedger where
 
+open import Agda.Primitive          using () renaming (Set to Type)
 open import Algebra.Morphism        using ( IsCommutativeMonoidMorphism )
 open import Data.Nat                using ( _≤_ ; _≤ᵇ_)
 open import Data.Nat.Properties     using ( +-*-semiring ; <-isStrictTotalOrder )
@@ -94,7 +93,7 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
 
 open import Data.Nat
 open import Data.Nat.Properties using (+-0-commutativeMonoid; _≟_)
-open import Ledger.TokenAlgebra using (TokenAlgebra)
+open import Ledger.TokenAlgebra.Base using (TokenAlgebra)
 open import Ledger.Transaction
 
 
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -1,21 +1,22 @@
 {-# OPTIONS --overlapping-instances #-}
+
+open import Agda.Primitive using () renaming (Set to Type)
+
 module Ledger.Foreign.HSLedger where
 
+open import Algebra.Morphism        using ( IsCommutativeMonoidMorphism )
+open import Data.Nat                using ( _≤_ ; _≤ᵇ_)
+open import Data.Nat.Properties     using ( +-*-semiring ; <-isStrictTotalOrder )
+open import Foreign.Convertible
+open import Foreign.Haskell.Coerce  using (coerce)
+open import Ledger.Crypto           using (isHashableSet ; mkIsHashableSet ; PKKScheme ; Crypto)
+open import Ledger.Epoch            using (GlobalConstants ; EpochStructure ; ℕEpochStructure )
 open import Ledger.Prelude
 
 import Data.Maybe as M
 import Data.Rational as ℚ
-
-open import Data.Nat using (_≤_; _≤ᵇ_)
-open import Data.Nat.Properties using (+-*-semiring; <-isStrictTotalOrder)
-
-open import Foreign.Convertible
-open import Foreign.Haskell.Coerce
-
 import Ledger.Foreign.LedgerTypes as F
 
-open import Ledger.Crypto
-open import Ledger.Epoch
 
 open GlobalConstants
 HSGlobalConstants : GlobalConstants
@@ -93,27 +94,31 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
 
 open import Data.Nat
 open import Data.Nat.Properties using (+-0-commutativeMonoid; _≟_)
-open import Ledger.TokenAlgebra
+open import Ledger.TokenAlgebra using (TokenAlgebra)
 open import Ledger.Transaction
 
-coinTokenAlgebra : TokenAlgebra
-coinTokenAlgebra = record
-  { PolicyId                = ℕ
-  ; Value-CommutativeMonoid = +-0-commutativeMonoid
-  ; coin                    = id
-  ; inject                  = id
-  ; policies                = λ x → ∅
-  ; size                    = λ x → 1 -- there is only ada in this token algebra
-  ; property                = λ x → refl
-  ; coinIsMonoidMorphism    = record
-    { mn-homo = record
-      { sm-homo = record { ⟦⟧-cong = λ z → z ; ∙-homo  = λ x y → refl }
-      ; ε-homo  = refl
-      }
-    }
-  ; _≤ᵗ_                    = _≤_
-  ; DecEq-Value             = record { _≟_ = Data.Nat._≟_ }
-  }
+
+module _ where
+  open TokenAlgebra
+  open IsCommutativeMonoidMorphism
+  coinTokenAlgebra : TokenAlgebra
+  coinTokenAlgebra .Value-CommutativeMonoid  = +-0-commutativeMonoid
+  coinTokenAlgebra .coin = id
+  coinTokenAlgebra .inject = id
+  coinTokenAlgebra .size = λ _ → 1
+  coinTokenAlgebra ._≤ᵗ_ = _≤_
+  coinTokenAlgebra .AssetNameType = ℕ
+  coinTokenAlgebra .property = λ _ → refl
+  coinTokenAlgebra .coinIsMonoidMorphism .mn-homo =
+    -- record
+    --   { mn-homo =
+          record
+            { sm-homo = record { ⟦⟧-cong = λ z → z ; ∙-homo  = λ x y → refl }
+            ; ε-homo = refl
+            }
+
+  --    }
+  coinTokenAlgebra .DecEq-Value = record { _≟_ = Data.Nat._≟_ }
 
 module _ where
   open TransactionStructure
@@ -121,21 +126,24 @@ module _ where
   HSTransactionStructure : TransactionStructure
   HSTransactionStructure .Ix              = ℕ
   HSTransactionStructure .TxId            = ℕ
+  HSTransactionStructure .AuxiliaryData   = ⊤
+  HSTransactionStructure .PolicyIdType    = ℕ
+  HSTransactionStructure .AssetNameType   = ℕ
   HSTransactionStructure .epochStructure  = HSEpochStructure
   HSTransactionStructure .globalConstants = HSGlobalConstants
-  HSTransactionStructure .AuxiliaryData   = ⊤
   HSTransactionStructure .crypto          = HSCrypto
   HSTransactionStructure .adHashingScheme = isHashableSet-⊤
   HSTransactionStructure .ppHashingScheme = isHashableSelf PParams
   HSTransactionStructure .ppUpd           = record { UpdateT = ⊤ ; updateGroups = λ _ → ∅ ; applyUpdate = λ p _ → p }
   HSTransactionStructure .txidBytes       = id
   HSTransactionStructure .networkId       = tt
-  HSTransactionStructure .tokenAlgebra    = coinTokenAlgebra
   HSTransactionStructure .DecEq-TxId      = DecEq-ℕ
   HSTransactionStructure .DecEq-Ix        = DecEq-ℕ
   HSTransactionStructure .DecEq-Netw      = DecEq-⊤
   HSTransactionStructure .DecEq-UpdT      = DecEq-⊤
+  HSTransactionStructure .tokenAlgebra    = coinTokenAlgebra
   HSTransactionStructure .ss              = HSScriptStructure
+
 
 open import Ledger.Utxo HSTransactionStructure
 open import Ledger.Utxo.Properties HSTransactionStructure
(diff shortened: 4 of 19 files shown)


williamdemeo:
TODO: In TokenAlgebra record, change the field containing proof that coin is a monoid hom (from Value to Nat).

Change from the deprecated
 
```agda        
  coinIsMonoidMorphism  : coin Is Value-CommutativeMonoid -CommutativeMonoid⟶ +-0-commutativeMonoid
```

to the following newer way to express the monoid hom property:

```agda
  coinIsMonoidHomomorphism  : IsMonoidHomomorphism coin
```

This is done in my clone of`TokenAlgebra.lagda`, `ValueRelation.lagda`, and `ValueSet.lagda`, but I'm still debugging other modules that depend on TokenAlgebra so that they type-check with the upgraded syntax.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+
+An inhabitant of `Value` is a map denoting a finite collection of quantities of assets.
+
+\begin{code}
+open TokenAlgebra
+open CommutativeMonoid renaming (_∙_ to _⋆_) hiding (refl ; sym ; trans)
+open Algebra
+
+
+Value-TokenAlgebra :  (assetNameType  : Type                                       )
```
WhatisRT, line 80:
```suggestion
Value-TokenAlgebra :  (AssetName  : Type                                       )
```
Adding the suffix `Type` to a type is bad style. Think of `n : Nat` vs. `n : NatType`. I've also seen this in a bunch of other places in this PR.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+                      (isTotal        : ∀ x {m : (PolicyId × assetNameType) ⇀ Quantity} → x ∈ dom (m ˢ))
```
WhatisRT, line 81: This lets you prove `False` again, because you can instantiate `m` to the empty set. You should use `Axiom.Set.TotalMap`.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+                      (specialPolicy  : PolicyId                                   )
+                      (specialAsset   : assetNameType                              )
+                      (decEqVal       : DecEq ((PolicyId × assetNameType) ⇀ Quantity))
```
WhatisRT, line 84: We'll get in trouble with this one, because propositional equality isn't the correct one here. We probably should change `DecEq` in `TokenAlgebra` to decidability of the equivalence relation on the monoid.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+                      (sz             : (PolicyId × assetNameType) ⇀ Quantity → ℕ  )
```
WhatisRT, line 85: Is this the `size` function? If so, you should name it like that.

WhatisRT (changes requested):
A few general notes:
- There are a bunch of changes in here that undo things from cleanup commits. Please review the diff and undo them.
- I think we need to discuss the policy of structuring imports. I don't actually like having all of them just alphabetically, I prefer grouping them thematically. Specifically, things like having `Prelude` always on its own, same with `Agda.Primitive`.
- I'm not sure why you moved the definition of a token algebra out of `Ledger.TokenAlgebra`, but I think you should revert that. We also don't need to reexport instances of token algebras from there, since only the Haskell interface cares about those. So there's no need for `Ledger.TokenAlgebra.Base`.

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: Yes, I went back and forth with this one... I settled on `assetNameType` because I thought that, in this instance especially, it would help to emphasize that this is a type and not an asset name; i.e., it's not an inhabitant of the asset name type, it *is* the asset name type!

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: Ah, okay, good catch!

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: Yeah, I wasn't sure what we actually needed to assume was decidable.  Presumably, we want to assume as little as possible.  I'll try to follow your advice here and change `DecEq` in `TokenAlgebra` to decidability of the equivalence relation on the monoid.  Thanks. 👍🏼

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: okay, will do. 👍🏼

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+    ; coinIsMonoidMorphism = CoinMM
+    ; DecEq-Value = decEqVal
+    }
+  where
+
+  AssetId : Type
+  AssetId = PolicyId × assetNameType
+
+  sp∈ : spec-∈ AssetId
+  sp∈ {X} = {!!}
```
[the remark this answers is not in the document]
williamdemeo: Not sure I understand this yet, but let me give it some thought.  I'll ask again if I can't figure it out.  (I'll try to lower my pain threshold. :)

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,249 @@
...
+    IsEquivalence.refl ≋-isEquivalence = refl
+    IsEquivalence.sym ≋-isEquivalence {x} {y} x≋y {a} = sym x≋y
+    IsEquivalence.trans ≋-isEquivalence x y = trans x y
+
+    _⊕_ : Op₂ (AssetId ⇀ Quantity)
+    u ⊕ v = m , lu
+      where
+
+      m : Set'(AssetId × Quantity)
+      m = {!!}
```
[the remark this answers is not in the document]
williamdemeo: Okay, I'll try.

williamdemeo:
> A few general notes:
> 
> * There are a bunch of changes in here that undo things from cleanup commits. Please review the diff and undo them.

Okay, sorry about that.  It was kind of a nightmare trying to resolve conflicts this time since this PR got so far behind master.  I tried to be careful, but I guess I wasn't careful enough.  Thanks for catching this!

> * I think we need to discuss the policy of structuring imports. I don't actually like having all of them just alphabetically, I prefer grouping them thematically. Specifically, things like having `Prelude` always on its own, same with `Agda.Primitive`.

I respectfully disagree.  I could appeal to authority and say "that's part of the standard library style guide" ...but instead I'd like to suggest that the reason alphabetical imports made it into the style guide is probably because this makes it much easier for people less familiar with the library to find a given import within a long list of them.

Would you consider adopting the standard that, in general, they should be *roughly* alphabetically, with special exceptions allowed?  e.g., `Prelude` and `Agda.Primitive` could always appear on their own, either at the top or bottom of the import list.


> * I'm not sure why you moved the definition of a token algebra out of `Ledger.TokenAlgebra`, but I think you should revert that. We also don't need to reexport instances of token algebras from there, since only the Haskell interface cares about those. So there's no need for `Ledger.TokenAlgebra.Base`.

I think I needed to do this to get the code to type-check even with holes remaining in `Ledger/TokenAlgebra/ValueSet.lagda` without turning off `--safe` flags in too many other modules.

Also, I was trying to find a way to debug the error "`ValueRelation.agdai` doesn't exist" that I get from github actions.  How do I make sure `Everything.agda` causes a new module to be type-checked even before the new module is used elsewhere in the library?  (This is why I wanted a more standard `Everything.agda` file... so I don't have to remember how our unconventional `Everything.agda` file works. 😞 )

williamdemeo:
By the way, I didn't rearrange the imports just for fun, or to annoy you.  It was because I was getting a number of "ambiguous definition" errors and, because of the way the imports were listed, and because most of them did not provide the list of things they were importing, it was more difficult/time-consuming to resolve the ambiguity and eliminate duplicate imports than I think it should have been.  With the way I've re-arranged the imports and provided more explicit lists of things they're importing, I think it's significantly easier to resolve such errors... but maybe it's just me...?

src/Axiom/Set/Rel.agda
```diff
@@ -1,33 +1,33 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 {-# OPTIONS -v allTactics:100 #-}
 
-open import Agda.Primitive renaming (Set to Type)
+open import Agda.Primitive using(lzero) renaming (Set to Type)
```
williamdemeo, line 4: (name clash avoidance)

src/Axiom/Set/Rel.agda
```diff
@@ -1,33 +1,33 @@
...
 open import Axiom.Set
-import Relation.Binary.Reasoning.Setoid as SetoidReasoning
-import Function.Related.Propositional as R
 
 module Axiom.Set.Rel (th : Theory {lzero}) where
-open Theory th
-open import Axiom.Set.Properties th
 
-open import Prelude hiding (filter)
```
williamdemeo, line 13: Revert ∵ appeasement

src/Axiom/Set/Sum.agda
```diff
@@ -1,6 +1,6 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 
-open import Agda.Primitive renaming (Set to Type)
+open import Agda.Primitive using (lzero) renaming (Set to Type)
```
williamdemeo, line 3: (name clash avoidance)

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,20 +1,20 @@
 {-# OPTIONS --safe --no-import-sorts #-}
 {-# OPTIONS -v allTactics:100 #-}
 
-open import Agda.Primitive renaming (Set to Type)
 open import Axiom.Set
 
 module Axiom.Set.TotalMap (th : Theory) where
 
-open import Prelude
```
williamdemeo, line 9: Revert ∵ appeasement

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -1,21 +1,21 @@
...
 
+open import Agda.Primitive          using () renaming (Set to Type)
+open import Algebra.Morphism        using ( IsCommutativeMonoidMorphism )
+open import Data.Nat                using ( _≤_ ; _≤ᵇ_)
+open import Data.Nat.Properties     using ( +-*-semiring ; <-isStrictTotalOrder )
+open import Foreign.Convertible
+open import Foreign.Haskell.Coerce  using (coerce)
+open import Ledger.Crypto           using (isHashableSet ; mkIsHashableSet ; PKKScheme ; Crypto)
+open import Ledger.Epoch            using (GlobalConstants ; EpochStructure ; ℕEpochStructure )
 open import Ledger.Prelude
```
williamdemeo, line 13: Revert ∵ appeasement

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -93,50 +93,57 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
...
+  coinTokenAlgebra : TokenAlgebra
+  coinTokenAlgebra .Value-CommutativeMonoid  = +-0-commutativeMonoid
+  coinTokenAlgebra .coin = id
+  coinTokenAlgebra .inject = id
+  coinTokenAlgebra .size = λ _ → 1
+  coinTokenAlgebra ._≤ᵗ_ = _≤_
+  coinTokenAlgebra .AssetNameType = ℕ
+  coinTokenAlgebra .property = λ _ → refl
+  coinTokenAlgebra .coinIsMonoidMorphism .mn-homo =
+    -- record
```
williamdemeo, line 112: We should remove the unnecessary parts (which were accidentally left commented out instead of removed).

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -93,50 +93,57 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
 
 open import Data.Nat
 open import Data.Nat.Properties using (+-0-commutativeMonoid; _≟_)
-open import Ledger.TokenAlgebra
+open import Ledger.TokenAlgebra.Base using (TokenAlgebra)
```
williamdemeo, line 96: We should revert the reorg that moves TokenAlgebra definition to `Ledger.TokenAlgebra.Base`, if possible, or at least do it after holes in `TokenAlgebra/ValueSet.lagda` are all filled.

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -93,50 +93,57 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
...
-    }
-  ; _≤ᵗ_                    = _≤_
-  ; DecEq-Value             = record { _≟_ = Data.Nat._≟_ }
-  }
+
+module _ where
+  open TokenAlgebra
+  open IsCommutativeMonoidMorphism
+  coinTokenAlgebra : TokenAlgebra
+  coinTokenAlgebra .Value-CommutativeMonoid  = +-0-commutativeMonoid
```
williamdemeo, line 104: make the style consistent with the definition below of `HSTransactionStructure`.

(By the way, why is it okay to use "Structure" as a suffix, but not "Type"?)

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -93,50 +93,57 @@ HSScriptStructure = record { p1s = HSP1ScriptStructure ; ps = HSP2ScriptStructur
...
 
   HSTransactionStructure : TransactionStructure
   HSTransactionStructure .Ix              = ℕ
   HSTransactionStructure .TxId            = ℕ
+  HSTransactionStructure .AuxiliaryData   = ⊤
+  HSTransactionStructure .PolicyIdType    = ℕ
+  HSTransactionStructure .AssetNameType   = ℕ
   HSTransactionStructure .epochStructure  = HSEpochStructure
   HSTransactionStructure .globalConstants = HSGlobalConstants
-  HSTransactionStructure .AuxiliaryData   = ⊤
```
williamdemeo, line 126: This makes the order of the fields consistent with the definition of the TransactionStructure type.

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -166,20 +173,20 @@ instance
...
-        ; txvldt     = coerce txvldt
-        ; txwdrls    = ∅ᵐ
-        ; txup       = nothing
-        ; txADhash   = nothing
-        ; netwrk     = nothing
-        ; txsize     = txsize
-        ; txid       = txid
-        ; txvote     = []
-        ; txprop     = []
+        { txins    = from ⦃ Convertible-FinSet ⦃ Coercible⇒Convertible ⦄ ⦄ txins
```
williamdemeo, line 176: This change was because this is a `.agda` file, not a `.lagda` file and I think @WhatisRT prefers to not have extra indents in `.agda` files (although think it would be simpler, and would prefer, if we could use one indentation style throughout---namely, a style that looks good in both emacs and the generated pdf).

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -223,7 +230,7 @@ instance
         ; maxTxSize        = maxTxSize
         ; maxHeaderSize    = maxHeaderSize
         ; maxValSize       = maxValSize
-        ; minUTxOValue     = minUTxOValue
+        ; minUtxOValue     = minUtxOValue
```
williamdemeo, line 233: I think, for consistency, we are changing all appearances of "UTxO" in a variable name to `Utxo` (but I may be wrong about this).

src/Ledger/Foreign/LedgerTypes.agda
```diff
@@ -81,7 +81,7 @@ record PParams : Set where
         maxTxSize        : ℕ
         maxHeaderSize    : ℕ
         maxValSize       : ℕ
-        minUTxOValue     : Coin
+        minUtxOValue     : Coin
```
williamdemeo, line 84: I think, for consistency, we are changing all appearances of "UTxO" in a variable name to Utxo (but I may be wrong about this).

src/Ledger/PParams.lagda
```diff
@@ -32,20 +32,20 @@ data PParamGroup : Set where
 
 record PParams : Set where
   field -- Network group
-        maxBlockSize      : ℕ
-        maxTxSize         : ℕ
-        maxHeaderSize     : ℕ
-        maxValSize        : ℕ
-        pv                : ProtVer -- retired, keep for now
+        maxBlockSize     : ℕ
```
williamdemeo, line 35: (same comment as above)
This change was because this is a .agda file, not a .lagda file and I think @WhatisRT prefers to not have extra indents in .agda files (although think it would be simpler, and would prefer, if we could use one indentation style throughout---namely, a style that looks good in both emacs and the generated pdf).

src/Ledger/PParams.lagda
```diff
@@ -32,20 +32,20 @@ data PParamGroup : Set where
...
+        pv               : ProtVer -- retired, keep for now
 
         -- Economic group
-        a                 : ℕ
-        b                 : ℕ
-        minUTxOValue      : Coin
-        poolDeposit       : Coin
+        a                : ℕ
+        b                : ℕ
+        minUtxOValue     : Coin
```
williamdemeo, line 44: (Same comment as above)
I think, for consistency, we are changing all appearances of "UTxO" in a variable name to Utxo (but I may be wrong about this).

williamdemeo (commented):
I don't see any instances of cleanups being reverted.  What am I missing?

I do see that I reverted some extra indentation in `.agda` files, since I thought that's what @WhatisRT wanted.  I would be glad to put the extra indentation back in, as I prefer it that way, for consistency with our indentation style in `.lagda` files.

williamdemeo:
@WhatisRT writes "There are a bunch of changes in here that undo things from cleanup commits. Please review the diff and undo them."

Could you please point to at least one of these?  I can't seem to find any.

(If you're referring to the removal of some extra indentation, that was intentional, but let me know if you want me to revert to extra indentation.)

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: Btw, `TransactionStructure` has a similar suffix.  Is this also bad style? ...or not *as* bad? 😄

src/Ledger/Foreign/HSLedger.agda
WhatisRT: Good point. I don't like the 'Structure' suffix either, if you have a better candidate let me know!

src/Ledger/Foreign/HSLedger.agda
WhatisRT: This is an instance where cleanup was undone: note that `txdonation` at the bottom is two characters longer so now the equality signs aren't aligned anymore.

src/Ledger/Foreign/HSLedger.agda
WhatisRT: This is another instance where cleanup was undone: I renamed it to `UTxO` recently. I'm in principle OK with naming it `Utxo`, but note that this changes it to `UtxO` which is definitely bad.

src/Ledger/PParams.lagda
WhatisRT: More cleanup undoing: note that this is a literate agda file.

src/Ledger/Transaction.lagda
```diff
@@ -35,11 +35,15 @@ the transaction body are:
   \item The size and the hash of the serialized form of the transaction that was included in the block.
 \end{itemize}
 
+Finally, $\fun{txid}$ computes the transaction id of a given transaction.
```
WhatisRT, line 38: This also undoes cleanup: I recently removed this paragraph, since it's wrong (it used to be true).

WhatisRT:
> Would you consider adopting the standard that, in general, they should be roughly alphabetically, with special exceptions allowed? e.g., Prelude and Agda.Primitive could always appear on their own, either at the top or bottom of the import list.

Yes, that's what I meant by 'discuss the policy of structuring imports', i.e. I don't actually like strict alphabetic sorting.

> How do I make sure Everything.agda causes a new module to be type-checked even before the new module is used elsewhere in the library?

You can temporarily put it into `Everything.agda` 😉

> With the way I've re-arranged the imports and provided more explicit lists of things they're importing, I think it's significantly easier to resolve such errors... but maybe it's just me...?

I just use `C-c C-w`, i.e. `agda2-why-in-scope-maybe-toplevel`. You give it a name, and it tells you all the places that define/import it. No need to search through import lists.

williamdemeo:
> > Would you consider adopting the standard that, in general, they should be roughly alphabetically, with special exceptions allowed? e.g., Prelude and Agda.Primitive could always appear on their own, either at the top or bottom of the import list.
> 
> Yes, that's what I meant by 'discuss the policy of structuring imports', i.e. I don't actually like strict alphabetic sorting.
> 
> > How do I make sure Everything.agda causes a new module to be type-checked even before the new module is used elsewhere in the library?
> 
> You can temporarily put it into `Everything.agda` wink
> 

Okay, I'll go back to that temporary work-around.  I was doing it that way before, but it seemed a bit hacky, and I thought maybe there was a better way I wasn't aware of.

> > With the way I've re-arranged the imports and provided more explicit lists of things they're importing, I think it's significantly easier to resolve such errors... but maybe it's just me...?
> 
> I just use `C-c C-w`, i.e. `agda2-why-in-scope-maybe-toplevel`. You give it a name, and it tells you all the places that define/import it. No need to search through import lists.

Cool!  I didn't know about that command.  Thanks!

PR #132, opened 2023-06-27
Add well-formedness for protocol parameter updates

### LINKED ISSUE #129: Protocol parameter checks
Some protocol parameters need to be restricted to certain ranges. Here are some:

- `collateralPercent` must not be 0
- `coinsPerUtxoByte` must not be 0
- add more here

We should have a function that sanity checks PParam updates and only
allows those proposals that stay withing the allowed bounds.

WhatisRT:
Closes #129


Soupstraw (approved):
LGTM

PR #134, opened 2023-06-28
Governance thresholds

### LINKED ISSUE #126: Governance thresholds
The `votingThresholds` are currently stubbed out and need to be replaced with all the things we need in there.

In particular, we need to figure out the thresholds for
withdrawals. We could do a piecewise linear curve that interpolates a
given set of points for example.

WhatisRT:
Closes #126. For now, we don't do anything special for withdrawals since it's not in the CIP anymore, but we can add some logic to make the threshold dependent on the withdrawal amount later.

### DIFF
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -254,7 +254,12 @@ instance
         ; Emax              = Emax
         ; collateralPercent = 0
         ; pv                = coerce pv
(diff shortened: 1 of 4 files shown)


WhatisRT:
I always forget to adjust the Haskell interface when I change `PParams`. I'll fix that tomorrow.

Soupstraw (approved):
LGTM

PR #135, opened 2023-06-28
Total map infrastructure

williamdemeo:
+  adds functionality to TotalMap module analogous to definitions in the Map module but using record types.

+  adds new module, `TotalMapOn`, with dependent record types for maps that are total on a given set `X : Set A`.

+  includes lookup and update functions as well as some lemmas about them.

### DIFF
--- a/src/Axiom/Set/TotalMap.agda
+++ b/src/Axiom/Set/TotalMap.agda
@@ -19,58 +19,44 @@ private variable A B : Type
 
 
 -- defines a total map for a given set
-total-map : Rel A B → Type
-total-map R = ∀ {a} → a ∈ map proj₁ R
+total : Rel A B → Type
+total R = ∀ {a} → a ∈ dom R
 
 
 record TotalMap (A B : Type) : Type where
-  field  rel : Set (A × B)
-         lun : left-unique rel
-         tot : ∀{a : A} → a ∈ dom rel
+  field  rel              : Set (A × B)
+         left-unique-rel  : left-unique rel
+         total-rel        : total rel
 
-
-  MapReduct : Map A B
-  MapReduct = rel , lun
-
+  toMap : Map A B
+  toMap = rel , left-unique-rel
 
   lookup : A → B
-  lookup a = proj₁ ξ
-    where
-    ξ : ∃[ b ] (a , b) ∈ rel
-    ξ = to dom∈ tot
+  lookup _ = proj₁ (to dom∈ total-rel)
+
+  -- verify that lookup is what we expect
+  lookup∈rel : {a : A} → (a , lookup a) ∈ rel
+  lookup∈rel = proj₂ (to dom∈ total-rel)
+
+  -- this is useful for proving equalities involving lookup
+  rel⇒lookup : {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
+  rel⇒lookup ab∈rel = sym (left-unique-rel ab∈rel (proj₂ (to dom∈ total-rel)))
 
 
-  lookupCert : {a : A} → (a , lookup a) ∈ rel
-  lookupCert = proj₂ (to dom∈ tot)
+module Update {B : Type} ⦃ _ : DecEq A ⦄ where
 
-
-  lookupLemma : {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
-  lookupLemma {a} ab∈rel = sym (lun ab∈rel alu∈rel)
-    where
-    alu∈rel : (a , lookup a) ∈ rel
-    alu∈rel = proj₂ (to dom∈ tot)
-
-
-module Unionᵗᵐ {B : Type} ⦃ _ : DecEq A ⦄ where
-
-  -- updateFn:  Helper for defining update functions.
-  --            Given (a , b) and x y, if x ≡ a return b, otherwise return y.
-  updateFn : A × B → A → B → B
-  updateFn (a , b) x y with (x ≟ a)
-  ... | yes _ = b
-  ... | no _ = y
+  private
+    updateFn : A × B → A → B → B
+    updateFn (a , b) x y with (x ≟ a)
+    ... | yes  _ = b
+    ... | no   _ = y
 
   open TotalMap
 
-  mapWithKeyTotal : ∀{B'}{f : A → B → B'}{R : Rel A B} → total-map R → total-map (map (λ { (x , y) → x , f x y }) R)
-  mapWithKeyTotal {f = f}{R} tot {a} = ψ (ϕ tot)
-    where
-    ϕ : a ∈ dom R → Σ (Σ A (λ x₁ → B)) (λ a₁ → a ≡ proj₁ a₁ × a₁ ∈ R)
-    ϕ = from ∈-map
-    ψ : Σ (Σ A (λ _ → B)) (λ uv → a ≡ proj₁ uv × uv ∈ R) → a ∈ dom (map (λ {(x , y) → x , f x y}) R)
-    ψ (_ , refl , aRb) = ∈-map′ (∈-map′ aRb)
+  mapWithKey : {B' : Type} → (A → B → B') → TotalMap A B → TotalMap A B'
+  mapWithKey f tm .rel              = map (λ{(x , y) → x , f x y}) (rel tm)
+  mapWithKey _ tm .left-unique-rel  = mapWithKey-uniq (left-unique-rel tm)
+  mapWithKey _ tm .total-rel        = ∈-map′ (∈-map′ (proj₂ (to dom∈ (total-rel tm))))
 
   update : A → B → TotalMap A B → TotalMap A B
-  update a b tm .rel = map (λ { (x , y) → x , updateFn (a , b) x y}) (rel tm)
-  update _ _ tm .lun = mapWithKey-uniq (lun tm)
-  update a b tm .tot {a'} = mapWithKeyTotal (tot tm)
+  update a b = mapWithKey (updateFn (a , b))
--- a/src/Axiom/Set/TotalMap.agda
+++ b/src/Axiom/Set/TotalMap.agda
@@ -1,34 +1,76 @@
 {-# OPTIONS --safe --no-import-sorts #-}
-{-# OPTIONS -v allTactics:100 #-}
 
-open import Agda.Primitive renaming (Set to Type)
-open import Axiom.Set
+open import Axiom.Set using ( Theory )
 
 module Axiom.Set.TotalMap (th : Theory) where
 
-open import Prelude
+open import Prelude hiding (lookup)
 
-open Theory th
-open import Axiom.Set.Map th
-open import Axiom.Set.Properties th
-open import Axiom.Set.Rel th
-open import Tactic.AnyOf
-open import Tactic.Defaults
+open import Agda.Primitive              using () renaming (Set to Type)
+open import Axiom.Set.Map th            using (left-unique; Map ; mapWithKey-uniq)
+open import Axiom.Set.Rel th            using (Rel ; dom ; dom∈)
+open import Interface.DecEq             using (DecEq ; _≟_)
+open import Relation.Nullary.Decidable  using (yes ; no)
+
+open Theory th    using (_∈_ ; map ; Set ; ∈-map ; ∈-map′)
+open Equivalence  using (to ; from)
 
 private variable A B : Type
 
--- Because of missing macro hygiene, we have to copy&paste this. https://github.com/agda/agda/issues/3819
-private macro
-  ∈⇒P = anyOfⁿᵗ (quote ∈-filter⁻' ∷ quote ∈-∪⁻ ∷ quote ∈-map⁻' ∷ quote ∈-fromList⁻ ∷ [])
-  P⇒∈ = anyOfⁿᵗ (quote ∈-filter⁺' ∷ quote ∈-∪⁺ ∷ quote ∈-map⁺' ∷ quote ∈-fromList⁺ ∷ [])
-  ∈⇔P = anyOfⁿᵗ (quote ∈-filter⁻' ∷ quote ∈-∪⁻ ∷ quote ∈-map⁻' ∷ quote ∈-fromList⁻ ∷ quote ∈-filter⁺' ∷ quote ∈-∪⁺ ∷ quote ∈-map⁺' ∷ quote ∈-fromList⁺ ∷ [])
 
 -- defines a total map for a given set
 total-map : Rel A B → Type
 total-map R = ∀ {a} → a ∈ map proj₁ R
 
-TotalMap : Type → Type → Type
-TotalMap A B = Σ (Rel A B) (λ R → left-unique R × total-map R)
 
-lookupMapᵗ : TotalMap A B → A → B
-lookupMapᵗ (_ , _ , tot) a = proj₂ (proj₁ (∈⇔P (tot {a})))
+record TotalMap (A B : Type) : Type where
+  field  rel : Set (A × B)
+         lun : left-unique rel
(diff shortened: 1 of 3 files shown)


src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
 total-map R = ∀ {a} → a ∈ map proj₁ R
 
-TotalMap : Type → Type → Type
-TotalMap A B = Σ (Rel A B) (λ R → left-unique R × total-map R)
 
-lookupMapᵗ : TotalMap A B → A → B
-lookupMapᵗ (_ , _ , tot) a = proj₂ (proj₁ (∈⇔P (tot {a})))
+record TotalMap (A B : Type) : Type where
+  field  rel : Set (A × B)
+         lun : left-unique rel
```
WhatisRT, line 28: I don't like `lun` - if I read it somewhere out of context I have no idea what it means. Just name it `left-unique-rel` or something.

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+         tot : ∀{a : A} → a ∈ dom rel
```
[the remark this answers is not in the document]
WhatisRT: Same with the name here, just name it `total-rel`.

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+
+
+  MapReduct : Map A B
```
WhatisRT, line 32: I don't like that name either. Maybe just call it `toMap`?

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+  MapReduct = rel , lun
+
+
+  lookup : A → B
+  lookup a = proj₁ ξ
+    where
+    ξ : ∃[ b ] (a , b) ∈ rel
```
WhatisRT, line 39: Please inline this, there's no reason for a `where` block

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+    ξ = to dom∈ tot
+
+
+  lookupCert : {a : A} → (a , lookup a) ∈ rel
```
WhatisRT, line 43: This should probably be called `lookup∈rel`

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+  lookupCert = proj₂ (to dom∈ tot)
+
+
+  lookupLemma : {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
+  lookupLemma {a} ab∈rel = sym (lun ab∈rel alu∈rel)
+    where
+    alu∈rel : (a , lookup a) ∈ rel
```
WhatisRT, line 50: Please inline this

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+    where
+    ξ : ∃[ b ] (a , b) ∈ rel
+    ξ = to dom∈ tot
+
+
+  lookupCert : {a : A} → (a , lookup a) ∈ rel
+  lookupCert = proj₂ (to dom∈ tot)
+
+
+  lookupLemma : {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
```
WhatisRT, line 47: This should also be renamed to reflect its type

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+    where
+    alu∈rel : (a , lookup a) ∈ rel
+    alu∈rel = proj₂ (to dom∈ tot)
+
+
+module Unionᵗᵐ {B : Type} ⦃ _ : DecEq A ⦄ where
+
+  -- updateFn:  Helper for defining update functions.
+  --            Given (a , b) and x y, if x ≡ a return b, otherwise return y.
+  updateFn : A × B → A → B → B
```
WhatisRT, line 58: If it's just a helper function, put it in a `private` or `where` block

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+  updateFn (a , b) x y with (x ≟ a)
+  ... | yes _ = b
+  ... | no _ = y
```
WhatisRT, line 61:
```suggestion
  ... | no  _ = y
```

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+    where
+    ϕ : a ∈ dom R → Σ (Σ A (λ x₁ → B)) (λ a₁ → a ≡ proj₁ a₁ × a₁ ∈ R)
+    ϕ = from ∈-map
+    ψ : Σ (Σ A (λ _ → B)) (λ uv → a ≡ proj₁ uv × uv ∈ R) → a ∈ dom (map (λ {(x , y) → x , f x y}) R)
+    ψ (_ , refl , aRb) = ∈-map′ (∈-map′ aRb)
+
+  update : A → B → TotalMap A B → TotalMap A B
+  update a b tm .rel = map (λ { (x , y) → x , updateFn (a , b) x y}) (rel tm)
+  update _ _ tm .lun = mapWithKey-uniq (lun tm)
+  update a b tm .tot {a'} = mapWithKeyTotal (tot tm)
```
WhatisRT, line 76:
```suggestion
  update _ _ tm .tot = mapWithKeyTotal (tot tm)
```

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+  --            Given (a , b) and x y, if x ≡ a return b, otherwise return y.
+  updateFn : A × B → A → B → B
+  updateFn (a , b) x y with (x ≟ a)
+  ... | yes _ = b
+  ... | no _ = y
+
+  open TotalMap
+
+  mapWithKeyTotal : ∀{B'}{f : A → B → B'}{R : Rel A B} → total-map R → total-map (map (λ { (x , y) → x , f x y }) R)
+  mapWithKeyTotal {f = f}{R} tot {a} = ψ (ϕ tot)
```
WhatisRT, line 66: Here's a much more readable way of defining this function:
```
  mapWithKeyTotal tot with from ∈-map tot
  ... | (_ , refl , aRb) = ∈-map′ (∈-map′ aRb)
```

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,76 @@
...
+  -- updateFn:  Helper for defining update functions.
+  --            Given (a , b) and x y, if x ≡ a return b, otherwise return y.
+  updateFn : A × B → A → B → B
+  updateFn (a , b) x y with (x ≟ a)
+  ... | yes _ = b
+  ... | no _ = y
+
+  open TotalMap
+
+  mapWithKeyTotal : ∀{B'}{f : A → B → B'}{R : Rel A B} → total-map R → total-map (map (λ { (x , y) → x , f x y }) R)
```
WhatisRT, line 65: This line is a little long

```suggestion
  mapWithKeyTotal : ∀{B'}{f : A → B → B'}{R : Rel A B}
                  → total-map R → total-map (map (λ { (x , y) → x , f x y }) R)
```

src/Axiom/Set/TotalMapOn.agda
```diff
@@ -0,0 +1,84 @@
...
+open Equivalence  using (to)
+open Unionᵗᵐ       using (updateFn)
+
+private variable A B : Type
+
+_TotalOn_ : Rel A B → Set A → Type
+R TotalOn X = X ⊆ dom R
+
+
+record TotalMapOn {A : Type}(X : Set A)(B : Type) : Type where
```
WhatisRT, line 26: Many comments from the other file apply here too

src/Axiom/Set/TotalMapOn.agda
```diff
@@ -0,0 +1,84 @@
...
+  lookupLemma-Σ : {aa : Σ A (_∈ domain)} {b : B} → (proj₁ aa , b) ∈ rel → lookup aa ≡ b
+  lookupLemma-Σ {aa} = lookupLemma {proj₁ aa}{proj₂ aa}
+
+
+module Union-tm {B : Type} ⦃ _ : DecEq A ⦄ where
+
+  open TotalMapOn
+
+  mapWithKeyTotalOn :  {X : Set A}{B' : Type}{f : A → B → B'}{R : Rel A B}
+    →                  R TotalOn X → (map (λ { (x , y) → x , f x y }) R) TotalOn X
```
WhatisRT, line 60:
```suggestion
                    →  R TotalOn X → (map (λ { (x , y) → x , f x y }) R) TotalOn X
```

src/Axiom/Set/TotalMapOn.agda
```diff
@@ -0,0 +1,84 @@
...
+
+  mapWithKeyTotalOn {B' = B'}{f}{R} tot {a} a∈X = Goal
```
WhatisRT, line 62: Here's a way more readable way of defining this:
```
mapWithKeyTotalOn tot a∈X = ∈-map′ (∈-map′ (proj₂ (to dom∈ (tot a∈X))))
```

WhatisRT (changes requested):
I've noticed that you're really inconsistent about whitespace for parenthesized arguments, for example:
```
∀{a : A} → a ∈ dom rel
{a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
{a : A} {a∈dom : a ∈ domain} {b : B} → (a , b) ∈ rel → lookup (a , a∈dom) ≡ b
```
The rest of the code base always has spaces between parentheses and `∀`, so please try to match that style.

williamdemeo:
> I've noticed that you're really inconsistent about whitespace for parenthesized arguments, for example:
> 
> ```
> ∀{a : A} → a ∈ dom rel
> {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
> {a : A} {a∈dom : a ∈ domain} {b : B} → (a , b) ∈ rel → lookup (a , a∈dom) ≡ b
> ```
> 
> The rest of the code base always has spaces between parentheses and `∀`, so please try to match that style.

I think I've been fairly consistent about using no space between forall and the next parenthesis because to me it looks more like what I'd expect to see in a logic textbook. I didn't notice that your preferred style is to include a space.  I'll add spaces from now on.

src/Axiom/Set/TotalMap.agda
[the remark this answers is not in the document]
williamdemeo: Yes, I agree the name `total` seems better than `total-map`.  Initially I went ahead and changed it, but then I changed it back in case you had a reason to prefer the latter. I'll change it to `total`.

src/Axiom/Set/TotalMap.agda
williamdemeo: Okay, I'll change the field names.  (My justification was that for a records that we use often, we could get used to these short three-letter names to keep the code more concise, assuming they're going to show up in many places.  But in general, I agree, clarity is more important than brevity.)

src/Axiom/Set/TotalMap.agda
williamdemeo: Initially I called it something like `toMap`, but then I realized it's a reduct and when possible I like to use a name that calls attention to the thing's mathematical properties or context... but I'll change it back to `toMap`.

src/Axiom/Set/TotalMap.agda
williamdemeo: I agree, the `where` block seems unnecessary here.  The reason I left it that way is because when I inline it I need to specify one of the implicit parameters (the `a`, if I remember correctly).  So I guess this is a good example of "broken" inference...?  Anyway, I'll inline it.

williamdemeo: (I inlined it and somehow it's no longer requiring an implicit parameter... weird.  Must be because the code is quite different now to when I had tried to inline it before.)

williamdemeo:
I added spaces after foralls, and... eeeww! 😆  That convention may be hard for me to get used to, but I'll try my best! 😄 

> > I've noticed that you're really inconsistent about whitespace for parenthesized arguments, for example:
> > ```
> > ∀{a : A} → a ∈ dom rel
> > {a : A}{b : B} → (a , b) ∈ rel → lookup a ≡ b
> > {a : A} {a∈dom : a ∈ domain} {b : B} → (a , b) ∈ rel → lookup (a , a∈dom) ≡ b
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The rest of the code base always has spaces between parentheses and `∀`, so please try to match that style.
> 
> I think I've been fairly consistent about using no space between forall and the next parenthesis because to me it looks more like what I'd expect to see in a logic textbook. I didn't notice that your preferred style is to include a space. I'll add spaces from now on.

src/Axiom/Set/TotalMap.agda
williamdemeo: Okay, so... `rel⇒lookup` ?

williamdemeo:
I think I've implemented all your recommendations, plus a few more... so it's ready for rereview and hopefully ready to be merged.

src/Axiom/Set/TotalMap.agda
```diff
@@ -1,34 +1,62 @@
...
-private macro
-  ∈⇒P = anyOfⁿᵗ (quote ∈-filter⁻' ∷ quote ∈-∪⁻ ∷ quote ∈-map⁻' ∷ quote ∈-fromList⁻ ∷ [])
-  P⇒∈ = anyOfⁿᵗ (quote ∈-filter⁺' ∷ quote ∈-∪⁺ ∷ quote ∈-map⁺' ∷ quote ∈-fromList⁺ ∷ [])
-  ∈⇔P = anyOfⁿᵗ (quote ∈-filter⁻' ∷ quote ∈-∪⁻ ∷ quote ∈-map⁻' ∷ quote ∈-fromList⁻ ∷ quote ∈-filter⁺' ∷ quote ∈-∪⁺ ∷ quote ∈-map⁺' ∷ quote ∈-fromList⁺ ∷ [])
 
 -- defines a total map for a given set
-total-map : Rel A B → Type
-total-map R = ∀ {a} → a ∈ map proj₁ R
+total : Rel A B → Type
+total R = ∀ {a} → a ∈ map proj₁ R
```
WhatisRT, line 23:
```suggestion
total R = ∀ {a} → a ∈ dom R
```

src/Axiom/Set/TotalMapOn.agda
```diff
@@ -0,0 +1,65 @@
...
+_TotalOn_ : Rel A B → Set A → Type
+R TotalOn X = X ⊆ dom R
+
+
+record TotalMapOn {A : Type}(X : Set A)(B : Type) : Type where
+  field  rel              : Set (A × B)
+         left-unique-rel  : left-unique rel
+         total-rel        : rel TotalOn X
+
+  domain : Set A
```
WhatisRT, line 29: Now that I look at this, I think I'd remove that. It's not actually the domain, since we just have `domain ⊆ dom rel`, but not the other inclusion, and it's confusing to have the two that aren't actually equal.

WhatisRT (approved):
I've pointed out two minor things, but other than that LGTM. If you're done, please squash merge.

src/Axiom/Set/TotalMapOn.agda
williamdemeo: I agree. I'll remove it and then merge.

Thanks for all the helpful comments!

PR #136, opened 2023-06-29
Add minimum AVS

### LINKED ISSUE #124: Minimum AVS for ratification
We need to add a new protocol parameter & a check in `RATIFY` that at least that amount of stake participated.

This is the companion issue to https://github.com/input-output-hk/cardano-ledger/issues/3456

WhatisRT:
Closes #124

### DIFF
--- a/src/Ledger/Ratify.lagda
+++ b/src/Ledger/Ratify.lagda
@@ -200,16 +200,24 @@ acceptedStake r cc dists votes =
 totalStake : GovRole → ℙ VDeleg → StakeDistrs → (VDeleg ⇀ Vote) → Coin
 totalStake r cc dists votes = Σᵐᵛ[ x ← getStakeDist r cc dists ∣ votedAbstainHashes votes r ᶜ ᶠᵐ ] x
 
+activeVotingStake : ℙ VDeleg → StakeDistrs → (VDeleg ⇀ Vote) → Coin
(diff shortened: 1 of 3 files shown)


src/Ledger/Ratify.lagda
```diff
@@ -200,17 +200,25 @@ acceptedStake r cc dists votes =
 totalStake : GovRole → ℙ VDeleg → StakeDistrs → (VDeleg ⇀ Vote) → Coin
 totalStake r cc dists votes = Σᵐᵛ[ x ← getStakeDist r cc dists ∣ votedAbstainHashes votes r ᶜ ᶠᵐ ] x
 
+activeVotingStake : ℙ VDeleg → StakeDistrs → (VDeleg ⇀ Vote) → Coin
+activeVotingStake cc dists votes = Σᵐᵛ[ x ← getStakeDist DRep cc dists ∣ dom (votes ˢ) ᶜ ᶠᵐ ] x
```
Soupstraw, line 204: `cc` is a bit misleading here, maybe `vdelegs` would be better?

WhatisRT: This is probably something to refactor. Note that the name `cc` is also used in the previous definition, and this argument is just used to pass the constitutional committee (anything else in there is ignored).

williamdemeo (approved):
All looks good (and it type-checks on my machine).  No notes.  LGTM

PR #139, opened 2023-07-12
Token algebra example: ValueSet

williamdemeo:
This PR provides a new module called `Ledger.TokenAlgebra.ValueSet` which uses the new functionality in `TotalMap` to define a token algebra using the `Theory` record of the `Axiom.Set` module.

To implement a token algebra using `Set` and modeling the value type as a *total* function on a given `X : Set A` I needed to add some functionality to `TotalMap` (and similar functionality to a new module, called `TotalMapOn` which restricts maps to the set `X`, but I have since abandoned the approach using `TotalMapOn` as I couldn't get it to work without adding a dependently typed version of the `replacement` axiom of the `Set` module, and we don't want to add axioms at this point for a number of reasons; this is why the comments below are crossed out).

The new functionality in `TotalMap` is part of [a separate PR](https://github.com/input-output-hk/formal-ledger-specifications/pull/138) which should be merged before this one.

~I considered somehow using @WhatisRT's restriction operator but the approach using records seems natural to me.  If someone wants to try integrating the existing restriction operator, by all means go for it or explain to me how it could/should be done.~

~I found it very difficult to get this to work and it seems I need a new axiom called `dreplacement` which is a version of `replacement` that works for functions whose domains are dependent pair (sigma) types.~

~To complete this example, so that `Everything` type-checks, would require expanding the `List` types that require a case for each axiom.  I haven't done that yet.  I started looking at it, but it seems complicated; maybe that's only because I don't understand the set-based list types yet.~

### DIFF
--- a/src/Axiom/Set.agda
+++ b/src/Axiom/Set.agda
@@ -75,6 +75,9 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   _∉_ : A → Set A → Type
   _∉_ = ¬_ ∘₂ _∈_
 
+  ∈-irrelevant : Set A → Type ℓ
+  ∈-irrelevant X = ∀ {a} (p q : a ∈ X) → p ≡ q
+
   open Equivalence
 
   _Preservesˢ_ : (Set A → Set B) → (∀ {A} → Set A → Type) → Type ℓ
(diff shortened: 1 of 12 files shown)


src/Axiom/Set.agda
```diff
@@ -140,6 +147,9 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   -- don't know that there's a set containing all members of a type, which this is equivalent to
   -- _⁻¹_ : (A → B) → Set B → Set A
   -- f ⁻¹ X = {!!}
+  ≡∈ : {X : Set A}{a a' : A} → a ∈ X → a ≡ a' → a' ∈ X
+  ≡∈ {a = a}{.a}a∈X refl = a∈X
+
```
williamdemeo, line 152: I would have thought this lemma is already provided somewhere else, but I didn't see it.  If someone has seen it before, please let me know where.

Issue #56, opened 2023-03-02
Remaining MA features/proofs


WhatisRT:
The current MA implementation is still missing a few things, in particular:

- [ ] One or more implementations of the `Value` type
- [ ] Adjusting `scriptsNeeded` to include the `PolicyID`s of `mint`
- [ ] Preservation of value for token algebras
- [ ] Maybe more?

PR #142, opened 2023-07-19
TokenAlgebra Example: ValueSet

williamdemeo:
# Description

This PR provides a new module called Ledger.TokenAlgebra.ValueSet which uses the new functionality in TotalMap to define a token algebra using the Theory record of the Axiom.Set module.

To implement a token algebra using Set and modeling the value type as a total function on a given X : Set A I needed to add some functionality to TotalMap (and similar functionality to a new module, called TotalMapOn which restricts maps to the set X, but I have since abandoned the approach using TotalMapOn as I couldn't get it to work without adding a dependently typed version of the replacement axiom of the Set module, and we don't want to add axioms at this point for a number of reasons; this is why the comments below are crossed out).

The new functionality in TotalMap is part of a separate PR which should be merged before this one.

This PR accomplishes the first task of issue #56.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Prelude.agda
+++ b/src/Ledger/Prelude.agda
@@ -86,7 +86,7 @@ open import Axiom.Set.Map th public
   renaming (Map to _⇀_)
 
 open import Axiom.Set.TotalMap th public
-open import Axiom.Set.TotalMapOn th public
+open import Axiom.Set.TotalMapOn th
 
 open L.Decˡ public
   hiding (_∈?_; ≟-∅)
--- a/src/Ledger/TokenAlgebra/ValueSet.lagda
+++ b/src/Ledger/TokenAlgebra/ValueSet.lagda
@@ -1,0 +1,214 @@
+\subsection{Value Set}
+\label{sec:tokenalgebra-valueset}
+
+\begin{code}[hide]
+{-# OPTIONS --safe --no-import-sorts #-}
+
+open import Agda.Primitive using () renaming (Set to Type)
+
+module Ledger.TokenAlgebra.ValueSet (PolicyId AssetName : Type) where
+
+open import Ledger.Prelude                         hiding (lookup ; update ; isMagma ; isEquivalence) renaming (TotalMap to _⇒_)
+open import Ledger.TokenAlgebra PolicyId           using (TokenAlgebra)
+
+open import Algebra                                using (CommutativeMonoid ; Op₂ ; IsSemigroup ; IsMonoid ; IsMagma ; IsCommutativeMonoid)
+open import Algebra.Morphism                       using (IsMonoidHomomorphism ; IsMagmaHomomorphism)
+open import Data.Nat                               using (_≤_)
+open import Data.Nat.Properties                    using (+-comm ; +-assoc ; +-identityʳ ; +-0-commutativeMonoid)
+open import Function.Related.TypeIsomorphisms      using (Σ-≡,≡→≡)
+open import Relation.Binary                        using (IsEquivalence)
+open import Relation.Binary.Morphism               using (IsRelHomomorphism)
+open import Relation.Binary.PropositionalEquality  using (module ≡-Reasoning)
+
+import Relation.Binary.PropositionalEquality as ≡
+import Relation.Binary.Core  as stdlib
+\end{code}
+
+\subsubsection{Derived types}
+
+(See Fig 3 of the
+\href{https://github.com/input-output-hk/cardano-ledger/releases/latest/download/mary-ledger.pdf}%
+{Mary ledger specification}.)
+
+\begin{itemize}
+\item \AgdaBound{AssetName} is a byte string used to distinguish different assets with the same \AgdaBound{PolicyId}.
+\item \AgdaBound{AssetId} is a product type consisting of a \AgdaBound{PolicyId} and an \AgdaBound{AssetName}.
+\item \AgdaBound{AdaId} is the Id for the asset Ada.
+\item \AgdaBound{Quantity} is the type of amounts of assets.
+\end{itemize}
+
+In the formal ledger specification \AgdaBound{AssetId} is sometimes viewed as a direct sum type,
+the inhabitants of which belong to either \AgdaBound{AdaIdType} or the product
+\AgdaBound{PolicyId}~\AgdaBound{×}~\AgdaBound{AssetName}; if we were adhering to that point of view,
+then we would have defined
+\AgdaBound{AssetId}
+  = \AgdaBound{AdaIdType}~\AgdaBound{⊎}~(\AgdaBound{PolicyId}~\AgdaBound{×}~\AgdaBound{AssetName}).
+
+Finally, we define a record type with a single inhabitant with which we may wish to
+represent the type of Ada (rather than viewing Ada as just another asset).
+
+\begin{code}
+record AdaIdType : Type where
+  instance constructor AdaId
+\end{code}
+
+
+\subsection{Definition of the value monoid}
+
+An inhabitant of `Value` is a map denoting a finite collection of quantities of assets.
+
+\begin{code}
+open CommutativeMonoid renaming (_∙_ to _⋆_) hiding (refl ; sym ; trans)
+
+AssetId : Type
+AssetId = PolicyId × AssetName
+
+module _
+  {X : ℙ AssetId}
+  {⋁A : isMaximal X}
+  ⦃ dec : DecEq (PolicyId × AssetName) ⦄
+  where
+
+  open _⇒_
+  open ≡-Reasoning
+  open FunTot X ⋁A
+
+  Quantity : Type
+  Quantity = ℕ
+
+  _⊕_ : Op₂ (AssetId ⇒ Quantity)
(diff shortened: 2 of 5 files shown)


src/Ledger/Prelude.agda
```diff
@@ -86,7 +86,7 @@ open import Axiom.Set.Map th public
   renaming (Map to _⇀_)
 
 open import Axiom.Set.TotalMap th public
-open import Axiom.Set.TotalMapOn th public
+open import Axiom.Set.TotalMapOn th
```
williamdemeo, line 89: This line was inserted just to make sure the build tests don't fail with error "TotalMapOn.agdai" doesn't exist.  (But it need not be inherited by modules that import this one, so no need for `public`.)

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,222 @@
...
+  isSemigrp .isMagma .∙-cong {u}{v}{x}{y}  = ⊕-cong {u}{v}{x}{y} -- (broken inference)
+  isSemigrp .assoc                         = ⊕-assoc
+
+  open IsMonoid
+  ≋-⊕-ι-isMonoid : IsMonoid _≋_ _⊕_ ι
+  ≋-⊕-ι-isMonoid .isSemigroup  = isSemigrp
+  ≋-⊕-ι-isMonoid .identity     = ι-identity
+\end{code}
+
+We are now in a position to define the commutative monoid.
```
williamdemeo, line 157: TODO: possibly add some more explanatory prose about the `Value-TokenAlgebra` example.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+  _≋_ : stdlib.Rel (AssetId ⇒ Quantity) 0ℓ
+  u ≋ v = ∀{aa} → (lookup u) aa ≡ (lookup v) aa
+
+  ≋-isEquivalence : IsEquivalence {0ℓ} _≋_
+  ≋-isEquivalence = record { refl = ≡.refl ; sym = λ x → ≡.sym x ; trans = λ x y → ≡.trans x y }
+
+  ⊕-cong : Algebra.Congruent₂ _≋_ _⊕_
+  ⊕-cong {x}{y}{u}{v} x≋y u≋v {a} = begin
+    lookup (x ⊕ u) a         ≡⟨ ⊕-lemma x u ⟩
+    lookup x a + lookup u a  ≡⟨ cong (λ w → w + lookup u a) x≋y ⟩
```
WhatisRT, line 112: You can use `cong₂ _+_ x≋y u≋v` to rewrite both at the same time

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+  ⊕-comm u v {aa} = begin
+    lookup (u ⊕ v) aa          ≡⟨ ⊕-lemma u v ⟩
+    lookup u aa + lookup v aa  ≡⟨ +-comm (lookup u aa) (lookup v aa) ⟩
+    lookup v aa + lookup u aa  ≡⟨ ≡.sym (⊕-lemma v u) ⟩
+    lookup (v ⊕ u) aa          ∎
+
+  ⊕-assoc : Algebra.Associative _≋_ _⊕_
+  ⊕-assoc x y z {a} = begin
+    lookup ((x ⊕ y) ⊕ z) a                  ≡⟨ ⊕-lemma (x ⊕ y) z ⟩
+    lookup (x ⊕ y) a + lookup z a           ≡⟨ cong (λ w → w + lookup z a) (⊕-lemma x y) ⟩
```
WhatisRT, line 127:
```suggestion
    lookup (x ⊕ y) a + lookup z a           ≡⟨ cong (_+ lookup z a) (⊕-lemma x y) ⟩
```
You can make a similar simplification lots of times below.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+  Vcm .Carrier                        = AssetId ⇒ Quantity
+  Vcm ._≈_                            = _≋_
+  Vcm ._⋆_                            = _⊕_
+  Vcm .ε                              = ι
+  Vcm .isCommutativeMonoid .isMonoid  = ≋-⊕-ι-isMonoid
+  Vcm .isCommutativeMonoid .comm      = ⊕-comm
+
+  Value-TokenAlgebra :  (specialPolicy : PolicyId)
+                        (specialAsset : AssetName)
+                        (policies : AssetId ⇒ Quantity → ℙ PolicyId)
```
WhatisRT, line 171: `policies` should just project out the `PolicyId`s from the domain, i.e. it shouldn't be a parameter here

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+    totalMap↠coin tm = lookup tm specId
+
+    coin↪totalMap : Coin → AssetId ⇒ Quantity
+    coin↪totalMap c = update specId c zeroMap
+
+    leq : AssetId ⇒ Quantity → AssetId ⇒ Quantity → Type
+    leq = λ u v → ∀ {a}{p}{q} → lookup u (a , p) ≤ lookup v (a , q)
+
+    compose-to-id : totalMap↠coin ∘ coin↪totalMap ≗ id
+    compose-to-id _ = lookup-update-id zeroMap where open LookupUpdate {X = X}{specId}{⋁A}
```
WhatisRT, line 209:
```suggestion
    compose-to-id _ = lookup-update-id zeroMap where open LookupUpdate {X = X} {specId} {⋁A}
```
Also, the `where` block should be on a new line, or use `let` if you want to do it in the same line instead.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+  Vcm ._⋆_                            = _⊕_
+  Vcm .ε                              = ι
+  Vcm .isCommutativeMonoid .isMonoid  = ≋-⊕-ι-isMonoid
+  Vcm .isCommutativeMonoid .comm      = ⊕-comm
+
+  Value-TokenAlgebra :  (specialPolicy : PolicyId)
+                        (specialAsset : AssetName)
+                        (policies : AssetId ⇒ Quantity → ℙ PolicyId)
+                        (size : AssetId ⇒ Quantity → ℕ)
+                        {decTot : DecEq (AssetId ⇒ Quantity)}
```
WhatisRT, line 173: We may need to do something about this, since we can't ever supply this constructively. I think we need to require that the equivalence relation in `Vcm` is decidable, but I don't know how this affects everything else...

WhatisRT (changes requested):
There's a bit of cleanup still to do, but looks mostly good!

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: Okay, thanks for pointing this out.  I'll have to think about how to resolve it.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+open CommutativeMonoid renaming (_∙_ to _⋆_) hiding (refl ; sym ; trans)
+
+AssetId : Type
+AssetId = PolicyId × AssetName
+
+module _ {X : ℙ AssetId} {⋁A : isMaximal X} where
+
+  open _⇒_
+  open ≡-Reasoning
+  open FunTot {AssetId}{X}{⋁A}
```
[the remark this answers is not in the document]
williamdemeo: Okay, I'll make `X` and `⋁A` explicit.  (`AssetId` can be infered.)

src/Ledger/TokenAlgebra/ValueSet.lagda
williamdemeo: I think this is something we should leave for future consideration.

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -0,0 +1,221 @@
...
+  ι-lem1 : ∀{a} → (a , lookup ι a) ∈ rel ι
+  ι-lem1 = lookup∈rel ι
+
+  ι-lem2 : ∀{a} → (a , 0) ∈ rel ι
+  ι-lem2 = ≡→∈ ι-lem1 (Σ-≡,≡→≡ (≡.refl , ι-lem0))
+
+  ι-lem3 : ∀{a : AssetId} → lookup ι a ≡ 0
+  ι-lem3 = (left-unique-rel ι) ι-lem1 ι-lem2
+
+  _≋_ : stdlib.Rel (AssetId ⇒ Quantity) 0ℓ
```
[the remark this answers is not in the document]
williamdemeo: I think we can leave this for future improvements.

WhatisRT (approved):
LGTM

PR #160, opened 2023-08-01
Improve building the PDF via Makefile

WhatisRT:
# Description

This fixes `Makefile` being outdated & improves `shell.nix` to be able to use `Makefile` more conveniently.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,14 @@ To install Agda locally and use that install with emacs, you can do the followin
 
 `nix-shell` provides Agda complete with the correct dependencies. So you should be able to run your preferred editor within `nix-shell` and it should see the required `agda` executable.
 
+## Building the PDF quickly
+
+The Makefile can be used to build the PDF without having to build everything else. Either run `make` from within `nix-shell`, or use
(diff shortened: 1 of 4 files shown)


williamdemeo (changes requested):
It works, but I did need to run `nix-shell --command make` three times to make the references resolve.  Would it be hard to make it so that the latex compilation step is executed as many times as needed to resolve all refs?

williamdemeo:
Also, the pdf ends up in the top level directory in the file `PDF.pdf`.  Should it not be called `cardano-ledger.pdf` ...and maybe for consistency should it end up in a `result` directory?

WhatisRT:
The `result` directory is a nix thing, so I prefer not messing with that. I've renamed it to `cardano-ledger.pdf` & switched to `latexmk`.

williamdemeo (approved):
Works great now.  Nice work!

williamdemeo:
I think your last rebase may have clobbered your improvements to the Makefile.  Sorry I missed that before merging.  I'll open a new PR with your previous Makefile mods.

PR #164, opened 2023-08-02
Fix missing latex dependency in `shell.nix`

WhatisRT:
# Description

This fixes that the proper latex distribution was missing from the default nix shell. Before, `nix-shell --pure --command make` would fail because `latexmk` was missing, and this fixes it.

# Checklist

- [x] Self-reviewed the diff


williamdemeo (approved):
Looks good.

(I accidentally merged a change to .gitignore into this PR, but I reset it back to where you left it... sorry.)

PR #168, opened 2023-08-08
Bump stdlib dependency to fix monoid solver performance

UlfNorell:
The change to the standard library is this:

https://github.com/input-output-hk/agda-stdlib/commit/bdfab77b179c937856c49d72321ca26e2a27d568

and makes the monoid solver not normalise goals before solving. This is what you want in most cases, and if you need normalisation the old version is still available as `solve-normalised`.

This fixes the performance issues type checking `Ledger.Utxo.Properties`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -13,9 +13,9 @@ let
     version = "1.7.2";
     src = fetchFromGitHub {
       repo = "agda-stdlib";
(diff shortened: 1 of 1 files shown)


WhatisRT (approved):
LGTM

PR #171, opened 2023-08-09
Bump stdlib-meta dependency to fix for slow type checking of derived DecEq instances

UlfNorell:
The stdlib-meta commit is here:

https://github.com/input-output-hk/stdlib-meta/commit/897555aefd77c83e88492255641da3bcc4879461

This change brings type checking time of `Ledger.PParams` down from 45s to 2s.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -25,8 +25,8 @@ let
     src = fetchFromGitHub {
       repo = "stdlib-meta";
       owner = "input-output-hk";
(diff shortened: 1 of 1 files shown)


WhatisRT (approved):
LGTM! I think that's now somewhere in between 1.5-2 minutes time saving in CI in total, but we'll be saving that lots of times interactively!

Issue #34, opened 2022-11-25
Type classes for common structures


WhatisRT:
We have a bunch of structures with lots of name clashes that would likely benefit a lot from being turned into type classes. Here are some:

- [x] Addition
- [x] Subtraction
- [x] Multiplication
- [x] All the inequalities
- [x] Decidable inequalities

williamdemeo:
I've checked multiplication off the list because I can't find any instances of a decorated multiplication symbol.

PR #173, opened 2023-08-10
Type classes for common structures

williamdemeo:
This addresses issue #34 for the special case of addition in the `Ledger.Utxo.Properties` module.

The main goal of this PR is to allow us to use a single operation symbol, `_+_`, to denote addition over each of the types `ℤ`, `ℕ`, and `TokenAlgebra.Value`.

Most of the changes in this PR---including the clean-up of proofs and the removal of some uses of tactics---were made in an effort to support this operator overloading (via implicit instances), including

+  a large number of changes are proposed that significantly clean up and simplify the presentation and the proofs in the `Utxo.Properties` module

+  eliminate unused code (imports, private variables, and unnecessary implicit arguments)

+  use of tactics in equational reasoning proofs were eliminated because I was unable to get the tactic proofs to type-check when using the new implicit instances for types with addition

# Description

# Checklist

- [ ] Self-reviewed the diff


src/Ledger/Utxo/Properties.lagda
```diff
@@ -8,24 +8,25 @@ open import Ledger.Transaction
...
+open import Ledger.Prelude renaming (_+_ to _+ℕ_)
+
+open import Algebra              using     (CommutativeMonoid ; Monoid)
+open import Algebra.Morphism     using     (module MonoidMorphisms ; IsMagmaHomomorphism)
+open import Data.Integer         using     (ℤ; _⊖_ ; _-_ ; 0ℤ)
+                                 renaming  (_+_ to _+ℤ_ ; +_ to _⁺ ; -_ to _⁻)
+open import Data.Integer.Ext     using     (posPart ; negPart)
+open import Data.Nat.Properties  hiding    (_≟_)
+open import Data.Sign            using     (Sign)
+open import Relation.Binary      using     (IsEquivalence)
```
williamdemeo, line 21: It was hard for me to carry out the revision/upgrade to implicit instances without simplifying and organizing the imports.  The use of instances like the `Add` record used below naturally cause conflicts with other imports, but this was certainly not the only name clash I encountered while working on this PR. 

Not only will fewer clashes occur if we are more careful/disciplined about imports, but also I find it much easier/faster to resolve such conflicts if the imports are well-organized, explicit, and as small as possible.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -8,24 +8,25 @@ open import Ledger.Transaction
...
-open import Data.Integer.Ext
-import Data.Integer.Properties as ℤ
-open import Interface.ComputationalRelation
-open import Relation.Binary
 open import Tactic.Cong
 open import Tactic.EquationalReasoning
 open import Tactic.MonoidSolver
 
+import Data.Nat as ℕ
+import Data.Integer.Properties as ℤ
```
williamdemeo, line 28: Minor remarks on some of the existing code:

1. Referring to `Data.Nat` as `ℕ` is okay, I suppose... but doesn't this mean the natural numbers as called `ℕ.ℕ`?
2. Referring to the `Data.Integer.Properties` module as `ℤ` seems even stranger still.

(I tried to stick with conventions such as these wherever possible.)

src/Ledger/Utxo/Properties.lagda
```diff
@@ -39,23 +40,35 @@ open Tx
...
+instance
+  addInt : Add ℤ
+  addInt = record { _+_ = _+ℤ_ }
+
+  addNat : Add ℕ
+  addNat = record { _+_ = _+ℕ_ }
+
+  addCommMon : Add (TokenAlgebra.Value tokenAlgebra)
+  addCommMon = record { _+_ = _+ᵛ_ }
+
```
williamdemeo, line 64: This is just one way to add a very basic "type-class"-like feature by which Agda knows which `_+_` symbol we mean.  Later we may want to add some more sophisticated instance records, such as Monoids, or Commutative Monoids, or Groups, or Symmetric Monoidal Categories, or... but I just wanted to get something simple done and working and avoid getting bogged down by unnecessarily complex or fancy type classes.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -39,23 +40,35 @@ open Tx
...
-  donation donations donations' : Coin
-  Γ : UTxOEnv
-  s s' : UTxOState
-  A : Set
+  tx                               : TxBody
+  utxo utxo'                       : UTxO
+  Γ                                : UTxOEnv
+  utxoState utxoState'             : UTxOState
+  fees fees' donations donations'  : Coin
+  deposits deposits'               : DepositPurpose ⇀ Coin
```
williamdemeo, line 71: Some unused private variables were eliminated.  Alignment is merely to make it easier for me to read and refer to. (I do realize this part of the code will not make it into the documentaion.)

src/Ledger/Utxo/Properties.lagda
```diff
@@ -39,23 +40,35 @@ open Tx
 open Equivalence
 open Properties
 
-import Data.Nat.Tactic.RingSolver as ℕ
```
williamdemeo, line 42: This import seems unnecessary.  Also, what happens when we give both `Data.Nat` and `Data.Nat.Tactic.RingSolver` the same handle?  That's frightening. 😨

src/Ledger/Utxo/Properties.lagda
```diff
@@ -8,24 +8,25 @@ open import Ledger.Transaction
...
 
-import Data.List as List
-import Data.Nat as ℕ
-open import Algebra.Morphism
-open import Data.Nat.Properties using
-  (+-0-commutativeMonoid; +-0-monoid; +-comm; +-identityʳ; +-assoc; +-∸-assoc; m≤n⇒m≤n+o; ≤″⇒≤; m+[n∸m]≡n)
-open import Data.Sign using (Sign)
-open import Data.Integer as ℤ using (ℤ; _⊖_)
-open import Data.Integer.Ext
-import Data.Integer.Properties as ℤ
```
williamdemeo, line 22: Minor remarks on some of the existing code:

1. Referring to Data.Nat as ℕ is okay, I suppose... but doesn't this mean the natural numbers are called ℕ.ℕ? 😆 
2. Referring to the Data.Integer.Properties module as ℤ seems even stranger still.

(I tried to stick with conventions such as these wherever possible.)

src/Ledger/Utxo/Properties.lagda
```diff
@@ -91,39 +104,42 @@ consumedCoinEquality : ∀ {pp} → coin (mint tx) ≡ 0
                      → coin (consumed pp utxoState tx)
                      ≡ cbalance ((UTxOState.utxo utxoState) ∣ txins tx) + depositRefunds pp utxoState tx
 consumedCoinEquality {tx} {utxoState} {pp} h = let utxo = UTxOState.utxo utxoState in begin
-  coin (balance (utxo ∣ txins tx) +ᵛ mint tx +ᵛ inject (depositRefunds pp utxoState tx))
+  coin (balance (utxo ∣ txins tx) + mint tx + inject (depositRefunds pp utxoState tx))
     ≡⟨ ∙-homo-Coin _ _ ⟩
-  coin (balance (utxo ∣ txins tx) +ᵛ mint tx) + coin (inject $ depositRefunds pp utxoState tx)
-    ≡⟨ cong (coin (balance (utxo ∣ txins tx) +ᵛ mint tx) +_) (property _) ⟩
-  coin (balance (utxo ∣ txins tx) +ᵛ mint tx) + depositRefunds pp utxoState tx
-    ≡⟨ cong! (∙-homo-Coin _ _) ⟩
```
williamdemeo, line 99: This is one of the proof steps that (I think) must be re-written in order to use the overloaded `_+_` operator symbol. If someone know how to get tactics and instances to work well together in this situation, please explain.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -91,39 +104,42 @@ consumedCoinEquality : ∀ {pp} → coin (mint tx) ≡ 0
...
-      coin (balance (outs tx) +ᵛ inject (txfee tx)) + newDeposits pp utxoState tx
-        ≡⟨ cong! (∙-homo-Coin _ _) ⟩
-      coin (balance (outs tx)) + coin (inject (txfee tx)) + newDeposits pp utxoState tx
-        ≡⟨ cong (λ x → cbalance (outs tx) + x + newDeposits pp utxoState tx) (property (txfee tx)) ⟩
-      cbalance (outs tx) + (txfee tx) + newDeposits pp utxoState tx                     ∎
-    )⟩
-  cbalance (outs tx) + (txfee tx) + newDeposits pp utxoState tx + coin (inject (txdonation tx))
-    ≡⟨ cong (cbalance (outs tx) + (txfee tx) + newDeposits pp utxoState tx +_) (property _) ⟩
-  cbalance (outs tx) + (txfee tx) + newDeposits pp utxoState tx + txdonation tx
-    ∎
```
williamdemeo, line 126: I think we should try to avoid nesting equational reasoning proofs.  The re-written version is easier to parse and understand (imho).

src/Ledger/Utxo/Properties.lagda
```diff
@@ -192,22 +218,22 @@ module DepositHelpers
...
+    dep ⁺  + (uDep ⊖ uDep)  ≡⟨ cong (dep ⁺ +_) (ℤ.n⊖n≡0 uDep) ⟩
+    dep ⁺ + 0ℤ              ≡⟨ ℤ.+-identityʳ _ ⟩
+    dep ⁺                   ∎
 
   ref-tot-0 : ref ≢ 0 → tot ≡ 0
   ref-tot-0 ref≢0 with Δdep
-  ... | ℤ.+_ n     = ⊥-elim (ref≢0 refl)
-  ... | ℤ.negsuc n = refl
+  ... | n ⁺         = ⊥-elim (ref≢0 refl)
+  ... | ℤ.negsuc n  = refl
```
williamdemeo, line 236: We should probably make an alias for `ℤ.negsuc` as well. 🤔

src/Ledger/Utxo/Properties.lagda
```diff
@@ -219,63 +245,58 @@ module DepositHelpers
...
-  rearrange0 : txfee tx + txdonation tx + (tot + remDepTot) + fees ≡
-        fees + txfee tx + getCoin deposits' + txdonation tx
+          in (disjoint-∪ᵐˡ-∪ (disjoint-sym res-ex-disjoint) ≡ᵉ-∘ ∪-sym) ≡ᵉ-∘ res-ex-∪ (_∈? txins tx))
+
+
+  rearrange0 : txfee tx + don + (tot + remDepTot) + fees ≡ fees + txfee tx + getCoin deposits' + don
   rearrange0 = begin
-    txfee tx + txdonation tx + (tot + remDepTot) + fees
-      ≡⟨ +-comm (txfee tx + txdonation tx + (tot + remDepTot)) fees ⟩
-    fees + (txfee tx + txdonation tx + (tot + remDepTot)) ≡t⟨⟩
```
williamdemeo, line 261: `≡t⟨⟩` is another tactic I couldn't get working with the overloaded `_+_` (cf. [the remark above](https://github.com/input-output-hk/formal-ledger-specifications/pull/173/files#r1290266719))

WhatisRT (changes requested):
`Add` is exactly what I had in mind, but it should go into a new module. In Lean this is called `HasAdd`, which I think I prefer, and then this should go into `Interface.HasAdd`. Then, `Ledger.Prelude` should reexport the contents of `Interface.HasAdd` publicly, and the name clashes can be resolved there, once and for all.

The only thing that has to change with this structure is that the instance for the token algebra can't be declared there, but it can be added to the `TokenAlgebra` record instead.

There's also a lot of refactoring of `Ledger.Utxo.Properties` in here, which I generally like but belongs in a separate PR. In particular, if the type class means we can't use the monoid solver, I think we should locally not use the type class, so that we can keep using the monoid solver. This can be done by having `ℕ._+_` still around and using it just in those few lines where we use the solver. Same thing applies to `cong!`.

I think the reason the monoid solver breaks here is because of Ulf's recent changes that provide major speedup, but it should be easy to slightly change the monoid solver so that it keeps working.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -135,8 +151,9 @@ balValueToCoin {tx} {utxoState} {pp} h h' = begin
   coin (produced pp utxoState tx) ≡⟨ producedCoinEquality {utxoState} {tx} {pp} ⟩
   cbalance (outs tx) + (txfee tx) + newDeposits pp utxoState tx + txdonation tx ∎
 
+
 posPart-negPart≡x : ∀ {x} → posPart x ⊖ negPart x ≡ x
-posPart-negPart≡x {ℤ.+_ n}     = refl
+posPart-negPart≡x {n ⁺}     = refl
 posPart-negPart≡x {ℤ.negsuc n} = refl
```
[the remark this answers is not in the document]
jmchapman: I think this is a step to far personally. I would stick with:
```
posPart-negPart≡x {ℤ.+ n}  = refl
```
Note you can omit the `_`.

jmchapman:
@williamdemeo , I discussed this with @WhatisRT. The aim is to use the automation from the tactics as much as possible. So, don't use the type class in proofs where the tactic doesn't work. Please make an issue about the limitations of the tactics interaction with type classes.

Please also also make an issue about import lists.

williamdemeo:
Thanks for the helpful feedback guys.  I'll improve this asap and let you know when it's ready for re-review.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -192,22 +218,22 @@ module DepositHelpers
...
+    uDep ⁺ + (ref ⊖ 0)      ≡˘⟨ cong (λ x → uDep ⁺ + (ref ⊖ x)) tot≡0 ⟩
+    uDep ⁺ + (ref ⊖ tot)    ≡⟨ cong (_+_ (uDep ⁺)) (ℤ.⊖-swap ref tot) ⟩
+    uDep ⁺ - (tot ⊖ ref)    ≡˘⟨ cong! deposits-change' ⟩
+    uDep ⁺ - Δdep           ≡˘⟨ cong (uDep ⁺ +_) (ℤ.⊖-swap dep uDep) ⟩
+    uDep ⁺ + (dep ⊖ uDep)   ≡⟨ ℤ.distribʳ-⊖-+-pos uDep dep uDep ⟩
+    uDep + dep ⊖ uDep       ≡⟨ cong (_⊖ uDep) (+-comm uDep dep) ⟩
+    dep + uDep ⊖ uDep       ≡˘⟨ ℤ.distribʳ-⊖-+-pos dep uDep uDep ⟩
+    dep ⁺  + (uDep ⊖ uDep)  ≡⟨ cong (dep ⁺ +_) (ℤ.n⊖n≡0 uDep) ⟩
+    dep ⁺ + 0ℤ              ≡⟨ ℤ.+-identityʳ _ ⟩
+    dep ⁺                   ∎
```
williamdemeo, line 231: @jmchapman re. your comment about `ℤ.+ n` -> `n ⁺` being a "step too far" ...do you feel the same here?  I think the superscript + notation for the positive part makes this proof look much nicer.  Please let me know if you prefer to revert to the blackboard-bold Z in this case as well.

src/Ledger/Utxo/Properties.lagda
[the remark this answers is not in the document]
williamdemeo: @jmchapman I don't think the `_` can be omitted.  If I do that, then I get the following error upon type-checking:

```
Could not parse the left-hand side posPart-negPart≡x {ℤ.+ n}
Problematic expression: (ℤ.+ n)
Operators used in the grammar:
  + (infixl operator, level 7) [_+_ (/home/williamdemeo/git/IO/fls/william/34-type-classes/src/Ledger/Utxo/Properties.lagda:51,9-12)]
when scope checking the left-hand side posPart-negPart≡x {ℤ.+ n} in
the definition of posPart-negPart≡x
```

williamdemeo:
@jmchapman and @WhatisRT I'll make an issue about the type-class tactic conflicts I encountered.  In the meantime, I'll change the code back to using `cong!` instead of `cong` and `≡t⟨⟩` instead of `≡⟨ [a proof step] ⟩`.

Just for the record I'll point out that `cong!` does not save a proof step; it just makes that step more terse (or, you might say, "more concise" 😏), so we're now using `cong!` to make the proof more concise at the cost of requiring a more verbose operator symbol around that step of the proof.

I sure hope we can resolve the type-class/tactic conflict soon, because one consequence of the change requests above will be that some proofs use both `+` and `ℤ.+` to denote the same operation... Similarly for `+` and `ℕ.+` in other proofs.  Given that we want to put the tactics back into the proofs and this requires the more verbose addition operator, do you think for now I should revert to `ℤ.+` for the entirety of such proofs so we are at least locally consistent?

williamdemeo:
> There's also a lot of refactoring of `Ledger.Utxo.Properties` in here, which I generally like but belongs in a separate PR.

I needed to do some refactoring in this PR in order to (a) make sense of the proofs and (b) resolve some import name clashes (the latter occurred as a result of importing various addition symbols for the type-class instances).

However, it seems you prefer I revert these improvements and submit them in a separate PR that does proof/import cleanup and nothing else, so that's what I'll do.

For the current PR, I'll revert the changes and make the file look as much like the original as possible. (Ironically, I was trying to take an extremely minimalist approach to introducing type classes. I guess it wasn't minimal enough. 😓) 

>  In particular, if the type class means we can't use the monoid solver, I think we should locally not use the type class, so that we can keep using the monoid solver. This can be done by having `ℕ._+_` still around and using it just in those few lines where we use the solver. Same thing applies to `cong!`.

Yes, I know that.  I will make those changes too, but as I mentioned in an earlier comment (above) this means we will have proofs that use two operator symbols to mean the same thing, which is a terrible idea.  For those proofs I think I should use the more verbose symbol throughout, if I need to use it at all.

> I think the reason the monoid solver breaks here is because of Ulf's recent changes that provide major speedup, but it should be easy to slightly change the monoid solver so that it keeps working.

Yes, Ulf's speedup is super! 💯 🥳  ...especially when type-checking this particular module.

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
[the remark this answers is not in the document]
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
```diff
@@ -153,34 +170,43 @@ module DepositHelpers
...
+
+    dep uDep ref tot remDepTot don tot+remDep : Coin
+    dep         = getCoin deposits
+    uDep        = getCoin (updateDeposits pp tx deposits)
+    ref         = depositRefunds pp utxoSt tx
+    tot         = newDeposits pp utxoSt tx
+    remDepTot   = getCoin deposits ∸ ref
+    don         = txdonation tx
+    tot+remDep  = tot + remDepTot
+
```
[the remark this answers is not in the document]
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
```diff
@@ -219,63 +245,58 @@ module DepositHelpers
 
   deposits-change : uDep ≡ dep + tot ∸ ref
   deposits-change = ℤ.+-injective $ begin
-    ℤ.+_ uDep                                 ≡˘⟨ ℤ.+-identityʳ _ ⟩
-    ℤ.+_ uDep ℤ.+ ℤ.0ℤ                        ≡˘⟨ cong! (ℤ.+-inverseˡ (ℤ.+_ dep)) ⟩
-    ℤ.+_ uDep ℤ.+ (ℤ.-_ (ℤ.+_ dep) ℤ.+ (ℤ.+_ dep))
-      ≡˘⟨ ℤ.+-assoc (ℤ.+_ uDep) (ℤ.-_ (ℤ.+_ dep)) (ℤ.+_ dep) ⟩
-    (ℤ.+_ uDep ℤ.- (ℤ.+_ dep)) ℤ.+ (ℤ.+_ dep) ≡⟨ cong! (ℤ.m-n≡m⊖n uDep dep) ⟩
-    Δdep ℤ.+ (ℤ.+_ dep)                       ≡⟨ ℤ.+-comm Δdep (ℤ.+_ dep) ⟩
```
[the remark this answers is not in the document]
williamdemeo: reverted ∵ change requests

src/Ledger/Utxo/Properties.lagda
williamdemeo: reverted ∵ change requests

WhatisRT:
Thank you, this is what I expected this module to look like. What's still missing is this part:

> `Add` is exactly what I had in mind, but it should go into a new module. In Lean this is called `HasAdd`, which I think I prefer, and then this should go into `Interface.HasAdd`. Then, `Ledger.Prelude` should reexport the contents of `Interface.HasAdd` publicly, and the name clashes can be resolved there, once and for all.

In particular, this means that there'll be a bunch more files where the notation for `+` needs to be changed appropriately, and this PR is exactly the place to do it.

williamdemeo:
I'll continue with this today while waiting for comments/change requests on the Ratification doc PR (which has higher priority).

jmchapman:
> @jmchapman and @WhatisRT I'll make an issue about the type-class tactic conflicts I encountered. In the meantime, I'll change the code back to using `cong!` instead of `cong` and `≡t⟨⟩` instead of `≡⟨ [a proof step] ⟩`.
> 
> Just for the record I'll point out that `cong!` does not save a proof step; it just makes that step more terse (or, you might say, "more concise" 😏), so we're now using `cong!` to make individual proof steps a bit more concise at the cost of requiring a more verbose operator symbol around that step.
> 
> I sure hope we can resolve the type-class/tactic conflict soon, because one consequence of the change requests above will be that some proofs use both `+` and `ℤ.+` to denote the same operation... Similarly for `+` and `ℕ.+` in other proofs. Given that we want to put the tactics back into the proofs and this requires the more verbose addition operator, do you think for now I should revert to `ℤ.+` for the entirety of such proofs so we are at least locally consistent?

Personally, given that there is an issue to fix this, I would be happy with the proof having both operators (with as much + as possible) for the moment. But, I defer to Andre for the final say.

Also, I think the motivation for `cong!` is to avoid boilerplate/noise that you wouldn't write on paper - i.e., identifying with precisely where you apply a rewrite in a term rather than leaving it up to the reader.

jmchapman:
> I needed to do some refactoring in this PR in order to (a) make sense of the proofs and (b) resolve some import name clashes (the latter occurred as a result of importing various addition symbols for the type-class instances).

1. What do you mean exactly by (a)?
2. Regarding (b) - I would recommend to do a minimal amount managing imports with only importing needed stuff/hiding stuff in this PR.

> However, it seems you prefer I revert these improvements and submit them in a separate PR that does proof/import cleanup and nothing else, so that's what I'll do.

Are you referring to a large scale parse of the repo introducing more use of import lists or something else? If the former, I recommend to make an issue, this can be a task on the backlog to pick up (by someone) at a later date if the team wants it. Don't make a PR yet.

williamdemeo:
> > I needed to do some refactoring in this PR in order to (a) make sense of the proofs and (b) resolve some import name clashes (the latter occurred as a result of importing various addition symbols for the type-class instances).
> 
> 1. What do you mean exactly by (a)?

I mean some of the proofs in the `UTxO.Properties` module are quite messy and convoluted.  It seems it a couple of cases they done quickly and never refined or simplified.  They could be made much easier to read and comprehend.

> 2. Regarding (b) - I would recommend to do a minimal amount of managing imports with only importing needed stuff/hiding stuff in this PR.

Yes, that's what I did, but again, resolving conflicts is harder when there are tons of unnecessary imports. It slows development to have to resolve these unnecessary conflicts (unnecessary since they will seldom arise if we are just a bit more disciplined about imports).

> 
> > However, it seems you prefer I revert these improvements and submit them in a separate PR that does proof/import cleanup and nothing else, so that's what I'll do.
> 
> Are you referring to a larger scale tour over the repo introducing more use of import lists or something else? If the former, I recommend to make an issue, this can be a task on the backlog to pick up (by someone) at a later date if the team wants it. Don't make a PR yet.

I already made an issue (#183) for this because I have done a fair amount of work cleaning up the code base---work that keeps getting scrapped from PRs... which is fine; I understand why PRs need to be very focused on a single task.  Still, I don't want to lose that work, so I've made a branch to keep it in, which I will keep up-to-date with master... but don't worry, I'll not make a PR for it yet. 😅

Btw, I also made an issue (#182) for automating documentation formatting, as we discussed during our 1:1... but no PR for that one either yet.

WhatisRT (approved):
I fixed a small typo & squashed everything. LGTM

PR #175, opened 2023-08-14
Don't use reduceDec in deriveComputational

UlfNorell:
`reduceDec` is very slow. It's possible it can be optimised, but it wasn't used in a critical way in `deriveComputational` so it was easier to just replace it for now.

Also, getting rid of the nested macro call to `by-derive⇔` saves a few seconds of quoting and unquoting the STS constructors.

All in all this brings down the type checking time of `Ledger.Utxo` from 44s to 13s.

### DIFF
--- a/src/Tactic/DeriveComp.agda
+++ b/src/Tactic/DeriveComp.agda
@@ -5,7 +5,9 @@ module Tactic.DeriveComp where
 open import Prelude
 open import Meta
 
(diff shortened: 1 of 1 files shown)


WhatisRT:
CI broke again, so I merged the fix to get CI to run on your PR

WhatisRT (approved):

PR #176, opened 2023-08-15
Adjust maximize buildspace

teodanciu:
# Description

To avoid the mirrors not being found when uninstalling things, we need to run `apt-get update` at the beginning. 

When uninstalling to free space, removing "large packages" seems to take the longest, but only frees about 4GBs, so I removed this action in order to save  some build time. 

# Checklist

- [ ] Self-reviewed the diff


WhatisRT (approved):
Thank you!

PR #179, opened 2023-08-16
Adding alonzo scripts to the ledger 

Ali-Hill:


src/Ledger/Prelude.agda
```diff
@@ -127,3 +129,23 @@ module Properties where
...
+                            → Injective _≡_ _≡_ f
+                            → left-unique (mapˡ f (map (λ x → x , x) S))
+mapˡ-inj-dup S f x x₁ x₂ = mapˡ-uniq (λ _ _ → x) dupIsLeftUnique x₁ x₂
+
+setToHashRel : ∀ {A B : Set} → {{Hashable A B}} -> ℙ A -> ℙ (B × A)
+setToHashRel x = mapˡ hash (map (λ x₁ → x₁ , x₁) x)
+
+setToHashMap : ∀ {A B : Set} → {{Hashable A B}} -> ℙ A → B ⇀ A
+setToHashMap x with mapˡ-inj-dup x hash hashInj
+... | ans  = _ᵐ (setToHashRel x) {{record { isLeftUnique = λ x₁ x₂ → ans x₁ x₂ }}}
```
omelkonian, line 150: This whole construction is not really necessary for the purposes of this PR; looking up a script hash (in `Ledger.ScriptValidation`) can be done directly like so:
```agda
scriptHashInTx : ScriptHash → Tx → Bool
scriptHashInTx sh tx = ⌊ any? (λ s → hash s ≟ sh) (tx .Tx.wits .TxWitnesses.scripts) ⌋
```

However, should we keep the "dup" construction (and the concrete use case that uses Hashable to pair elements with their hash), e.g. if you already have a use for it down the road to avoid people re-implementing all over again, or even without a use case in mind just to have it lying there under `Axiom.Set.Map`?
@WhatisRT WDYT?

WhatisRT: I agree it's not necessary, but it may be nice for the PDF. I don't recall this particular function in the Alonzo spec, so maybe it's a helper, but there are a couple of places where we do lookup things by their hashes, and writing `lookup h (setToHashMap s)` is much clearer than using `any?`.

That being said, I prefer to keep the set theory part separate from the ledger stuff (I want it to be its own library at some point). Maybe we could have a separate branch in the module tree for ledger-specific set theory stuff, like `Ledger.Set.X`?

omelkonian: Just to clarify, this is what would be added in the set theory:
```agda
map⦅×-dup⦆-uniq : ∀ {x : Set A} → left-unique (mapˢ ×-dup x)
map⦅×-dup⦆-uniq x y with ∈-map⁻' x | ∈-map⁻' y
... | fst , refl , _ | .fst , refl , _ = refl

mapˡ∘map⦅×-dup⦆-uniq : ∀ {S : Set A} {f : A → B}
  → Injective _≡_ _≡_ f
  → left-unique $ mapˡ f (mapˢ ×-dup S)
mapˡ∘map⦅×-dup⦆-uniq f-inj = mapˡ-uniq (λ _ _ → f-inj) map⦅×-dup⦆-uniq
```

and then we just have to instantiate this to the hashing construction under Ledger/:
```agda
setToHashRel : ∀ {A B : Set} → {{Hashable A B}} -> ℙ A -> ℙ (B × A)
setToHashRel x = mapˡ hash (map ×-dup x)

setToHashMap : ∀ {A B : Set} → {{Hashable A B}} -> ℙ A → B ⇀ A
setToHashMap x = setToHashRel x ᵐ
  where instance _ = record { isLeftUnique = mapˡ∘map⦅×-dup⦆-uniq hashInj }
```

Or is it better to not even include the `×-dup` construction in `Axiom.Set.Map`?

omelkonian:
Rebased onto #217; the full repo now type-checks (including `Foreign.*`) but there are still a lot of postulates/unfinished dedfinitions, so still a WIP.

omelkonian:
@Ali-Hill Now rebased on master (that includes my clean-up PR), feel free to base your follow-up work on scripts onto this PR.

omelkonian:
@WhatisRT Any idea why the CI is failing? It seems to build all Haskell modules successfully and then fails with no error message :S

WhatisRT:
Here you go: https://github.com/input-output-hk/formal-ledger-specifications/actions/runs/6509719705/job/17737234122?pr=179#step:10:5101

I think it's compiling things in parallel, so it still continues on errors.

WhatisRT (approved):
I found a few small things, but I decided that it'd probably be faster if I just did it myself

src/Ledger/Transaction.lagda
```diff
@@ -95,11 +95,9 @@ the transaction body are:
     ; scriptStructure = scriptStructure
     ; govParams = govParams
     }
+
   open Ledger.GovernanceActions govStructure hiding (yes; no) public
   open Ledger.Deleg             govStructure public
-
-  _ : Value → Value → Set
-  _ = _≈_
```
omelkonian: This was a unit test to make sure we have not messed up the instances in all the structures above which `open` several sub-structures, although I agree it seems arbitrary (i.e. we are missing a principles way to test that instances are properly set up).

src/Ledger/Transaction.lagda
```diff
@@ -173,16 +171,15 @@ the transaction body are:
...
   txinsScript txins utxo = txins ∩ dom (proj₁ (scriptOuts utxo))
 
   lookupScriptHash : ScriptHash → Tx → Maybe Script
-  lookupScriptHash sh tx
-    = M.map (lookupMap m)
-    $ decToMaybe
-    $ sh ∈? mapˢ proj₁ (m ˢ)
+  lookupScriptHash sh tx = case sh ∈? mapˢ proj₁ (m ˢ) of λ where
+      (no _)  → nothing
+      (yes _) → just $ lookupᵐ m sh
```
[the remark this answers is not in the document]
WhatisRT: That's a good point! I think we could also change `ifᵈ_then_else_` so that it passes the proof along as a hidden argument:
```agda
ifᵈ_then_else_ : (P : Set a) → ⦃ Dec P ⦄ → ({P} → X) → ({¬ P} → X) → X
ifᵈ P then t else f with ¿ P ¿
... | yes p = t {p}
... | no ¬p = f {¬p}
```
And then we could do this:
```agda
ifᵈ sh ∈ mapˢ proj₁ (m ˢ)
  then just $ lookupᵐ m sh
  else nothing
```
This should work, since the tactic should find hidden arguments just fine, and they are implicitly qualified here.

src/Ledger/Transaction.lagda
omelkonian: One could argue that this is precisely what definitions named `_` signify, i.e. unit tests or examples which are kind of the same thing in the dependently-typed world.

PR #192, opened 2023-08-23
Cleanup imports in module `Utxo.Properties`

williamdemeo:
# Description

This cleans up imports of one module---`Utxo.Properties`

It is based on [william/34-type-classes](https://github.com/input-output-hk/formal-ledger-specifications/tree/william/34-type-classes), so the diffs will be large at first, but will become small---limited to just the imports and private variable definitions of the `Utxo.Properties` module---once `william/34-type-classes` is merged.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Utxo.lagda
+++ b/src/Ledger/Utxo.lagda
@@ -1,4 +1,4 @@
-\section{UTxO}
+n\section{UTxO}
 \label{sec:utxo}
 
 \subsection{Accounting}
--- a/src/Ledger/Utxo.lagda
+++ b/src/Ledger/Utxo.lagda
@@ -1,4 +1,4 @@
-n\section{UTxO}
+\section{UTxO}
 \label{sec:utxo}
 
 \subsection{Accounting}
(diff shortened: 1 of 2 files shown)


WhatisRT:
There's no reason to make this dependent on a previous PR, since it touches a completely different file. This should just contain the most recent commit, and it should apply just fine when we rebase it to merge it. It also means that the order of the merging doesn't matter. Same for your other PR.

williamdemeo:
> There's no reason to make this dependent on a previous PR, since it touches a completely different file. This should just contain the most recent commit, and it should apply just fine when we rebase it to merge it. It also means that the order of the merging doesn't matter. Same for your other PR.

Not true.  Both this and the type classes PR touch `Utxo/Properties.lagda`.  Now that the type classes PR is merged into master and I have rebased this on master, there is only one file with diffs (`Utxo/Properties`).  If I had not based this on type classes, there would now be diffs that I don't want to be part of this PR.

WhatisRT:
I've pushed your other PR to show you what I mean. Now that other PR just has the changes you want and can be merged independently of this one.

williamdemeo:
> I've pushed your other PR to show you what I mean. Now that other PR just has the changes you want and can be merged independently of this one.

I still don't see what you mean.  I see you forced push my other branch, which is what I would have done after rebasing on master... so, I'm not sure what the problem is.

WhatisRT (approved):
Thank you! LGTM

PR #196, opened 2023-08-31
Fix several smaller issues

### LINKED ISSUE #143: Add script to the constitution
We want to add an optional `ScriptHash` to the constitution. If it is present, every time a transaction proposes a treasury withdrawal or protocol parameter change governance action, the transaction has to be validated by that script. We also have to add a `ScriptPurpose` for that.

### LINKED ISSUE #161: Move donations to treasury at the epoch boundary
We currently only store the donated ADA in the donations field, but we should also move it to the treasury in the EPOCH rule at the epoch boundary.

### LINKED ISSUE #174: `CERTBASE` needs to only update registered DRep epochs
Currently it just adds the DReps that vote, which lets you skip registration for free.

WhatisRT:
# Description

- Move donations into treasury (closes #161)
- Turn CC hot and cold keys into `Credential`s
- Only refresh registered DReps when voting (closes #174)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Axiom/Set/Map.agda
+++ b/src/Axiom/Set/Map.agda
@@ -231,9 +231,13 @@ module Restrictionᵐ (sp-∈ : spec-∈ A) where
   _∣_ᶜ : Map A B → Set A → Map A B
   m ∣ X ᶜ = ⊆-map (R._∣ X ᶜ) R.ex-⊆ m
 
+  -- map only values in X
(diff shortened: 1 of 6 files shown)


WhatisRT:
I forgot that the proposal policy was already done in #188. This PR should be merged after #188 because this should be rebased on top.

Soupstraw (approved):

PR #197, opened 2023-08-31
Type classes for common structures: preorders, partial orders, strict total orders

williamdemeo:
# Description

This addresses issue https://github.com/input-output-hk/formal-ledger-specifications/issues/34 for the special case of inequalities.

The main goal of this PR is to allow us to use a single operation
symbol, `_<_`, to denote the less than relation on the types `ℤ`, `ℕ`,
and the `Slot` and `Epoch` types occurring in fields of the
`EpochStructure` record type.

In fact, I could only find one instance of a special, decorated
less than symbol---namely, the symbol `<ˢ` occurrs on line 97 of
the `PPUp.lagda` file.

The new `HasStrictTotalOrder` module, and the instance for
`HasStrictTotalOrder Slot _≡_` in the `HasStrictTotalOrder/Instance`
module, allows us to write `<` on line 97 of `PPUp.lagda` (instead of `<ˢ`).

Some purely cosmetic/no-op improvements were removed from this PR, but
these should probably be included in a future PR.  In particular,

+  a couple of modules we touch here import the entire `Algebras` module
   in cases where we only really need `open import Algebras.Bundles using
   (Semiring)`.

+  Another minor cosmetic improvement would be to change the lines in
   equational reasoning proofs near the bottom of `Utxo.Properties` that
   begin with `ℤ.+_`; these can all begin with `ℤ.+` instead.

# Checklist

- [X] Self-reviewed the diff


src/Ledger/PPUp.lagda
```diff
@@ -92,7 +96,7 @@ data _⊢_⇀⦇_,PPUP⦈_ : PPUpdateEnv → PPUpdateState → Maybe Update →
   PPUpdateCurrent : let open PPUpdateEnv Γ in
     dom (pup ˢ) ⊆ dom (genDelegs ˢ)
     → All (isViableUpdate pparams) (range (pup ˢ))
-    → (slot + (2 * StabilityWindow)) <ˢ firstSlot (sucᵉ (epoch slot))
+    → (slot + (2 * StabilityWindow)) < firstSlot (sucᵉ (epoch slot))
```
williamdemeo, line 99: This single mod is the whole point of this PR and is the reason for adding the `HasStrictTotalOrder` type class.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -86,11 +86,11 @@ consumedCoinEquality {tx} {utxoState} {pp} h = let utxo = UTxOState.utxo utxoSta
     ≡⟨ ∙-homo-Coin _ _ ⟩
   coin (balance (utxo ∣ txins tx) + mint tx) + coin (inject $ depositRefunds pp utxoState tx)
     ≡⟨ cong (coin (balance (utxo ∣ txins tx) + mint tx) +_) (property _) ⟩
-  coin (balance (utxo ∣ txins tx) +ᵛ mint tx) ℕ.+ depositRefunds pp utxoState tx
+  coin (balance (utxo ∣ txins tx) + mint tx) ℕ.+ depositRefunds pp utxoState tx
```
williamdemeo, line 89: I noticed we don't need a decorated `+` here, but I don't want to make an entire PR for this small improvement.

src/Ledger/Utxo/Properties.lagda
```diff
@@ -86,11 +86,11 @@ consumedCoinEquality {tx} {utxoState} {pp} h = let utxo = UTxOState.utxo utxoSta
...
     ≡⟨ cong! (∙-homo-Coin _ _) ⟩
-  coin (balance (utxo ∣ txins tx)) ℕ.+ coin (mint tx) ℕ.+ depositRefunds pp utxoState tx
+  coin (balance (utxo ∣ txins tx)) + coin (mint tx) ℕ.+ depositRefunds pp utxoState tx
     ≡⟨ cong (λ x → cbalance (utxo ∣ txins tx) + x + depositRefunds pp utxoState tx) h ⟩
-  cbalance (utxo ∣ txins tx) ℕ.+ 0 ℕ.+ depositRefunds pp utxoState tx
+  cbalance (utxo ∣ txins tx) + 0 ℕ.+ depositRefunds pp utxoState tx
```
williamdemeo, line 93: I noticed the first `+` symbol on lines 91 and 93 do not need to be qualified, but I don't want to make an entire PR for this small improvement.

williamdemeo:
I have no idea why there are so many repeated commits in this PR. 😥

WhatisRT:
I've been looking at your inequalities PRs, and I think we should try as much as possible just to have one type class. Our orders are not necessarily total, so I think partial orders are a reasonable choice. We could choose between strict and non-strict, but it doesn't really matter since there are `Relation.Binary.Construct.NonStrictToStrict` and `Relation.Binary.Construct.StrictToNonStrict` which can convert them in either direction.

With that in mind, I think it's best to close this PR & derive a strict partial order from the partial order in your other PR instead.

williamdemeo:
> I've been looking at your inequalities PRs, and I think we should try as much as possible just to have one type class. Our orders are not necessarily total, so I think partial orders are a reasonable choice. We could choose between strict and non-strict, but it doesn't really matter since there are `Relation.Binary.Construct.NonStrictToStrict` and `Relation.Binary.Construct.StrictToNonStrict` which can convert them in either direction.
> 
> With that in mind, I think it's best to close this PR & derive a strict partial order from the partial order in your other PR instead.

@WhatisRT Okay, I'll close this and open a new one.

However, I think it would be nicer to have alternative choices allowing us to specify exactly what type of order we mean to use implicitly.  That seems safer to me.  Also, I don't like the idea of "casting" from a strict relation to a non-strict relation and back, especially if this will be happening implicitly.  That seems dangerous.

What would you think about having a single `HasOrder` module that supports the various order type-classes, as follows:

```agda
{-# OPTIONS --safe #-}

module Interface.HasOrder where

open import Level using (_⊔_ ; suc)
open import Relation.Binary using (Rel)
open import Relation.Binary.Structures

record HasPreorder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _≤_
  field
    _≤_ : Rel A ℓ₂
    isPreorder : IsPreorder {A = A} _≈_ _≤_

record HasPartialOrder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _≤_
  field
    _≤_ : Rel A ℓ₂
    isPartialOrder  : IsPartialOrder {A = A} _≈_ _≤_

record HasStrictPartialOrder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ)  : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _<_
  field
    _<_ : Rel A ℓ₂
    isStrictPartialOrder : IsStrictPartialOrder {A = A} _≈_ _<_

record HasTotalOrder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _≤_
  field
    _≤_ : Rel A ℓ₂
    isTotalOrder : IsTotalOrder {A = A} _≈_ _≤_

record HasStrictTotalOrder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ)  : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _<_
  field
    _<_ : Rel A ℓ₂
    isStrictTotalOrder : IsStrictTotalOrder {A = A} _≈_ _<_

record HasRawLeq {a ℓ} (A : Set a) : Set (a ⊔ suc ℓ) where
  infix 4 _≤_
  field _≤_ : Rel A ℓ
```

WhatisRT:
It's not less safe, since the conversions are done exactly like you'd expect. In fact, I think it's more safe: going this route it's guaranteed that `a ≤ b` iff `a < b` or `a ≈ b`, while if we allow other instances this might not always hold.

This many type classes approach won't work, since if you write `a ≤ b` it's again not clear what it should be. Does it refer to a preorder or partial order? This doesn't even work if there'd only be one possible instance, since Agda will complain even before it goes into instance search that there are multiple definitions for `_≤_`.

williamdemeo:
We've decided against having a type class for each type of inequality, so this PR is obsolete.  As a result, we cannot have a inequality type class for, e.g., the `TokenAlgebra`, since the inequality for that type is abstract so it's impossible to prove the order properties.  (Previously, we included a "raw partial order" which didn't require proof of the properties and so could be used for abstract types like the `TokenAlgebra` type.)

PR #199, opened 2023-09-01
Type classes for common structures: preorders and partial orders

williamdemeo:
# Description

**Updated 6 Sep 2023** (to reflect some of the changes requested in conversation below)

This addresses issue #34 for the special cases of preorders and partial orders.

The goal of this PR is to allow us to use

+ a single relation symbol `≤`, to denote the "less or equal" relation on each of the types `ℤ`, `ℕ`, and `Slot`---on which `≤` denotes a *partial order*---and on the `Epoch` type, on which ≤` denotes a *preorder*.

+ a single relation symbol `<` to denote the "strictly less than" relation on each of the types `ℤ`, `ℕ`, and `Slot`.

# Checklist

- [X] Self-reviewed the diff


src/Interface/HasRawPartialOrder/Instance.agda
```diff
@@ -0,0 +1,24 @@
...
+  leqNat : HasRawPartialOrder ℕ
+  leqNat = record { _≤_ = _≤ℕ_ }
+
+
+module leqTokenAlgebra (PolicyId : Set)(ta : TokenAlgebra PolicyId)  where
+
+  open TokenAlgebra ta
+
+  instance
+    leqTokenAlgebra : HasRawPartialOrder Value
```
WhatisRT, line 23: This instance should probably be in `Ledger.TokenAlgebra`

src/Interface/HasRawPartialOrder.agda
```diff
@@ -0,0 +1,15 @@
...
+
+module Interface.HasRawPartialOrder where
+
+open import Level                       using (_⊔_ ; suc)
+open import Relation.Binary             using (Rel)
+
+-- A "raw" partial order is just a relation that we expect to play the role of
+-- a partial order, but does not come with a proof that the relation satisfies
+-- any of the three properties required to be a partial order.
+record HasRawPartialOrder {a ℓ} (A : Set a) : Set (a ⊔ suc ℓ) where
```
WhatisRT, line 11: I think we should do at least one of the following two things:
- rename this record to `HasLeq` (or something similar)
- make a lawful typeclass using `IsPartialOrder` from `Relation.Binary`

The issue with having a raw partial order is that you're promising something that you could also just prove, which I think we should avoid.

src/Ledger/Prelude.agda
```diff
@@ -20,6 +20,7 @@ open import Interface.DecEq public
 open import Ledger.Interface.HasCoin public
 open import Interface.HasAdd public
 open import Interface.HasAdd.Instance public
+open import Interface.HasRawPartialOrder public
```
WhatisRT, line 23: The instance should also be exported from here

WhatisRT (changes requested):

src/Ledger/Prelude.agda
williamdemeo: That's how I did it originally, but it resulted in a name clash/ambiguity.  But okay, I'll see if I can put the instance back here and resolve the name clash.

src/Interface/HasRawPartialOrder.agda
williamdemeo: @WhatisRT, note that by eliminating the "raw partial order" type class we also eliminate the instance of `HasRawPartialOrder TokenAlgebra _≡_`, and we can't make such an instance of the new type class that requires proof of order properties, e.g., `HasPartialOrder TokenAlgebra _≡_`, since `_≤ᵗ_` is abstract.

src/Interface/HasRawPartialOrder/Instance.agda
williamdemeo: The `leqTokenAlgebra` was eliminated with the latest commit (since without the `HasRawPartialOrder` type class, there can no longer be such a `leqTokenAlgebra` instance for the Value type; see comment below).

src/Ledger/Utxo.lagda
```diff
@@ -15,7 +15,7 @@ open import Ledger.Prelude hiding (Dec₁)
 
 open import Algebra                        using (CommutativeMonoid)
 open import Data.Integer.Ext               using (posPart; negPart)
-open import Data.Nat                       using (_≤?_; _≤_)
+open import Data.Nat as ℕ                  using (_≤?_)
```
williamdemeo, line 18: When I remove `as ℕ` and change each occurrence of `ℕ.≤` to `≤`, type-checking hangs.  When I include `as ℕ` and use `ℕ.≤` instead of `≤` in the `Computational'-UTXO` proof, it type-checks fairly quickly.

williamdemeo:
I had to modify some proofs to make the type classes for inequalities work.  

Note: although we're not using the inequality for `Epoch` yet, I included a type class for preorders, since that would be required for the `Epoch` inequality (which is not antisymmetric).

src/Interface/HasOrder.agda
```diff
@@ -0,0 +1,24 @@
...
+
+open import Level using (_⊔_ ; suc)
+open import Relation.Binary using (Rel)
+open import Relation.Binary.Structures using (IsPreorder ; IsPartialOrder)
+
+record HasPreorder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
+  infix 4 _≤_
+  field
+    _≤_ : Rel A ℓ₂
+    isPreorder : IsPreorder {A = A} _≈_ _≤_
```
williamdemeo, line 13: Preorder type class is included in case we want to use it for inequality on the `Epoch` type, which is not antisymmetric.

src/Interface/HasOrder/Instance.agda
```diff
@@ -0,0 +1,37 @@
...
+open import Data.Nat.Properties      using ( ) renaming (≤-isPartialOrder to ≤ℕ-isPartialOrder)
+
+instance
+  leqInt : HasPartialOrder ℤ _≡_
+  leqInt = record { _≤_ = _≤ℤ_ ; isPartialOrder = ≤ℤ-isPartialOrder }
+
+  leqNat : HasPartialOrder ℕ _≡_
+  leqNat = record { _≤_ = _≤ℕ_ ; isPartialOrder = ≤ℕ-isPartialOrder }
+
+-- (The following no longer works since we've switched the type class from "raw"
```
williamdemeo, line 19: If we decide to include a "raw partial order" type class, we can infer `≤` for the `TokenAlgebra` type.

src/Ledger/Deleg.lagda
```diff
@@ -239,6 +239,11 @@ instance
   Computational-POOL = fromComputational' Computational'-POOL
 
 instance
+  Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
+  Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
+  Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
+  Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
+
```
williamdemeo, line 246: Had to move `Dec-⊎` here, where it is used, so that it doesn't conflict with one of the `_≤_` instances that we want Agda to use in the `PPUp.Properties` module.

src/Ledger/Epoch.agda
```diff
@@ -33,16 +34,51 @@ record EpochStructure : Set₁ where
...
+
+  ≤ˢ-isPreorder : IsPreorder _≡_ _≤ˢ_
+  ≤ˢ-isPreorder = IsPartialOrder.isPreorder ≤ˢ-isPartialOrder
+
+  ≤ˢ-isTransitive : Transitive _≤ˢ_
+  ≤ˢ-isTransitive = IsPreorder.trans ≤ˢ-isPreorder
+
+  ≤ˢ-isAntisymmetric : Antisymmetric _≡_ _≤ˢ_
+  ≤ˢ-isAntisymmetric = IsPartialOrder.antisym ≤ˢ-isPartialOrder
+
```
williamdemeo, line 79: These are included because some of them are convenient below and others may be convenient in the future.

src/Ledger/PPUp/Properties.agda
```diff
@@ -4,7 +4,7 @@ open import Ledger.Transaction
 
 module Ledger.PPUp.Properties (txs : TransactionStructure) where
 
-open TransactionStructure txs
+open TransactionStructure txs hiding (Dec-⊎)
```
williamdemeo, line 7: `Dec-⊎` clashes with an instance of `_≤_` that we want Agda to infer in this module.

src/Ledger/Utxo.lagda
```diff
@@ -271,10 +271,10 @@ instance
     case ¿ txins tx ≢ ∅
          × txins tx ⊆ dom (UTxOState.utxo s ˢ)
          × inInterval (UTxOEnv.slot Γ) (txvldt tx)
-         × minfee (UTxOEnv.pparams Γ) tx ≤ txfee tx
+         × minfee (UTxOEnv.pparams Γ) tx ℕ.≤ txfee tx
```
williamdemeo, line 274: It seems type class inference doesn't work in this computation proof.

src/Ledger/PPUp.lagda
```diff
@@ -92,6 +92,8 @@ data _⊢_⇀⦇_,PPUP⦈_ : PPUpdateEnv → PPUpdateState → Maybe Update →
   PPUpdateCurrent : let open PPUpdateEnv Γ in
     dom (pup ˢ) ⊆ dom (genDelegs ˢ)
     → All (isViableUpdate pparams) (range (pup ˢ))
+    -- I would like to remove the `ˢ` decoration from `<ˢ`, but I believe we need a *strict*
+    -- order type class for that and including such a type class was voted down.
     → (slot + (2 * StabilityWindow)) <ˢ firstSlot (sucᵉ (epoch slot))
```
williamdemeo, line 97: I don't see how or why we would use `StrictToNonStrict` or `NonStrictToStrict` to try to make inference for `<` work with only a (nonstrict) partial order type class available.

src/Ledger/PPUp/Properties.agda
```diff
@@ -55,15 +55,18 @@ instance
...
   helper : ∀ {Γ u} → Dec (Current-Property Γ u) × Dec (Future-Property Γ u) → PPUpdateState → Maybe PPUpdateState
-  helper (no _  , no  _) record { pup = pupˢ ; fpup = fpupˢ } = nothing
-  helper {u = (pup , _)} (no _  , yes _) record { pup = pupˢ ; fpup = fpupˢ } = just record { pup = pupˢ ; fpup = pup ∪ᵐˡ fpupˢ }
-  helper {u = (pup , _)} (yes _ , no  _) record { pup = pupˢ ; fpup = fpupˢ } = just record { pup = pup ∪ᵐˡ pupˢ ; fpup = fpupˢ }
-  helper (yes (_ , _ , p , _) , yes (_ , _ , p' , _)) _       = case p' p of λ ()
+  helper                  (no _                 , no _                  ) _                                     = nothing
+  helper {u = (pup , _)}  (no _                 , yes _                 ) record { pup = pupˢ ; fpup = fpupˢ }  = just record { pup = pupˢ ; fpup = pup ∪ᵐˡ fpupˢ }
+  helper {u = (pup , _)}  (yes _                , no _                  ) record { pup = pupˢ ; fpup = fpupˢ }  = just record { pup = pup ∪ᵐˡ pupˢ ; fpup = fpupˢ }
+  helper                  (yes (_ , _ , p , _)  , yes (_ , _ , p' , _)  ) _                                     = contradiction p (≤ˢ⇒¬>ˢ p')
+
```
williamdemeo, line 69: Not sure the new indentation is helpful.  If you think not, I'll revert.

src/Ledger/PPUp/Properties.agda
```diff
@@ -55,15 +55,18 @@ instance
...
   Dec-Future-Property {Γ} = ¿ _ ¿ ×-dec all?' ⦃ Dec₁-isViableUpdate {Γ} ⦄ ×-dec ¿ _ ¿ ×-dec _ ≟ _
 
 private
   helper : ∀ {Γ u} → Dec (Current-Property Γ u) × Dec (Future-Property Γ u) → PPUpdateState → Maybe PPUpdateState
-  helper (no _  , no  _) record { pup = pupˢ ; fpup = fpupˢ } = nothing
-  helper {u = (pup , _)} (no _  , yes _) record { pup = pupˢ ; fpup = fpupˢ } = just record { pup = pupˢ ; fpup = pup ∪ᵐˡ fpupˢ }
-  helper {u = (pup , _)} (yes _ , no  _) record { pup = pupˢ ; fpup = fpupˢ } = just record { pup = pup ∪ᵐˡ pupˢ ; fpup = fpupˢ }
-  helper (yes (_ , _ , p , _) , yes (_ , _ , p' , _)) _       = case p' p of λ ()
+  helper                  (no _                 , no _                  ) _                                     = nothing
+  helper {u = (pup , _)}  (no _                 , yes _                 ) record { pup = pupˢ ; fpup = fpupˢ }  = just record { pup = pupˢ ; fpup = pup ∪ᵐˡ fpupˢ }
```
williamdemeo, line 66: Not sure the new indentation is helpful. If you think not, I'll revert.

src/Interface/HasOrder.agda
```diff
@@ -0,0 +1,24 @@
...
+
+-- Remove the following; otherwise, we can't resolve _≤_ for ℕ type in GovernanceActions.
+-- open HasPreorder ⦃ ... ⦄ public
+
+record HasPartialOrder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
+  infix 4 _≤_
+  field
+    _≤_ : Rel A ℓ₂
+    isPartialOrder  : IsPartialOrder {A = A} _≈_ _≤_
+
```
williamdemeo, line 23: Add `open NonStrictToStrict (_≤_) using (_<_) public` here.

src/Ledger/PPUp.lagda
williamdemeo: I tried this approach, but a problem occurs when a type (e.g., `ℕ`) has it's own strict partial order; that order will conflict with the strict partial order derived from the inferred partial order, so we must override the inferred `_<_` with a decorated symbol (e.g., `_<ℕ_`)... and now we're back to needing annotated relation symbols... even worse, we're now annotating relations defined on primitive types, instead of annotating the relations of special "user defined" types.

I really think it makes more sense, would be easier and cleaner, to include a separate `StrictPartialOrder` type class in the HasOrder module.

williamdemeo:
I prefer #221 to this one since #221 does more/better inequality inference and #221 is based off the latest master so has no conflicts.

Because of Ulf's new computation proofs, resolving conflicts of this PR and rebasing on the latest master will be quite challenging.  That's why I created the new PR #221

williamdemeo:
We've merged an alternative PR making this one obsolete.

PR #200, opened 2023-09-01
Fix that governance actions were removed immediately

### LINKED ISSUE #180: Delay enactment via `RatifyState`
Currently there's an `EnactState` that rolls over, but we really need a `RatifyState`. Instead of applying the changes we get from `RATIFY` directly, we need to store the result and apply it at the next epoch boundary.

WhatisRT:
# Description

They are removed at the same epoch where they come into effect (via rollover of `EnactState`). Fixes #180.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Chain.lagda
+++ b/src/Ledger/Chain.lagda
@@ -94,9 +94,12 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
 
       retired = retiring ⁻¹ e
 
-      certState' = record certState {
-        pState = record pState { pools = pools ∣ retired ᶜ ; retiring = retiring ∣ retired ᶜ };
-        dState = record dState { rewards = rewards ∪⁺ refunds } }
+      certState' = record certState
+        { pState = record pState { pools = pools ∣ retired ᶜ ; retiring = retiring ∣ retired ᶜ }
+        ; dState = record dState { rewards = rewards ∪⁺ refunds } }
+        ; gState = if not (null govSt')
+                     then gState
+                     else record gState { dreps = mapValues sucᵉ (GState.dreps gState) }
       utxoSt' = record utxoSt
         { fees = 0
         ; deposits = deposits ∣ map (proj₁ ∘ proj₂) removedGovActions ᶜ
--- a/src/Ledger/Chain.lagda
+++ b/src/Ledger/Chain.lagda
@@ -34,7 +34,8 @@ record NewEpochState : Set where
   field lastEpoch  : Epoch
         acnt       : Acnt
         ls         : LState
-        es esFut   : EnactState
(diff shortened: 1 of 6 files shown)


src/Ledger/Chain.lagda
```diff
@@ -70,42 +71,48 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
...
       acnt' = record acnt { treasury = Acnt.treasury acnt + UTxOState.fees utxoSt + getCoin unclaimed }
     in
     e ≡ sucᵉ lastEpoch
     → record { currentEpoch = e ; GState gState ; NewEpochEnv Γ }
-        ⊢ ⟦ es  , [] , [] , false ⟧ʳ ⇀⦇ govSt ,RATIFY⦈ ⟦ es' , govSt' , removed , d ⟧ʳ
+        ⊢ ⟦ es , ∅ , false ⟧ʳ ⇀⦇ govSt' ,RATIFY⦈ fut'
     -- TODO: remove keys that aren't in the CC from the hot key map
     ────────────────────────────────
-    Γ ⊢ nes ⇀⦇ e ,NEWEPOCH⦈ ⟦ e , acnt' , ls' , es , es' ⟧ⁿᵉ
+    Γ ⊢ nes ⇀⦇ e ,NEWEPOCH⦈ ⟦ e , acnt' , ls' , es , fut' ⟧ⁿᵉ
```
Soupstraw, line 115: Currently in the implementation we store just the `EnactState` inside `NewEpochState`

Soupstraw: I think it makes more sense to do it like that here as well, since we only need to use the `removed` set inside `NEWEPOCH`

Soupstraw: same goes for `delay`

Soupstraw: what I meant is that in the EpochState we store an EnactState where in the spec we have a RatifyState. When the EPOCH rule is called, we create an ephemeral RatifyState inside the EPOCH rule to pass to RATIFY and then use the `removed` field to remove the governance actions that need to be removed. Then we move the new `EnactState` to `EpochState` and forget about the rest of the `RatifyState` once we're done executing the EPOCH rule.

WhatisRT: Yes, that's what used to be the case. The issue is that to know the value of `removed`, we have to do the entire `RATIFY` computation which we can only do once we've done the incremental stake distribution calculation. So we can't look at `removed` immediately, we have to wait until the next epoch boundary to use it. So we can either add a new field (say `futRemoved`) that stores them for one epoch, or we just store the entire result of `RATIFY` (which is what I did here).

Soupstraw (approved):
LGTM

PR #206, opened 2023-09-11
Make `RATIFY` total

### LINKED ISSUE #194: Delay GAs that can't be enacted
Some GAs can only be enacted if some extra conditions are met:
- `NewCC` if the members to be removed are actually in the CC
- `Withdrawal` if there's enough money in the treasury

WhatisRT:
# Description

This is important since we can get stuck on an epoch boundary otherwise. This also moves the treasury from `EnactState` into `EnactEnv`, which is equivalent but cleaner.

Closes #194.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/GovernanceActions.lagda
+++ b/src/Ledger/GovernanceActions.lagda
@@ -279,8 +279,9 @@ protocol version, protocol parameters, withdrawals from treasury, and treasury b
 {\small
 \begin{code}
 record EnactEnv : Set where
-  constructor ⟦_⟧ᵉ
-  field gid  : GovActionID
+  constructor ⟦_,_⟧ᵉ
+  field gid       : GovActionID
+        treasury  : Coin
 
 record EnactState : Set where
   field cc            : HashProtected (Maybe (Credential ⇀ Epoch × ℚ))
@@ -288,7 +289,6 @@ record EnactState : Set where
         pv            : HashProtected ProtVer
         pparams       : HashProtected PParams
         withdrawals   : RwdAddr ⇀ Coin
-        treasury      : Coin
 \end{code}
 } %% end small
 \caption{Enactment types}
@@ -308,7 +308,7 @@ private variable
   h : PPHash
   v : ProtVer
   wdrl : RwdAddr ⇀ Coin
-  newTreasury : Coin
+  t : Coin
   gid : GovActionID
 
 instance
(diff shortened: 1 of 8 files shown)


src/Ledger/GovernanceActions.lagda
```diff
@@ -324,22 +324,31 @@ It represents how the \agdaboundEnactState changes when a specific governance ac
...
-  Enact-PParams   : ⟦ gid ⟧ᵉ ⊢ s ⇀⦇ ChangePParams up ,ENACT⦈
+
+  Enact-NoConf    : ⟦ gid , t ⟧ᵉ ⊢ s ⇀⦇ NoConfidence ,ENACT⦈  record s { cc = nothing , gid }
+
+  Enact-NewComm   :
+    let old = maybe proj₁ ∅ᵐ (proj₁ (EnactState.cc s))
+    in rem ⊆ dom (old ˢ)
+    ────────────────────────────────
+    ⟦ gid , t ⟧ᵉ ⊢ s ⇀⦇ NewCommittee new rem q ,ENACT⦈
+      record s { cc = just ((new ∪ᵐˡ old) ∣ rem ᶜ , q) , gid }
```
Soupstraw, line 335: Should we also check that the intersection of old and new is empty?

Soupstraw: Right now it looks like it's possible to add and remove the same CC in one transaction

Soupstraw: or why do we need to check this at all?

WhatisRT: There are two ways to reelect someone in the current design: put them in `old` and `new`, or just in `new`. Putting them in `old` and `new` is slightly different since it only works if someone is already on the CC, while the other option always works.

Since they are slightly different I'm inclined to keep both options, but we could also prevent the one with an intersection.

WhatisRT: Ah, sorry, putting someone in `old` and `new` always removes that member, so there's no point in putting them in `new`. Yeah, then we should prevent non-trivial intersections.

WhatisRT: Actually, you're making a good point with whether the `rem ⊆ dom (old ˢ)` check is necessary. Since this GA is hash protected, it's already bound to updating one particular state to another, so all this check does is that if the community votes for something that includes some garbage in `rem` it gets delayed forever, which is arguably bad. If we add sanity checks for proposals (which we might), then that check really needs to go there. So I'll remove it from here.

Soupstraw (approved):
LGTM

PR #221, opened 2023-09-15
Type classes for common structures: orders

williamdemeo:
**Purpose**. allow us to use a generic (undecorated) relation symbols, like `<` and `≤`, to denote the "less" or "less or equal" relations over both primitive "built-in" types like ℤ, ℕ, and our "user-defined" types, like `Slot`, `Epoch`, etc.

**Issue**. This addresses issue #34 for the special cases of preorders and partial orders.

That is, we want to be able to use

+ a single relation symbol `≤`, to denote the "less or equal" relation on each of the types ℤ, ℕ, and Slot---on which ≤ denotes a partial order---and on the Epoch type, on which `≤` denotes a preorder.

+ a single relation symbol `<` to denote the "strictly less than" relation on each of the types ℤ, ℕ, and Slot.

# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Interface/Decidable/Instance.agda
+++ b/src/Interface/Decidable/Instance.agda
@@ -50,10 +50,10 @@ instance
   DecEq⇒Dec : ⦃ DecEq X ⦄ → {x y : X} → Dec (x ≡ y)
   DecEq⇒Dec ⦃ record { _≟_ = _≟_ } ⦄ {x} {y} = x ≟ y
 
-  -- Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
-  -- Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
-  -- Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
-  -- Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
+  Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
+  Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
+  Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
+  Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
 
   Dec-≤ : ∀ {n m} → Dec (n ≤ m)
   Dec-≤ = Decidable²⇒Dec _≤?_
--- a/src/Interface/Decidable/Instance.agda
+++ b/src/Interface/Decidable/Instance.agda
@@ -50,10 +50,10 @@ instance
   DecEq⇒Dec : ⦃ DecEq X ⦄ → {x y : X} → Dec (x ≡ y)
   DecEq⇒Dec ⦃ record { _≟_ = _≟_ } ⦄ {x} {y} = x ≟ y
 
-  Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
-  Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
-  Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
-  Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
+  -- Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
+  -- Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
+  -- Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
+  -- Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
 
   Dec-≤ : ∀ {n m} → Dec (n ≤ m)
   Dec-≤ = Decidable²⇒Dec _≤?_
--- a/src/Interface/HasOrder.agda
+++ b/src/Interface/HasOrder.agda
@@ -1,0 +1,85 @@
+{-# OPTIONS --safe #-}
+
+module Interface.HasOrder where
+
+open import Data.Product using (_×_; _,_; proj₁; proj₂)
+open import Data.Sum using (_⊎_; inj₁; inj₂; [_,_])
+open import Function using (mk⇔; case_of_; _⇔_)
+open import Level using (_⊔_; suc)
+open import Relation.Binary using (Rel)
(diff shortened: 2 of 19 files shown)


src/Ledger/Utxo.lagda
```diff
@@ -246,9 +248,9 @@ data _⊢_⇀⦇_,UTXO⦈_ where
           donations     = UTxOState.donations s
       in
     txins tx ≢ ∅                           → txins tx ⊆ dom (utxo ˢ)
-    → inInterval slot (txvldt tx)          → minfee pp tx ≤ txfee tx
+    → inInterval slot (txvldt tx)          → minfee pp tx ℕ.≤ txfee tx
     → consumed pp s tx ≡ produced pp s tx  → coin (mint tx) ≡ 0
-    → txsize tx ≤ maxTxSize pp
+    → txsize tx ℕ.≤ maxTxSize pp
```
williamdemeo, line 253: (Agda couldn't resolve the overloaded leq symbol in this case.)

src/Ledger/Utxo.lagda
```diff
@@ -264,20 +266,20 @@ instance
     case ¿ txins tx ≢ ∅
          × txins tx ⊆ dom (UTxOState.utxo s ˢ)
          × inInterval (UTxOEnv.slot Γ) (txvldt tx)
-         × minfee (UTxOEnv.pparams Γ) tx ≤ txfee tx
+         × minfee (UTxOEnv.pparams Γ) tx ℕ.≤ txfee tx
```
williamdemeo, line 269: (Agda couldn't resolve the overloaded leq symbol in this case.)

src/Ledger/Utxo.lagda
```diff
@@ -264,20 +266,20 @@ instance
...
   Computational'-UTXO .Computational'.completeness Γ s tx s' h@(UTXO-inductive q₀ q₁ q₂ q₃ q₄ q₅ q₆)
     with  ¿ txins tx ≢ ∅
           × txins tx ⊆ dom (UTxOState.utxo s ˢ)
           × inInterval (UTxOEnv.slot Γ) (txvldt tx)
-          × minfee (UTxOEnv.pparams Γ) tx ≤ txfee tx
+          × minfee (UTxOEnv.pparams Γ) tx ℕ.≤ txfee tx
           × consumed (UTxOEnv.pparams Γ) s tx ≡ produced (UTxOEnv.pparams Γ) s tx
           × coin (mint tx) ≡ 0
-          × txsize tx ≤ maxTxSize (UTxOEnv.pparams Γ) ¿
+          × txsize tx ℕ.≤ maxTxSize (UTxOEnv.pparams Γ) ¿
```
williamdemeo, line 282: (Agda couldn't resolve the overloaded leq symbol in this case.)

src/Interface/HasOrder.agda
```diff
@@ -0,0 +1,40 @@
+{-# OPTIONS --safe #-}
+
+module Interface.HasOrder where
+
+open import Level using (_⊔_; suc)
+open import Relation.Binary using (Rel)
+open import Relation.Binary.Definitions using (Decidable)
+open import Relation.Binary.Structures using (IsPreorder; IsPartialOrder; IsStrictTotalOrder)
+
+record HasPreorder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
```
WhatisRT, line 10: The problem is that now strict and non-strict orders are completely independent. If you want orders where the two can be defined independently, you need to relate them with a theorem. For example, you can use this:

```agda

record HasPreorder {a ℓ ℓ₂} (A : Set a) (_≈_ : Rel A ℓ) : Set (a ⊔ ℓ ⊔ suc ℓ₂) where
  infix 4 _≤_
  field
    _≤_ _<_    : Rel A ℓ₂
    isPreorder : IsPreorder _≈_ _≤_
    irrefl     : Irreflexive _≈_ _<_
    ≤⇔<∨≡      : ∀ {a b} → a ≤ b ⇔ (a < b ⊎ a ≈ b)

module _ {a ℓ} (A : Set a) (_≈_ : Rel A ℓ) (_≤_ : Rel A ℓ) (isPreorder : IsPreorder _≈_ _≤_)
         (_≈?_ : ∀ a b → (Dec (a ≈ b))) where
  open import Relation.Binary.Construct.NonStrictToStrict _≈_ _≤_

  hasPreorderFromNonStrict : HasPreorder A _≈_
  hasPreorderFromNonStrict = record
    { _≤_        = _≤_
    ; _<_        = _<_
    ; isPreorder = isPreorder
    ; irrefl     = <-irrefl
    ; ≤⇔<∨≡      = λ {a b} → mk⇔
      (λ a≤b → case a ≈? b of λ where (yes p) → inj₂ p ; (no ¬p) → inj₁ (a≤b , ¬p))
      λ where (inj₁ a<b) → proj₁ a<b ; (inj₂ a≈b) → IsPreorder.reflexive isPreorder a≈b
    }
```

This means it's easy to define `HasPreorder` instances because you can just use `hasPreorderFromNonStrict`, but if you really need separate definitions for the two orders you can supply them & prove that equivalence.

src/Interface/Decidable/Instance.agda
```diff
@@ -48,10 +48,8 @@ instance
   DecEq⇒Dec : ⦃ DecEq X ⦄ → {x y : X} → Dec (x ≡ y)
   DecEq⇒Dec ⦃ record { _≟_ = _≟_ } ⦄ {x} {y} = x ≟ y
 
-  Dec-⊎ : ∀ {a b} {A : Set a} {B : Set b} → ⦃ Dec A ⦄ → ⦃ Dec B ⦄ → Dec (A ⊎ B)
-  Dec-⊎ ⦃ yes p ⦄ ⦃ _     ⦄ = yes (inj₁ p)
-  Dec-⊎ ⦃ no _  ⦄ ⦃ yes q ⦄ = yes (inj₂ q)
-  Dec-⊎ ⦃ no ¬p ⦄ ⦃ no ¬q ⦄ = no λ { (inj₁ p) → ¬p p; (inj₂ q) → ¬q q }
+  -- Moved `Dec-⊎` to where it is used (in `Deleg` module); otherwise it conflicts with
```
WhatisRT, line 51: I have no idea what error you're getting, but you should fix that error instead of moving a general purpose definition into a special purpose file.

WhatisRT (changes requested):

src/Interface/HasOrder.agda
williamdemeo: [Here](https://input-output-rnd.slack.com/archives/C02HDH9BVU1/p1695865566318169?thread_ts=1695823291.359499&cid=C02HDH9BVU1) I explain why we should not add the `≤⇔<∨≡` field to the `HasPreorder` record (unless we believe there's something wrong with the definitions of strict and nonstrict orders on the `Epoch` type).

WhatisRT: This is a slack link, which isn't public, so there's no point in adding it here.

If there is some order where you can't make an instance of this record for that's fine, but in the example you've made on slack you can still make an instance by simply using `hasPreorderFromStrict`. It seems like the real problem you're having is that `EpochStructure` defines everything in terms of `_<ˢ_`, but there's no reason not to change that.

williamdemeo:
Yes, that's what I did in my latest commit.  I've rewritten everything in
terms of less or equal and it all works now except for one computation
proof that I've just commented out until I can get it working again.

Re. the slack link: that was for you... but okay, I'll avoid adding
non-public links to comments on GitHub in future.

src/Interface/HasOrder.agda
williamdemeo: Well, the slack link was for you... but okay, I'll avoid adding
non-public links to comments on GitHub in future.

Your suggestion is interesting... Correct me if I'm wrong, but I believe you're suggesting that I delete the existing definition of `_≤ᵉ_` and let it be defined by `hasPreorderFromStrict`, which will take care of proving `a ≤ᵉ b ⇔ (a <ᵉ b ⊎ a ≡ b)`.  However, I claim that for the Epoch type the "only if" direction of that equivalence is false... so something about this shouldn't work.

In any case, re. changing `EpochStructure` to not define everything in terms of `_<ˢ_`, I already did that in my latest commit (as I proposed to do in the comments on Slack).  I've rewritten everything in terms of `_≤ˢ_`  and it all works now except for one computation proof that I just commented out until I can get it working again.

src/Ledger/Utxo.lagda
williamdemeo: @omelkonian's latest commit resolves this issue.

WhatisRT (approved):

Issue #212, opened 2023-09-13
GOV-Vote rule irregularities


UlfNorell:
In the `Vote` step, the `ast` (`GovActionState`) is not determined by any of the inputs:
```agda
  GOV-Vote : ∀ {x k ast} → let open GovEnv Γ in
    (aid , ast) ∈ setFromList s
    → canVote pparams (action ast) role
    ────────────────────────────────
    (Γ , k) ⊢ s ⇀⦇ inj₁ record { gid = aid ; role = role ; credential = cred ; vote = v ; anchor = x } ,GOV'⦈
              addVote s aid role cred v
```
In case the same `aid` appears several time in the state this could be a problem. This should never happen but there is nothing in the types that prevents it.

At first I thought this made the rule non-computational, but looking at the details I'm not sure this is true. The `addVote` function updates all instances of the given `aid`, so even if you pick different ones you get the same state.

It's still a bit fishy since you could have two occurrences of `aid` one where you `canVote` and one where you can't, and we'll update both in the rule.

A simple thing we could do is add a precondition `GOV-Vote` requiring `aid` to be unique. More work (but arguably nicer) would be to change the type of the state to something that guarantees left-uniqueness, keeping in mind that the order of the entries is important.

UlfNorell:
But the `compute` function can't be `nothing` if at least one of the `ast`s `canVote` (since in that case there is a valid `STS` value), so the only possible `compute` function is the one that checks all copies of `aid` and succeeds if any of them `canVote`.

WhatisRT:
Now that you've implemented it, obviously this is `Computational`. Just to remember this in the future, since the hash of a transaction goes into the `aid`, there won't be any clashes in practice. But it may still be the case that we'll run into this issue again once we try proving some properties.

PR #223, opened 2023-09-19
Computational instance for GOV' rule

UlfNorell:
This doesn't do anything about #212, the existing definition of the rule *is* computational even without enforcing unique GovActionIds.

Question: This defines `any⇔∃` locally with the computational instance, but it feels like something that should either already be in the standard library (where I couldn't find it), or be defined in some more reasonable place in our repo.


WhatisRT:
I found it, it's `Any↔` from `Data.List.Membership.Propositional.Properties`. However, it's still in the old function hierarchy, so I had to convert it to the new one. That conversion function is exactly the reason why we're not on any released version of the stdlib btw.

WhatisRT (approved):

PR #224, opened 2023-09-26
Fix broken build from changed GOV' rule

UlfNorell:
There were conflicts between #223 and 8613454.

Also fixed a bug in the new precondition added to rule.


WhatisRT (approved):
Thank you! You made your PR just after I started working on this, but it least I saw it before I've finished it

PR #225, opened 2023-09-26
Remove the old Computational in favour of the new representation

### LINKED ISSUE #211: Should we use `Computational'` everywhere instead of `Computational`?
Since we now use `Computational'` for defining all the instances, it may be worth switching everything to `Computational'`. `Computational` is still at least good for presentation, but we can look into replacing it everywhere else.

UlfNorell:
Also
- make SS⇒BS more amenable to instance search
- cleaned up Computational instance for LEDGER

Fixes #211

### DIFF
--- a/src/Interface/ComputationalRelation.agda
+++ b/src/Interface/ComputationalRelation.agda
@@ -22,22 +22,12 @@ private variable
 
 module _ (STS : C → S → Sig → S → Set) where
 
-  record Computational : Set where
-    constructor MkComputational
-    field
-      compute : C → S → Sig → Maybe S
-      ≡-just⇔STS : compute c s sig ≡ just s' ⇔ STS c s sig s'
-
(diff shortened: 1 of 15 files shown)


UlfNorell:
@WhatisRT this is now rebased on master

omelkonian:
Oops, I broke it again..

omelkonian:
OK the error makes sense, it's complaining that `Tactic.DeriveComp` is not used anymore (the check done by `checkTypechecked.sh`). I will push a quick fix that imports it in `Ledger.Prelude` for now.

omelkonian (approved):
LGTM; I've just refactored the computational proof in `Ledger.Properties` to avoid code duplication and removed obsolete imports to `Tactic.DeriveComp` (but we still need a dummy import in `Ledger.Prelude` to keep it around past CI checks).

PR #226, opened 2023-09-28
Introduce (Bi)functor interface

omelkonian:
Provides a uniform way (i.e. typeclass) to perform (bi)functorial actions, rather than having to remember the names of each instantiation/module.


WhatisRT:
Thank you, that's a great addition! What do you think about renaming `fmap` to `map`? Obviously it has name clashes everywhere, but since we're already going in the direction of putting `using` on our imports that shouldn't really matter.

WhatisRT (approved):
LGTM

PR #228, opened 2023-10-02
PDF cleanup

WhatisRT:
# Description

Cleanup I did for the talk at the R&D seminar

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Chain.lagda
+++ b/src/Ledger/Chain.lagda
@@ -75,15 +75,14 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
       removedGovActions = flip concatMapˢ removed λ (gaid , gaSt) →
         mapˢ (GovActionState.returnAddr gaSt ,_)
              ((deposits ∣ ❴ GovActionDeposit gaid ❵) ˢ)
-      govActionReturns = aggregate₊ $
-        mapˢ (λ (a , _ , d) → a , d) removedGovActions , finiteness _
+      govActionReturns = aggregate₊ $ mapˢ (λ (a , _ , d) → a , d) removedGovActions ᶠˢ
 
       es        = record esW { withdrawals = ∅ᵐ }
       retired   = retiring ⁻¹ e
       refunds   = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ)
       unclaimed = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ) ᶜ
 
-      govSt' = filter (¬? ∘ (_∈? mapˢ proj₁ removed) ∘ proj₁) govSt
+      govSt' = filter (λ x → ¿ ¬ proj₁ x ∈ mapˢ proj₁ removed ¿) govSt
 
       gState' = record gState { ccHotKeys = ccHotKeys ∣ ccCreds (es .EnactState.cc) }
 
@@ -120,6 +119,7 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
 \end{figure*}
 
 \begin{code}[hide]
+-- TODO: do we still need this for anything?
 maybePurpose : DepositPurpose → (DepositPurpose × Credential) → Coin → Maybe Coin
 maybePurpose prps (prps' , _) c with prps ≟ prps'
 ... | yes _ = just c
--- a/src/Ledger/Chain.lagda
+++ b/src/Ledger/Chain.lagda
@@ -79,10 +79,10 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
 
       es        = record esW { withdrawals = ∅ᵐ }
       retired   = retiring ⁻¹ e
-      refunds   = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ)
-      unclaimed = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ) ᶜ
+      refunds   = govActionReturns ∪⁺ trWithdrawals ∣ dom rewards
+      unclaimed = govActionReturns ∪⁺ trWithdrawals ∣ dom rewards ᶜ
 
-      govSt' = filter (λ x → ¿ ¬ proj₁ x ∈ mapˢ proj₁ removed ¿) govSt
+      govSt' = filter (λ x → ¿ proj₁ x ∉ mapˢ proj₁ removed ¿) govSt
 
       gState' = record gState { ccHotKeys = ccHotKeys ∣ ccCreds (es .EnactState.cc) }
 
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -20,6 +20,7 @@ module Ledger.Gov (gs : _) (open GovStructure gs hiding (epoch)) where
 open import Ledger.GovernanceActions gs
 \end{code}
 \begin{figure*}[h]
+\emph{Derived types}
 \begin{code}
 record GovActionState : Set where
   field votes       : (GovRole × Credential) ⇀ Vote
@@ -37,6 +38,15 @@ record GovEnv : Set where
         epoch    : Epoch
         pparams  : PParams
 \end{code}
+\emph{Transition relation types}
+\begin{code}[hide]
+data
+\end{code}
+\begin{code}
+  _⊢_⇀⦇_,GOV'⦈_ : GovEnv × ℕ → GovState → GovVote ⊎ GovProposal → GovState → Set
+
+_⊢_⇀⦇_,GOV⦈_ : GovEnv → GovState → List (GovVote ⊎ GovProposal) → GovState → Set
+\end{code}
 \begin{code}[hide]
 open GovActionState
 
@@ -51,21 +61,23 @@ private variable
   addr : RwdAddr
   a : GovAction
   prev : NeedsHash a
+  k : ℕ
 
 private -- FIXME: this should be part of a typeclass
   _<_ : Epoch → Epoch → Set
   a < b = a ≤ b × a ≢ b
-\end{code}
-\begin{code}
+
 -- could be implemented using a function of type:
 --   ∀ {a} {A : Set a} → (A → Maybe A) → List A → List A
 modifyMatch : ∀ {a} {A : Set a} → (A → Bool) → (A → A) → List A → List A
 modifyMatch P f = map (λ x → if P x then f x else x)
-
+\end{code}
+\emph{Functions used in the GOV rules}
+\begin{code}
 addVote : GovState → GovActionID → GovRole → Credential → Vote → GovState
 addVote s aid r kh v =
   modifyMatch
-    (λ x → ⌊ aid ≟ proj₁ x ⌋)
+    (λ (x , _) → aid ≡ᵇ x)
     (λ (gid , s') → gid , record s' { votes = insert (votes s') (r , kh) v }) s
 
 addAction : GovState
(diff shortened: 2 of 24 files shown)


src/Ledger/Ratify.lagda
```diff
@@ -204,7 +204,7 @@ mostStakeDRepDist-∅ {dist} = suc (Σᵐᵛ[ x ← dist ᶠᵐ ] x) , Propertie
 
 ∃topNDRepDist : ∀ {n dist} → lengthˢ (dist ˢ) ≥ n → n > 0
                 → ∃[ c ] lengthˢ (mostStakeDRepDist dist c ˢ) ≥ n
-                       × lengthˢ (mostStakeDRepDist dist (suc c) ˢ) < n
+                       × lengthˢ (mostStakeDRepDist dist (suc c) ˢ) Data.Nat.< n
```
WhatisRT, line 207: Since we don't have a typeclass for `_<_` yet and I wanted to have a `_<_` in the PDF (which compares epochs) I had to change this to be fully qualified (and there's a hacky `_<_` for epochs later in this file).

src/Ledger/GovernanceActions.lagda
```diff
@@ -86,34 +89,45 @@ Figure~\ref{defs:governance} defines several data types used to represent govern
   to non-backwards compatible updates of a network. In Cardano, we formalize the definition slightly more by calling any upgrade that
   would lead to \emph{more blocks} being validated a ``hard fork'' and force nodes to comply with the new protocol version, effectively
   obsoleting nodes that are unable to handle the upgrade.}
-\begin{code}[hide]
+\begin{figure*}[h]
+\begin{code}
+⟨_,_,_⟩ : Maybe ℚ → Maybe ℚ → Maybe ℚ → GovRole → Maybe ℚ
+⟨ q₁ , q₂ , q₃ ⟩ = λ { CC → q₁ ; DRep → q₂ ; SPO → q₃ }
```
omelkonian, line 95: This should be private, in this case just put it under `threshold`'s `where`.

Also the notation as a tuple does not seem very intuitive in this case, which might decrease the readibility of the "voting matrix" below (which I believe to be the central piece of interest in this page).

How about a notation that actually mentions which role is voting?
```agda
CC∶ q₁ DRep∶ q₂ SPO∶ q₃ = λ { CC → q₁ ; DRep → q₂ ; SPO → q₃ }
```
The signature is of no interest and could be hidden (also make it polymorphic, no need for being specific here):
```agda
CC∶_DRep∶_SPO∶_ : ∀ {a} {A : Set a} → A → A → A → GovRole → A
```

src/Ledger/Gov.lagda
```diff
@@ -89,7 +89,7 @@ data _⊢_⇀⦇_,GOV'⦈_ : GovEnv × ℕ → GovState → GovVote ⊎ GovPropo
     actionWellFormed a ≡ true
```
omelkonian, line 89: Since this is a PDF cleanup, line 78 above  (the type signature of the relation) overflows to the right of the page.

src/Ledger/GovernanceActions.lagda
```diff
@@ -23,6 +23,9 @@ open import Tactic.Derive.DecEq
 module Ledger.GovernanceActions (gs : _) (open GovStructure gs) where
 
 2ℚ = 1ℚ Data.Rational.+ 1ℚ
+
+maximum : ℙ ℚ → ℚ
+maximum x = foldl Data.Rational._⊔_ 0ℚ (proj₁ $ finiteness x)
```
omelkonian, line 28: Maybe add a note that we should use a more generic version of this in the future, i.e.
1. `maximum` defined over any (decidably) ordered type:
```agda
module _ {A : Type ℓ} ⦃ _ : Ord A ⦄ ⦃ _ : DecOrd A ⦄ where

min max : Op₂ A
min x y = if ⌊ x ≤? y ⌋ then x else y
max x y = if ⌊ y ≤? x ⌋ then x else y

minimum maximum : A → List A → A
minimum = foldl min
maximum = foldl max
```
2. `maximumˢ` defined for finite sets by `maximum ∘ proj₁ ∘ finiteness`

src/Ledger/GovernanceActions.lagda
```diff
@@ -124,28 +138,12 @@ module _ (pp : PParams) (ccThreshold' : Maybe ℚ) where
...
-    (TreasuryWdrl _)
-               → λ { CC → vote ccThreshold ; DRep → vote P6     ; SPO → noVote }
-    Info       → λ { CC → vote 2ℚ          ; DRep → vote 2ℚ     ; SPO → vote 2ℚ }
-
 -- TODO: this doesn't actually depend on PParams so we could remove that argument,
 --       but we don't have a default ATM
 canVote : PParams → GovAction → GovRole → Set
 canVote pp a r = Is-just (threshold pp nothing a r)
 \end{code}
+\end{figure*}
```
omelkonian, line 146: Please add a caption (do not exceed one line, there is minimal space left in this page) and `\label{fig:..}` to be consistent with the rest of the document.

src/Ledger/Chain.lagda
```diff
@@ -75,15 +75,14 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
...
 
       es        = record esW { withdrawals = ∅ᵐ }
       retired   = retiring ⁻¹ e
-      refunds   = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ)
-      unclaimed = govActionReturns ∪⁺ trWithdrawals ∣ dom (rewards ˢ) ᶜ
+      refunds   = govActionReturns ∪⁺ trWithdrawals ∣ dom rewards
+      unclaimed = govActionReturns ∪⁺ trWithdrawals ∣ dom rewards ᶜ
 
-      govSt' = filter (¬? ∘ (_∈? mapˢ proj₁ removed) ∘ proj₁) govSt
+      govSt' = filter (λ x → ¿ ¬ proj₁ x ∈ mapˢ proj₁ removed ¿) govSt
```
omelkonian, line 85: Please use `_∉_` here.

src/Ledger/Chain.lagda
```diff
@@ -120,24 +119,24 @@ data _⊢_⇀⦇_,NEWEPOCH⦈_ : NewEpochEnv → NewEpochState → Epoch → New
...
+--   → x ≡ prps
+-- maybePurpose-prop {prps = prps} {x} {y} _ xy∈dom with to dom∈ xy∈dom
+-- ... | z , ∈mmwk with prps ≟ x | ∈-mapMaybeWithKey {f = maybePurpose prps} ∈mmwk
+-- ... | yes refl | _ = refl
+
+-- filterPurpose : DepositPurpose → (DepositPurpose × Credential) ⇀ Coin → Credential ⇀ Coin
+-- filterPurpose prps m = mapKeys proj₂ (mapMaybeWithKeyᵐ (maybePurpose prps) m)
+--   {λ where x∈dom y∈dom refl → cong (_, _)
+--                             $ trans (maybePurpose-prop {prps = prps} m x∈dom)
+--                             $ sym   (maybePurpose-prop {prps = prps} m y∈dom)}
```
omelkonian, line 139: These still type-check, so if you have future use for them please keep them un-commented. Otherwise, you can delete them properly.

src/Ledger/Ratify.lagda
omelkonian: Wouldn't it be better to add hidden `where` clauses that import/define the appropriate for each of the two uses (`Nat` and `Epoch`)? (until the "orders" PR #34 gets merged, that is), e.g. 
```agda
\begin{code}
∃topNDRepDist : ∀ {n dist} → lengthˢ (dist ˢ) ≥ n → n > 0
                → ∃[ c ] lengthˢ (mostStakeDRepDist dist c ˢ) ≥ n
                       × lengthˢ (mostStakeDRepDist dist (suc c) ˢ) < n
                       × lengthˢ (mostStakeDRepDist dist (suc c) ˢ) < n
\end{code}\begin{code}[hide,inline]  where open import Data.Nat using (_<_) \end{code}
...
\begin{code}
expired : Epoch → GovActionState → Set
expired current record { expiresIn = expiresIn } = expiresIn < current
\end{code}\begin{code}[hide,inline] 
  where
    infix 4 _<_
     _<_ : Epoch → Epoch → Set
    a < b = a ≤ b × a ≢ b
\end{code}
```

src/Ledger/Prelude.agda
```diff
@@ -121,7 +124,22 @@ module Properties where
...
+insertIfJust : ∀ {A B} → ⦃ DecEq A ⦄ → A → Maybe B → A ⇀ B → A ⇀ B
+insertIfJust x nothing  m  = m
+insertIfJust x (just y) m  = insert m x y
+
+
+filterᵐ? : ∀ {A B} {P : A × B → Set} → (∀ x → Dec (P x)) → A ⇀ B → A ⇀ B
+filterᵐ? P? = filterᵐ (to-sp P?)
+
+filterᵐᵇ : ∀ {A B} → (A × B → Bool) → A ⇀ B → A ⇀ B
+filterᵐᵇ P = filterᵐ? (λ x → P x ≟ true)
```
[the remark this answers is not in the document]
WhatisRT: I've moved `insertIfJust` to `Axiom.Set.Map`, but `to-sp` is defined here so I don't think there's a good place to define these ATM. I'd like to refactor the whole `SpecProperty` stuff at some point anyway, and then we can probably move those into `Axiom.Set.Map` as well.

src/Ledger/GovernanceActions.lagda
WhatisRT: Yeah, there's an even better way to define `maximumˢ` which uses `Axiom.Set.Sum`. It's the same underlying definition, but it requires commutativity of the operation and gives you a bunch of theorems for that.

src/Ledger/Ratify.lagda
WhatisRT: I've put the second one in a `where` block, but I don't really care about the first, since it'll be most likely be removed entirely anyway.

src/Ledger/GovernanceActions.lagda
WhatisRT: I've tried your `CC∶_DRep∶_SPO∶_` suggestion, but the extra characters make it quite noisy and hard to read. I'm thinking that maybe we could just go for a notation that emulates a proper table: `|_|_|_|`. We could then even do some LaTeX trickery to give it a header line in the figure so it's clear what's what. Obviously we should also explain that in the prose.

omelkonian: This doesn't seem noisy to me to be honest:
```agda
   NoConfidence           → CC∶ noVote            DRep∶ vote P1      SPO∶ vote Q1
     (NewCommittee _ _ _)   → case ccThreshold' of λ where
     (just _)             → CC∶ noVote            DRep∶ vote P2a     SPO∶ vote Q2a
     nothing              → CC∶ noVote            DRep∶ vote P2b     SPO∶ vote Q2b
   (NewConstitution _ _)  → CC∶ vote ccThreshold  DRep∶ vote P3      SPO∶ noVote
   (TriggerHF _)          → CC∶ vote ccThreshold  DRep∶ vote P4      SPO∶ vote Q4
   (ChangePParams x)      → CC∶ vote ccThreshold  DRep∶ vote (P5 x)  SPO∶ noVote
   (TreasuryWdrl _)       → CC∶ vote ccThreshold  DRep∶ vote P6      SPO∶ noVote
   Info                   → CC∶ vote 2ℚ           DRep∶ vote 2ℚ      SPO∶ vote 2ℚ
```

WhatisRT: Maybe you have a higher pain threshold than me, I find this presentation quite overwhelming. I already dislike all the `vote/noVote` functions quite a bit.

I've made `⟨_,_,_⟩` private.

PR #229, opened 2023-10-02
Remove --overlapping-instances

omelkonian:
- In light of a recent issue with `--overlapping-instances`
  (c.f. https://github.com/agda/agda/issues/6895), I removed
  all uses of overlapping in the codebase, which we actually redundant.

- `Foreign.Convertible` no longer overlaps instances of `Coercible`.

- `Axiom.Set.List` now exports general decidable equality to avoid ambiguity
  when `Interface.DecEq.DecEq⇒Dec` is also in scope.


WhatisRT (approved):
Cool stuff, thank you!

PR #231, opened 2023-10-03
Tactic to remove trivial injectivity proofs

omelkonian:
- Introduce the `by-eq` tactic that automates proofs of the form:

    `λ where _ refl _ refl → refl`

- Use on injectivity proofs for `Axiom.Set.Map:mapKeys`, replacing the explicit argument of an injectivity proof with a tactic argument that invokes `by-eq`.

### DIFF
--- a/src/Axiom/Set/Map.agda
+++ b/src/Axiom/Set/Map.agda
@@ -28,6 +28,7 @@ open import Tactic.AnyOf
 open import Tactic.Assumption
 open import Tactic.Defaults
 open import Tactic.Helpers
+open import Tactic.ByEq
 
 -- Because of missing macro hygiene, we have to copy&paste this.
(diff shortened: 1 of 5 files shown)


WhatisRT (approved):
Cool, thank you! Two things:
- I usually went with the naming pattern of `<tac>` for a function, and `by-<tac>` for the corresponding macro. I'm open to changing that, but I think we should try to find a consistent scheme.
- One easy generalization that might be useful would be to take a tactic argument, and instead of unifying with `refl` instead try that tactic for the rhs of the pattern-lambda. The default can then be the `tryConstrs` tactic, which recursively applies constructors (so it'll try `refl`). I believe there are a few places where that generalization would be useful.

If you want you can merge this now & make issues for those points, or do them before you merge, whatever you prefer.

PR #238, opened 2023-10-09
Respect the exit status of `nix` commands

teodanciu:
to prevent fake successful builds: the build succeeding after the nix-command failed.

# Description

The processing of the nix-build command is causing the exit code of the nix command to be ignored. 
In this PR we are processing the result in a new step, so that if "nix-build" or "nix-store" fails, the build is also failing. 

For example, this is a PR reverting to a commit that caused a Haskell compilation error (and hence a nix-build failure) when building ledger: https://github.com/teodanciu/formal-ledger-specifications/pull/2
This build was "fake" succeeding. 
With the change in this commit, it is failing: https://github.com/teodanciu/formal-ledger-specifications/actions/runs/6456348932/job/17525659662?pr=2#step:10:5606

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,35 +56,43 @@ jobs:
       - name: Build agda
         id: agda
         run: |
-          v=$(nix-build -A agdaWithStdLibMeta)
-          closure=$(nix-store --query --requisites --include-outputs $v | tr '\n' ' ')
-          echo "derivation=$v" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


omelkonian:
LGTM, other than the fact that there is a new `ci/eval` check that's failing.

omelkonian (approved):

WhatisRT:
Cool, thank you! The new check looks like the IOG Hydra, which I believe is independent of this PR and needs a `flake.nix` to build the right things.

WhatisRT (approved):

PR #239, opened 2023-10-10
declare hydra jobs

dermetfan:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,10 +8,24 @@
     inherit (pkgs) lib;
   in {
     hydraJobs = lib.genAttrs systems (system: let
-      jobs = import ./. {
+      exposed = import ./. {
         inherit sources;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,22 @@
+{
+  outputs = _: let
+    systems = [ "x86_64-linux" ];
```
dermetfan, line 3: Currently only builds for linux. Add more here if desired.

WhatisRT (approved):
Thank you!

One extra thing that would be very nice to have would be if we could download the PDF that's contained in `*.docs`, but it doesn't need to be part of this PR. This was something you could do in the `cardano-ledger` repo in the past (by navigating hydra), and was very convenient.

teodanciu (approved):
Looks good to me, thank you!

PR #240, opened 2023-10-10
Final computational instances

### LINKED ISSUE #147: Make all Conway STSs executable
There are two ways to achieve this:

1. Improve `deriveComp` to be able to automatically derive what we need, or
2. implement the computation strategies manually.

Obviously the former is the preferred way, but if this takes too long we may have to do the latter.

UlfNorell:
Fixes #147

This adds Computational instances for the remaining rules.

Also fixes some minor issues in a few of the rules (`RATIFY'`, `NEWPP` and `CHAIN`)

### DIFF
--- a/src/Ledger/Introduction.lagda
+++ b/src/Ledger/Introduction.lagda
@@ -5,6 +5,9 @@
 module Ledger.Introduction where
 
 open import Prelude
+import Data.Maybe as Maybe
+open import Data.Maybe.Properties
+open import Relation.Binary.PropositionalEquality
 \end{code}
 
 Repository: https://github.com/input-output-hk/formal-ledger-specifications
@@ -39,7 +42,7 @@ other, forming a directed graph that is almost a tree.
 
 Since all such state machines need to be evaluated by the node and all
 nodes should compute the same states, the relations specified by them
-should be computable by functions. This is captured by the following
+should be computable by functions. This can be captured by the following
 record, which is parametrized over the step relation.
 
 \begin{code}[hide]
@@ -55,8 +58,14 @@ record Computational (_⊢_⇀⦇_,X⦈_ : C → S → Sig → S → Set) : Set 
   field
     compute     : C → S → Sig → Maybe S
     ≡-just⇔STS  : compute Γ s b ≡ just s' ⇔ Γ ⊢ s ⇀⦇ b ,X⦈ s'
+
+  nothing⇒∀¬STS : compute Γ s b ≡ nothing → ∀ s' → ¬ Γ ⊢ s ⇀⦇ b ,X⦈ s'
 \end{code}
 \end{figure*}
+\begin{code}[hide]
+  nothing⇒∀¬STS comp≡nothing s' h rewrite ≡-just⇔STS .Equivalence.from h =
+    case comp≡nothing of λ ()
+\end{code}
 
 \subsection{Sets \& maps}
 
--- a/src/Ledger/NewPP/Properties.agda
+++ b/src/Ledger/NewPP/Properties.agda
@@ -1,0 +1,29 @@
+{-# OPTIONS --safe #-}
+
+open import Relation.Nullary.Decidable
+
+open import Ledger.Prelude
+open import Ledger.Transaction
+
+module Ledger.NewPP.Properties (txs : _) (open TransactionStructure txs) where
+
+open import Ledger.PPUp txs
+open import Ledger.NewPP txs
+
+instance
+  Computational-NEWPP : Computational _⊢_⇀⦇_,NEWPP⦈_
+  Computational-NEWPP = record {M} where module M Γ s (open NewPParamState s) where
+    computeProof = λ where
+      nothing → just (_ , NEWPP-Reject)
+      (just upd) → let newpp = applyUpdate pparams upd in
+        case ¿ viablePParams newpp ¿ of λ where
+          (yes p) → just (_ , NEWPP-Accept p)
+          (no _)  → nothing
+
+    completeness : _
+    completeness sig s' h with sig | h
+    ... | nothing  | NEWPP-Reject   = refl
+    ... | just upd | NEWPP-Accept p
+      rewrite let newpp = applyUpdate pparams upd in
+              dec-yes (¿ viablePParams newpp ¿) p .proj₂
+              = refl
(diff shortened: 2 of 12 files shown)


WhatisRT:
Thank you! Maybe we can have a quick look at it in the meeting soon? One thing I can point out right now is that CI is failing since `Ledger.NewPP.Properties` isn't imported anywhere, so it isn't type checked. You can stick it into `Everything.agda` for now so that we have it for when we get around to formalizing some pre-Conway stuff.

src/Ledger/Introduction.lagda
```diff
@@ -53,10 +56,30 @@ private variable
 \begin{code}
 record Computational (_⊢_⇀⦇_,X⦈_ : C → S → Sig → S → Set) : Set where
   field
-    compute     : C → S → Sig → Maybe S
```
WhatisRT, line 56: I think it's better to leave the presentation in the PDF in the old style because it should be easier to understand by people who don't know too much type theory.

I do however like adding a few of the theorems like you did here. Maybe we could prove them by translating this version of `Computational` into the new one?

src/Ledger/NewPP/Properties.agda
```diff
@@ -0,0 +1,26 @@
+
```
WhatisRT, line 1: Some extra empty lines here

src/Ledger/Ratify.lagda
```diff
@@ -362,25 +362,29 @@ activeVotingStake : ℙ VDeleg → StakeDistrs → (VDeleg ⇀ Vote) → Coin
 activeVotingStake cc dists votes =
   Σᵐᵛ[ x  ← getStakeDist DRep cc dists ∣ dom votes ᶜ ᶠᵐ ] x
 
-accepted' : RatifyEnv → EnactState → GovActionState → Set
-accepted' Γ (record { cc = cc , _ ; pparams = pparams , _ }) gs =
-  acceptedBy CC ∧ acceptedBy DRep ∧ acceptedBy SPO ∧ meetsMinAVS
+module _ (Γ : RatifyEnv) (es : EnactState) (gs : GovActionState)
```
WhatisRT, line 365: One issue with putting this in a module is that we want to avoid module as much as we can in the PDF, which is why the `where` block was a good solution. The issue is of course that we now need `acceptedBy` outside of the scope of `accepted`. I think in that case it might be better to split this into two functions at the top level at the cost of repeating the telescope of arguments twice.

UlfNorell: Can do, but not that it's not just the telescope, but also `votes'`, `cc'` and `redStakeDistr`.

src/Ledger/Introduction.lagda
UlfNorell: Ok, I'll revert this back to the old Computational and add the `nothing⇒∀¬STS` property.

src/Ledger/Ratify.lagda
WhatisRT: Ahh, you're right. Well, I have some good news: I'm almost certain we won't have `minimumAVS` in the final version of the spec, so I think it's safe to remove it, and then those definitions can just go in `acceptedBy`.

src/Ledger/Introduction.lagda
WhatisRT: Thank you!

src/Ledger/Ratify/Properties.agda
```diff
@@ -8,31 +8,101 @@ module Ledger.Ratify.Properties (txs : _) (open TransactionStructure txs) where
...
+                                           (compute ⟦ a .proj₁ , treasury , currentEpoch ⟧ᵉ es action) refl s' h)
+
+Computational-RATIFY : Computational _⊢_⇀⦇_,RATIFY⦈_
+Computational-RATIFY = it
+
+RATIFY-total : ∀ {Γ s sig} → ∃[ s' ] Γ ⊢ s ⇀⦇ sig ,RATIFY⦈ s'
+RATIFY-total = SS⇒BS-total RATIFY'-total
+
+RATIFY-complete : ∀ {Γ s sig s'} → Γ ⊢ s ⇀⦇ sig ,RATIFY⦈ s' → RATIFY-total {Γ} {s} {sig} .proj₁ ≡ s'
+RATIFY-complete h = computational⇒rightUnique it (RATIFY-total .proj₂) h
```
omelkonian, line 108: This is quite cumbersome, given the evident repetition of common types and terms, c.f. the general review comment for possible solutions.

omelkonian (changes requested):
LGTM overall, but the RATIFY proof could use some love.

Here's an [alternative approach](https://github.com/input-output-hk/formal-ledger-specifications/commit/ef1aa895ca70ab161d64fbc52a65a6bf95ad9e45) that tackles the issue at its root, namely making the initial definitions of accepted/expired/... properly `abstract`, which leads to [a much cleaner proof](https://github.com/input-output-hk/formal-ledger-specifications/blob/ef1aa895ca70ab161d64fbc52a65a6bf95ad9e45/src/Ledger/Ratify/Properties.agda) that the RATIFY relation is computational.

The downside of course is that the decision procedures have to be put in the same file (i.e. `Ledger.Ratify.lagda`), breaking modularity a bit, however that can be tackled in Agda-2.6.4 using `opaque`. Since @UlfNorell's next step is to bump the Agda version, it might seem reasonable to revisit this PR immediately after that so that we can have the best of both worlds.

PS: The [branch above](https://github.com/input-output-hk/formal-ledger-specifications/compare/final-computational-instances-orestis-review) also contains some minor cleanups on the other files (Chain/NewPP) and a handful of constructor patterns to make life easier when working on RATIFY:
```agda
pattern RATIFY-Continue₁  x y = RATIFY-Continue (inj₁ (x , y))
pattern RATIFY-Continue₂  x y = RATIFY-Continue (inj₂ (x , y))
pattern RATIFY-Continue₂₁ x y = RATIFY-Continue₂ x (inj₁ y)
pattern RATIFY-Continue₂₂ x y = RATIFY-Continue₂ x (inj₂ y)
```
Feel free to cherry-pick regardless of how we handle the abstraction issue above.

Alternatively, if we want this PR merged ASAP, we can create a new issue from the above discussion, and clean up the current approach as I demonstrate [here](https://github.com/input-output-hk/formal-ledger-specifications/commit/dd5f1586d36a585d3c45ef450e4579809f9c893e): apart from overall styling fixes (e.g. for smaller lines), it avoids repetition by formulating what we are abstracting over explicitly and bundling all (abstracted) arguments in a tuple.

src/Ledger/Ratify/Properties.agda
UlfNorell: That branch looks good, I'm merging it in here. Then we can upgrade the abstract to opaque in the 2.6.4 migration.

omelkonian (approved):
LGTM; I also checked the .pdf to make sure the insertion of `abstract` did not indent stuff.

omelkonian:
@WhatisRT it's now ready to approve and merge.

WhatisRT (approved):

Issue #146, opened 2023-07-19
PDF cleanup


WhatisRT:
This is a blanket issue for various refactoring and improvements regarding the LaTeX code.

- [X] ~Explain (Agda) notations~
  - [X] ~`record` definitions~ (moved to #258)
  - [X] ~`open` for records~ (moved to #258)
- [X] ~Remove superscripts from `⟦_,...,_⟧` (probably via a typeclass)~ (moved to #258)
- [X] ~Remove `_ᶠᵐ` notation from sums by introducing a typeclass~ (moved to #258)
- [X] ~Potentially unify notations for empty/singleton maps and sets via a typeclass~ (moved to #258)
- [x] Rename `∪ᵐˡ` to `∪ˡ` or `∪⃗`
- [X] ~Replace `≡ᵉ` with `≡` (we may not be able to actually make this replacement in the Agda code, so we may have to do some preprocessing of the LaTeX)~ (moved to #258)
- [X] ~Lots of `mapˢ`. Since we now have a `Functor` class, we can make an instance and write `map` instead~ (moved to #258)
- [X] ~Can we remove `+ᵉ` somehow? We could change `HasAdd` so that `+` can take to different types, and then for a ring define some instances where you can always add natural numbers/integers. But maybe we can do something easier?~ (moved to #258)
- [x] Figure 4: Add space below `RwdAddr`
- [x] Figure 6: Align `actionWellFormed`
- [x] Figure 8 (Functions related to voting)
  - [x] Remove comments
  - [x] Add an alias for `2ℚ`, and explain that any number above `1` is fine
  - [X] ~Potentially replace `ℚ` with a unit interval type~
  - [X] Maybe replace `⟨_,_,_⟩` with `|_|_|_|` for the thresholds table
- [X] Figure 8 (Protocol parameter declarations)
  - [X] ~Can we move `Acnt` elsewhere?~ (moved to #258)
  - [x] Replace `pparamsWellFormed` with a predicate (instead of a boolean predicate), which should look cleaner
  - [X] ~This figure is placed inside `\small`, which is OK, but it'd be better to find some other way to fit it so that the font sizes are consistent.~
  - [x] Maybe rename `minCCSize` to `ccMinSize` for consistency?
  - [x] Maybe rename `ccTermLimit` to `ccMaxTermLength`, since that's what it really is?
- [x] Figure 16 (Types used for CERTS transition system): Remove comments
[...]

PR #245, opened 2023-10-16
PDF cleanup: simplify code in Figure 34

williamdemeo:
# Description

Rearrange code to simplify nested where blocks in Fig 34, as suggested in one of the items in issue #146.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Ratify.lagda
+++ b/src/Ledger/Ratify.lagda
@@ -234,25 +234,32 @@ restrictedDists coins rank dists = dists
 \begin{figure*}[h!]
 {\small
 \begin{code}
-actualVotes : RatifyEnv → CCData → (GovRole × Credential) ⇀ Vote → GovAction → PParams
-            → VDeleg ⇀ Vote
-actualVotes Γ cc votes ga pparams
-  =    mapKeys (credVoter CC) actualCCVotes
-  ∪ᵐˡ  actualPDRepVotes ∪ᵐˡ actualDRepVotes
-  ∪ᵐˡ  actualSPOVotes
-  where
-    open RatifyEnv Γ
-    open PParams pparams
+module _ (Γ : RatifyEnv)(pparams : PParams) where
+
+  open RatifyEnv Γ
+  open PParams pparams
+
+  activeDReps : ℙ Credential
+  activeDReps = dom (filterᵐ? (λ x → currentEpoch ≤? (proj₂ x)) dreps)
+
+  activeCC : CCData → ℙ Credential
+  activeCC (just (cc , _))  = dom (filterᵐᵇ (is-just ∘ proj₂) (ccHotKeys ∣ dom cc))
+  activeCC nothing          = ∅
+
+  spos : ℙ VDeleg
+  spos = filterˢ isSPOProp $ dom (StakeDistrs.stakeDistr stakeDistrs)
+
+  module _ (votes : (GovRole × Credential) ⇀ Vote) where
 
     roleVotes : GovRole → VDeleg ⇀ Vote
     roleVotes r = mapKeys (uncurry credVoter) (filterᵐ? ((r ≟_) ∘ proj₁ ∘ proj₁) votes)
 
-    activeCC activeDReps : ℙ Credential
-    activeCC = case cc of λ where
-      (just (cc , _))  → dom (filterᵐᵇ (is-just ∘ proj₂) (ccHotKeys ∣ dom cc))
-      nothing          → ∅
+    actualPDRepVotes actualSPOVotes : GovAction → VDeleg ⇀ Vote
+    actualPDRepVotes NoConfidence = ❴ abstainRep , Vote.abstain ❵ᵐ ∪ᵐˡ ❴ noConfidenceRep , Vote.yes ❵ᵐ
+    actualPDRepVotes _ = ❴ abstainRep , Vote.abstain ❵ᵐ ∪ᵐˡ ❴ noConfidenceRep , Vote.no ❵ᵐ
 
-    activeDReps = dom (filterᵐ? (λ x → currentEpoch ≤? (proj₂ x)) dreps)
+    actualSPOVotes (TriggerHF _)  = roleVotes GovRole.SPO ∪ᵐˡ constMap spos Vote.no
+    actualSPOVotes _              = roleVotes GovRole.SPO ∪ᵐˡ constMap spos Vote.abstain
 
     actualCCVote : Credential → Epoch → Vote
     actualCCVote c e =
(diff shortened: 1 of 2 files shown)


omelkonian:
I though we shouldn't be using `module` in the code displayed on PDF, only in extremely rare cases.

williamdemeo:
@WhatisRT I think this PR is good to know... please lmkwyt.

src/Ledger/Ratify.lagda
```diff
@@ -231,77 +231,82 @@ restrictedDists coins rank dists = dists
...
-    activeCC activeDReps : ℙ Credential
-    activeCC = case cc of λ where
-      (just (cc , _))  → dom (filterᵐᵇ (is-just ∘ proj₂) (ccHotKeys ∣ dom cc))
-      nothing          → ∅
+    actualCCVotes : CCData → Credential ⇀ Vote
+    actualCCVotes nothing          = ∅ᵐ
+    actualCCVotes (just (cc , q))  =
+      case ¿ ccMinSize ≤ lengthˢ (activeCC $ just (cc , q)) ¿ᵇ of
+      λ where  true   → mapWithKey actualCCVote cc
+               false  → constMap (dom cc) Vote.no
```
omelkonian, line 278: *Code smell:* `case b of λ where true → ...; false → ...` is equivalent to `if b then ... else ...`

src/Ledger/Ratify.lagda
```diff
@@ -231,77 +231,82 @@ restrictedDists coins rank dists = dists
...
+  actualPDRepVotes NoConfidence  =  ❴ abstainRep , Vote.abstain ❵ᵐ
+                                    ∪ᵐˡ ❴ noConfidenceRep , Vote.yes ❵ᵐ
+  actualPDRepVotes _             =  ❴ abstainRep , Vote.abstain ❵ᵐ
+                                    ∪ᵐˡ ❴ noConfidenceRep , Vote.no ❵ᵐ
+
+  actualVotes : CCData → GovAction → (GovRole × Credential) ⇀ Vote → VDeleg ⇀ Vote
+  actualVotes cc ga votes  =  mapKeys (credVoter CC) (actualCCVotes cc)
+                              ∪ᵐˡ actualPDRepVotes ga
+                              ∪ᵐˡ actualDRepVotes
+                              ∪ᵐˡ actualSPOVotes ga
```
omelkonian, line 262: I would prefer aligning the unions with the `=`, thereby not putting unnecessary focus to the head, unless you really mean to.

(also applies to other definitions below)

omelkonian (changes requested):

src/Ledger/Ratify.lagda
WhatisRT: You can even use `ifᵈ` here, then you don't need the `¿`

src/Ledger/Ratify.lagda
```diff
@@ -231,77 +231,82 @@ restrictedDists coins rank dists = dists
...
-            → VDeleg ⇀ Vote
-actualVotes Γ cc votes ga pparams
-  =    mapKeys (credVoter CC) actualCCVotes
-  ∪ᵐˡ  actualPDRepVotes ∪ᵐˡ actualDRepVotes
-  ∪ᵐˡ  actualSPOVotes
-  where
-    open RatifyEnv Γ
-    open PParams pparams
+  activeDReps : ℙ Credential
+  activeDReps = dom $ filterᵐ? (λ x → currentEpoch ≤? (proj₂ x)) dreps
```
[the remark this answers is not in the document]
williamdemeo: Okay, but what you're describing @WhatisRT is exactly what I did already.  We had everything in a where clause and an item of issue #146 (PDF cleanup) requests a simplification of the proof and suggests removing some stuff from the where clause.  I 
simplified the proof and, to further simplify the presentation, I created the module and, indeed, I even added to the documentation a phrase like, "we assume a fixed ratify environment" (around line 299).

src/Ledger/Ratify.lagda
williamdemeo: Here I was trying to simplify while keeping the code as close to the original as possible (see lines 259---261 of the original code) ...but now I see that it can be simplified further as you suggest.

src/Ledger/Ratify.lagda
williamdemeo: Aligning operators with the `=` symbol is not my preferred style; it looks strange from a pure math perspective... but since it's 2 against 1, I'll make the change.

src/Ledger/Ratify.lagda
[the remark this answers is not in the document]
WhatisRT: The issue is that nothing in those changes justifies breaking up the `where` block. You did some simplifications, but you can just put all that new code back under the `where` block and it's fine. The issue I'm having with that figure isn't that it's a big **`where`** block, it's that it is a **big** `where` block. So what should be done about that? Well, definitions that don't necessarily need to be in that block can be moved out (`isHF` only depends on the governance action and can be general purpose, so that's a good candidate. Are there others?). Also, are there definitions that are used only once? If yes, does readability suffer if you inline it? If no, go ahead and inline it.

src/Ledger/Ratify.lagda
```diff
@@ -271,22 +271,21 @@ module _ (Γ : RatifyEnv)(pparams : PParams) where
...
     actualSPOVotes : GovAction → VDeleg ⇀ Vote
-    actualSPOVotes (TriggerHF _)  = roleVotes GovRole.SPO ∪ᵐˡ constMap spos Vote.no
-    actualSPOVotes _              = roleVotes GovRole.SPO ∪ᵐˡ constMap spos Vote.abstain
+    actualSPOVotes (TriggerHF _)  = roleVotes GovRole.SPO ∪ˡ constMap spos Vote.no
+    actualSPOVotes _              = roleVotes GovRole.SPO ∪ˡ constMap spos Vote.abstain
 
     actualDRepVotes : VDeleg ⇀ Vote
     actualDRepVotes  =  roleVotes GovRole.DRep
-                        ∪ᵐˡ constMap (mapˢ (credVoter DRep) activeDReps) Vote.no
+                     ∪ˡ  constMap (mapˢ (credVoter DRep) activeDReps) Vote.no
```
omelkonian, line 288: Misaligned `roleVotes ...`.

williamdemeo:
I removed the module introducing the module parameters that are assumed throughout the block of code that computes actual vote.

@WhatisRT and/or @omelkonian, please approve if you're okay with the changes.

omelkonian (approved):
LGTM

PR #248, opened 2023-10-18
PDF cleanup: Fig 36

williamdemeo:
# Description

Address the following items of issue #146 concerning Figure 36.

- [ ] We shouldn't show `SpecProperty` stuff, `sp-∘` in this case.
- [X]  The pattern-matching trick in `acceptedBy` to expose the fact that the number isn't 0 (i.e. `x@(suc _)`) is black magic. We shouldn't do this in the figure.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Ratify.lagda
+++ b/src/Ledger/Ratify.lagda
@@ -79,12 +79,12 @@ above the threshold.
 \begin{code}[hide]
 {-# OPTIONS --safe #-}
 
-import Data.Integer as ℤ
-open import Data.Rational as ℚ using (ℚ; 0ℚ)
+open import Data.Integer using () renaming (+_ to _ᶻ)
+open import Data.Rational as ℚ using (ℚ; 0ℚ; _/_; _≥_)
 open import Data.Nat.Properties hiding (_≟_; _≤?_)
 open import Data.Nat.Properties.Ext
 
-open import Ledger.Prelude hiding (_∧_)
+open import Ledger.Prelude hiding (_∧_; _/_; _≥_)
 open import Ledger.Transaction
 
 module Ledger.Ratify (txs : _) (open TransactionStructure txs) where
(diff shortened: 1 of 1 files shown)


williamdemeo:
I'm not sure how to present `getStakeDist` without exposing the `SpecProperty` stuff (`sp-∘`).  @WhatisRT suggestions?

src/Ledger/Ratify.lagda
```diff
@@ -201,16 +201,16 @@ mostStakeDRepDist-∅ {dist} = suc (Σᵐᵛ[ x ← dist ᶠᵐ ] x) , Propertie
         <⟨ v>sum ⟩
       v ∎
 
-∃topNDRepDist : ∀ {n dist} → lengthˢ (dist ˢ) ≥ n → n > 0
-                → ∃[ c ] lengthˢ (mostStakeDRepDist dist c ˢ) ≥ n
+∃topNDRepDist : ∀ {n dist} → n ≤ lengthˢ (dist ˢ) → n > 0
```
williamdemeo, line 204: Reserve `≥` for rationals in this module.

(quick and dirty way to use `≥` in place of `ℚ.≥` below)

src/Ledger/Ratify.lagda
```diff
@@ -364,22 +364,40 @@ abstract
...
+    redStakeDistr : StakeDistrs
+    redStakeDistr = restrictedDists coinThreshold rankThreshold stakeDistrs
+
+    t : ℚ
+    t = maybe id ℚ.0ℚ $ threshold pparams (proj₂ <$> cc) action role
+
+    stakeFilter : Coin → Set
+    stakeFilter zero = t ≡ ℚ.0ℚ
+    stakeFilter (suc x) = (acceptedStake role cc' redStakeDistr votes')ᶻ / suc x ≥ t
+
```
williamdemeo, line 397: The proof of `acceptedBy` is now longer but hopefully much simpler and easier for non-experts to follow and maybe even comprehend.

src/Ledger/Ratify.lagda
```diff
@@ -364,22 +364,40 @@ abstract
...
+  accepted : RatifyEnv → EnactState → GovActionState → Set
+  accepted Γ es gs = acceptedBy Γ es gs CC ∧ acceptedBy Γ es gs DRep ∧ acceptedBy Γ es gs SPO
+
```
williamdemeo, line 400: We might simplify the presentation of `acceptedBy` and `accepted` further with a (pdf-hidden) `module _` directive with parameters `Γ : RatifyEnv`, `es : EnactState`, and `gs : GovActionState`.

src/Ledger/Ratify.lagda
WhatisRT: Now that you worked so hard on the typeclasses, you should just make an instance for `ℚ`!

src/Ledger/Ratify.lagda
WhatisRT: Or just put `acceptedBy' = acceptedBy Γ es gs` in a `where` block

WhatisRT (changes requested):
I didn't really like that this is longer and it only seemed marginally better, so I decided to help you out, see #250. There's still some more stuff to do though.

williamdemeo:
@WhatisRT has a better approach (#250) addressing most of the stuff here.

PR #249, opened 2023-10-18
PDF cleanup: remove `\small` latex commands

williamdemeo:
# Description

Some code blocks were enclosed inside `{\small ... }` which resulted in inconsistent code font size across the document.

This PR removes all `\small` latex directives and changes the global Agda code font size to "small."

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/latex/agda.sty
+++ b/src/latex/agda.sty
@@ -358,9 +358,8 @@
 % ----------------------------------------------------------------------
 % The code environment.
 
-%\newcommand{\AgdaCodeStyle}{}
-\newcommand{\AgdaCodeStyle}{\small}
-%\newcommand{\AgdaCodeStyle}{\tiny}
+\newcommand{\AgdaCodeStyle}{}
+% \newcommand{\AgdaCodeStyle}{\tiny}
 
 \ifdefined\mathindent
   {}
--- a/src/latex/agda.sty
+++ b/src/latex/agda.sty
@@ -358,8 +358,9 @@
 % ----------------------------------------------------------------------
 % The code environment.
 
-\newcommand{\AgdaCodeStyle}{}
-% \newcommand{\AgdaCodeStyle}{\tiny}
+%\newcommand{\AgdaCodeStyle}{}
+\newcommand{\AgdaCodeStyle}{\small}
+%\newcommand{\AgdaCodeStyle}{\tiny}
 
 \ifdefined\mathindent
   {}
--- a/src/Ledger/GovernanceActions.lagda
+++ b/src/Ledger/GovernanceActions.lagda
@@ -32,7 +32,6 @@ maximum : ℙ ℚ → ℚ
 maximum x = foldl Data.Rational._⊔_ 0ℚ (proj₁ $ finiteness x)
 \end{code}
 \begin{figure*}[h]
-{\small
 \begin{code}
 GovActionID : Set
 GovActionID = TxId × ℕ
@@ -62,7 +61,6 @@ actionWellFormed : GovAction → Bool
 actionWellFormed (ChangePParams x)  = ppdWellFormed x
 actionWellFormed _                  = true
 \end{code}
-} %% end small
 \caption{Governance actions}
 \label{defs:governance}
 \end{figure*}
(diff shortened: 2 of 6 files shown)


src/latex/agda.sty
```diff
@@ -358,8 +358,9 @@
 % ----------------------------------------------------------------------
 % The code environment.
 
-\newcommand{\AgdaCodeStyle}{}
-% \newcommand{\AgdaCodeStyle}{\tiny}
+%\newcommand{\AgdaCodeStyle}{}
+\newcommand{\AgdaCodeStyle}{\small}
+%\newcommand{\AgdaCodeStyle}{\tiny}
```
omelkonian, line 363: First, `agda.sty` is generated by Agda and you should never change it manually; in this case, you can simply do `\renewcommand{\AgdaCodeStyle}{\small}` in our preamble.

Second, this only changes the `\begin{code}` blocks and not inlined things (c.f. the size difference between the two `expired` in pg.41 of the PDF).

Third, should we really be modifying all code blocks globally? If you take a closer look at `latex/preamble.tex` you see that I've picked up a sweet spot between `MatchLowercase` and `MatchUppercase` that makes the font size in the Agda code match the font size in the prose:
```latex
\newcommand\agdaFont{StrippedJuliaMono}
\newcommand\agdaFontOptions{
Path=fonts/,
Extension=.ttf,
UprightFont=*-Regular,
BoldFont=*-Bold,
ItalicFont=*-RegularItalic,
BoldItalicFont=*-BoldItalic,
%% Scale=MatchLowercase
Scale=0.85
%% Scale=MatchUppercase
}
```

I suggest we keep the sane defaults as they are, and then locally change the fonts in specific files/definitions that use very wide lines, e.g. I see there a quite a few occurrences of `\small` in `Ledger/GovernanceActions.lagda` so it might make sense to do `\renewcommand{\AgdaCodeStyle}{\small}` at the start of the file and `\renewcommand{\AgdaCodeStyle}{}` (or define aliases for `\setAgdaSmall` and `unsetAgdaSmall` and use these).

Having said that, at the end of the day I think this should be applied locally to specific figures (possibly adding a `smallcode` environment to make things easier to maintain), as doing anything on a broader scope will inevitably lead to nasty surprises.

omelkonian (changes requested):

src/latex/agda.sty
WhatisRT: Ah, I didn't see that you already added some font scaling. In that case, I'd just get rid of all `\small` stuff for now. Then, once we're in the final stages of polishing the PDF we can think about doing stuff like that if we can't find a nicer way.

Also, good point that `agda.sty` can be auto generated. I'll make an issue to include that as part of our build process, then we can remove it.

williamdemeo: > First, `agda.sty` is generated by Agda and you should never change it manually; in this case, you can simply do `\renewcommand{\AgdaCodeStyle}{\small}` in our preamble.
>
 
Fair point.  I should have used `\renewcommand` instead of editing the `agda.sty` file.

> Second, this only changes the `\begin{code}` blocks and not inlined things (c.f. the size difference between the two `expired` in pg.41 of the PDF).
> 

I see no problem with having inlined code that is slightly larger than code that appears in figures.  In fact, that seems quite natural to me.

> Third, should we really be modifying all code blocks globally? If you take a closer look at `latex/preamble.tex` you see that I've picked up a sweet spot between `MatchLowercase` and `MatchUppercase` that makes the font size in the Agda code match the font size in the prose:
>

I think the way you change the code font size is fine, but I don't think the code font size in figures should be as large as the font size as the prose.  That seems unnecessarily large to me.

> 
> I suggest we keep the sane defaults as they are, and then locally change the fonts in specific files/definitions that use very wide lines, e.g. I see there a quite a few occurrences of `\small` in `Ledger/GovernanceActions.lagda` so it might make sense to do `\renewcommand{\AgdaCodeStyle}{\small}` at the start of the file and `\renewcommand{\AgdaCodeStyle}{}` (or define aliases for `\setAgdaSmall` and `unsetAgdaSmall` and use these).

Yes, I added those `\small`s to prevent line overruns, but I thought you didn't like inconsistent code font size.  The point of this PR is to correct that flaw.  But maybe the more serious complaint was that the *way* I changed the code locally was too hard to maintain...?

> 
> Having said that, at the end of the day I think this should be applied locally to specific figures (possibly adding a `smallcode` environment to make things easier to maintain), as doing anything on a broader scope will inevitably lead to nasty surprises.

I suppose we just put concerns about code font size and line overruns on the back burner for now and worry about it once the PDF is in a more stable state.

williamdemeo:
I'm reopening this because it merely removes the `\small`s and decreases the code scale using @omelkonian's preamble parameter, so it might now be acceptable to all.

omelkonian (approved):
LGTM

WhatisRT:
In the future, please try to rebase your branch on master instead of merging master into it. Fixing the merge conflicts is a lot easier that way.

PR #253, opened 2023-10-18
Add Vote and Propose script purposes

### LINKED ISSUE #125: Voting script purpose
This depends on #13 which is currently WIP.

WhatisRT:
# Description

Closes #125. I've also switched `scriptsNeeded` to the `credsNeeded` based version, which removes a bit of code duplication, and should hopefully be cleaner.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -173,7 +173,7 @@ HSGovStructure = record
   }
 instance _ = HSGovStructure
 
-open import Ledger.GovernanceActions it
+open import Ledger.GovernanceActions it hiding (Vote)
 open import Ledger.Deleg it
 open import Ledger.Gov it
 
@@ -262,8 +262,8 @@ instance
 
   Convertible-Tag : Convertible Tag F.Tag
   Convertible-Tag = λ where
-    .to   → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → F.Vote; Propose → Propose }
-    .from → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → Tag.Vote; Propose → Propose }
+    .to   → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → Vote; Propose → Propose }
+    .from → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → Vote; Propose → Propose }
    where open F.Tag
 
   Convertible-TxWitnesses : Convertible TxWitnesses F.TxWitnesses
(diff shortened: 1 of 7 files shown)


src/Ledger/Foreign/HSLedger.agda
```diff
@@ -260,8 +262,8 @@ instance
 
   Convertible-Tag : Convertible Tag F.Tag
   Convertible-Tag = λ where
-    .to   → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd }
-    .from → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd }
+    .to   → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → F.Vote; Propose → Propose }
+    .from → λ{ Spend → Spend; Mint → Mint; Cert → Cert; Rewrd → Rewrd; Vote → Tag.Vote; Propose → Propose }
    where open F.Tag
```
omelkonian, line 267: These qualifications shouldn't be necessary I hope.

src/Ledger/ScriptValidation.agda
```diff
@@ -16,7 +16,7 @@ open import Ledger.Crypto
 
 
 module Ledger.ScriptValidation
-  (txs : _) (open TransactionStructure txs)
+  (txs : _) (open TransactionStructure txs hiding (Vote))
```
omelkonian, line 19: Previously we didn't need to hide `Cert/Rwrd/Mint/Spend`, so why is this now necessary for `Vote`?

src/Ledger/Transaction.lagda
```diff
@@ -28,7 +28,7 @@ open import MyDebugOptions
 open import Relation.Nullary.Decidable using (⌊_⌋)
 
 data Tag : Set where
-  Spend Mint Cert Rewrd : Tag
+  Spend Mint Cert Rewrd Vote Propose : Tag
```
omelkonian, line 31: If this creates all the visibility issues in this PR (i.e. `hiding (Vote)` everywhere), I would suggest naming these something else (e.g. `spend/mint/` or `SpendTag/MintTag/...`) to avoid such issues, unless this is super-important for the PDF presentation.

omelkonian (changes requested):

src/Ledger/Transaction.lagda
WhatisRT: The issue isn't the constructor itself, but the `Vote` type. There's both a `Vote` `Tag` and `ScriptPurpose`, and I went for the name `Vote` for consistency with all the other constructor names. I think my preferred option would be to not publicly reexport the `Vote` type from `TransactionStructure`.

omelkonian: Sounds good!

omelkonian (approved):

PR #255, opened 2023-10-19
CI revamp: base everything on Makefile

omelkonian:
+ Haskell test suite instead of an arbitrary Main, which now runs on CI
   - also adds the Haskell files in the generated website for completeness, c.f. the [PR's source branch Github page](https://omelkonian.github.io/formal-ledger-specifications/) for the resulting website.
+ [fix] LaTeX issue with `lmroman` fonts 
+ Upload PDFs in a .zip artifact on CI builds of PRs (for better reviewing workflow)
   - c.f. the [new item in the Actions page](https://github.com/input-output-hk/formal-ledger-specifications/actions/runs/6573113111) that lets us [download the generated PDFs](https://github.com/input-output-hk/formal-ledger-specifications/suites/17401301539/artifacts/994580899) uploaded from this PR's CI build

NB: The more principled approach introduced in this PR will lay out a better foundation for the upcoming *conformance testing* pipeline.


WhatisRT (commented):
This looks fine to me, but I don't know enough about CI to review this properly, so I'll defer this to @teodanciu

omelkonian:
**ACTION REQUIRED** 
After merge, for the Github page to be deployed correctly, the `gh-pages` folder must be set to `docs/` (in contrast to the previous root folder `/`) in `Settings -> Pages -> Branch`.

omelkonian:
> This looks fine to me, but I don't know enough about CI to review this properly, so I'll defer this to @teodanciu

@WhatisRT there are still a lot of changes unrelated to CI which you might want to take a look at though.
Specifically, anything other than `build.yml` has to do with building the project and, in particular, how the Haskell package is built and tested.

src/latex/preamble.tex
```diff
@@ -35,6 +35,7 @@
 % Math fonts
 \usepackage{unicode-math}
 \setsansfont{XITSMath-Regular.otf}
+\setmathfont{XITSMath-Regular.otf}
```
WhatisRT, line 38: What's the issue this is fixing?

WhatisRT (commented):
Yeah, what I should have said is: I don't know enough about CI, Make, HSpec or Cabal to review this properly😅 All I can really do here is run everything myself and see if it works

WhatisRT:
I just noticed that the section [Test the spec using nix-shell](https://github.com/input-output-hk/formal-ledger-specifications#test-the-spec-using-nix-shell) of the README will be outdated by this PR, could you update that as well?

.github/workflows/build.yml
```diff
@@ -122,59 +124,29 @@ jobs:
...
-          mkdir -p html
-          cp -L -r outputs/ledger/html html/ledger
-          cp -L -r outputs/midnight/html html/midnight
-          ledger=$(nix eval -f default.nix "ledger.name")
-          midnight=$(nix eval -f default.nix "midnight.name")
-          echo "ledger=$ledger" >> $GITHUB_OUTPUT
-          echo "midnight=$midnight" >> $GITHUB_OUTPUT
-
-      - name: Build static file
+      - name: Upload PDF artifacts
```
teodanciu, line 129: Nice that we can remove all of this, but I don't get it - where are the pdfs created?

teodanciu: Ah I think I see,  the `Build ledger` above  is already preparing the pdfs, right? with this :`rsync -r --exclude={'**/nix-support','**/lib'} outputs/ledger*/* docs/`.

default.nix
```diff
@@ -63,68 +63,97 @@ rec {
...
+        # OUT_DIR=$out make "${project}".hsTest
+      '';
+      dontInstall = true;
     };
 
-  ledger = specsDerivations {
-    project = "cardano-ledger";
-    dir = "Ledger";
-    agdaLedgerFile = "Foreign/HSLedger.agda";
+    # hsDocs = stdenv.mkDerivation {
```
teodanciu, line 128: This left on purpose commented out?

default.nix
teodanciu, line 1: It looks like the restructuring of the derivations is breaking the nix hydra build, which is configured via flake.nix: https://github.com/input-output-hk/formal-ledger-specifications/pull/255/checks?check_run_id=17905281422

Without spending more time on it, I'm not sure if it's just a matter of renaming - but it's probably good to fix it, if we want to keep going with hydra (which from what I understand we do)

.github/workflows/build.yml
omelkonian: Everything gets done by the Makefile; the PDFs get crated by `make docs`. This is what you would do locally, but `nix` just wraps around the Makefile to provide a reproducible environment. So, these PDFs get created by `nix` in the end, which itself calls `make docs`.

default.nix
omelkonian: The Hydra checks show a :tick: which suggests they are successful, can you direct me to the error?

.github/workflows/build.yml
teodanciu: Cool, thanks for the explanation :+1:

default.nix
teodanciu: Yes, you're right. I couldn't see any jobs building on hydra before, maybe I was looking in the wrong place, because  now I do see the derivations built! https://ci.iog.io/eval/17948
I'm still a little surprised that it works, because flake.nix references `executableSpec` which seems to no longer mean anything..but then again, I don't understand how hydra works..

teodanciu (approved):
Looks good to me

default.nix
omelkonian: Good catch actually, I've fixed the name just to be safe.

PR #256, opened 2023-10-20
Agda 2.6.4 migration

UlfNorell:
These are the minimal changes required to build with Agda 2.6.4.

I am clueless when it comes to nix, so I need some help:
- I updated the commit hashes for stdlib and stdlib-meta in `default.nix` but it didn't complain about the sha256 hash being the same as before.
- I don't know how to get Agda 2.6.4 in the nix shell

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -16,7 +16,7 @@ let
     src = fetchFromGitHub {
       repo = "agda-stdlib";
       owner = "input-output-hk";
-      rev = "bdfab77b179c937856c49d72321ca26e2a27d568";
+      rev = "cb72ba52dfbd4f83d1034e352eb88550a3e1f681";
       sha256 = "+PMZjmMK5xz+3Va7RN1ErtQghzpUlzsc9JBUoL+iasc=";
     };
   });
@@ -27,7 +27,7 @@ let
     src = fetchFromGitHub {
       repo = "stdlib-meta";
       owner = "input-output-hk";
-      rev = "897555aefd77c83e88492255641da3bcc4879461";
+      rev = "f434542c4baf667805161eeb35e5ec772233e180";
       sha256 = "EIKcLjGdnI/6nSCI18v7kPxoojAWsz8O5d9wwol2M0w=";
     };
     meta = { };
--- a/default.nix
+++ b/default.nix
@@ -17,7 +17,7 @@ let
       repo = "agda-stdlib";
       owner = "input-output-hk";
       rev = "cb72ba52dfbd4f83d1034e352eb88550a3e1f681";
-      sha256 = "+PMZjmMK5xz+3Va7RN1ErtQghzpUlzsc9JBUoL+iasc=";
+      sha256 = "+OByLIWv+pdHvWt41hniE4oeo2DJZewRyYYmNXvCix0=";
     };
   });
 
@@ -28,7 +28,7 @@ let
       repo = "stdlib-meta";
       owner = "input-output-hk";
       rev = "f434542c4baf667805161eeb35e5ec772233e180";
-      sha256 = "EIKcLjGdnI/6nSCI18v7kPxoojAWsz8O5d9wwol2M0w=";
+      sha256 = "e+gb3z+cTFW4QS0c/SQqnNVBxf9hGHKOZa/vSMkHDvw=";
     };
     meta = { };
     libraryFile = "stdlib-meta.agda-lib";
--- a/src/Ledger/Utxo.lagda
+++ b/src/Ledger/Utxo.lagda
@@ -30,7 +30,7 @@ instance
   _ = ExUnit-CommutativeMonoid
 
   HasCoin-Map : ∀ {A} → ⦃ DecEq A ⦄ → HasCoin (A ⇀ Coin)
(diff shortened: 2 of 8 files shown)


WhatisRT:
Yeah, the thing about the hashes is an anti-feature of nix. If nix finds something that has the hash you've provided, it doesn't actually check anything else - it just assumes it's the right thing (even if, for example, you change the version number which is literally right next to the hash). To get a proper error message, you can just change a few characters of the hash and it'll tell you what the correct one would be.

Getting Agda 2.6.4 is slightly more complicated, because nixpkgs gets its Agda from Hackage, which means you can't just tell it what hash to use. I've done it before, so I'll just add a commit on top of your PR.

WhatisRT:
Turns out that for some reason even the most recent Hackage update I could find in nixpkgs doesn't have Agda 2.6.4 (which is weird, they have a script for that and it ran less than a day ago). It also took a bit when we updated to 2.6.3. I think for now the strategy is to just wait, it shouldn't take too long.

WhatisRT:
I think this only fails because Agda isn't cached yet & takes forever to build. We were looking into getting the github action fetch from the IOG cache which gets populated by hydra, which should solve this problem.

WhatisRT:
Looks good, I think in this case I'd let CI finish just to be sure

src/Ledger/Utxo.lagda
```diff
@@ -79,7 +79,7 @@ module _ (let open Tx; open TxBody) where
   outs tx = mapKeys (tx .txid ,_) (tx .txouts)
 
   balance : UTxO → Value
-  balance utxo = Σᵐᵛ[ x ← utxo ᶠᵐ ] getValue x
+  balance utxo = indexedSumᵐᵛ ⦃ Value-CommutativeMonoid ⦄ getValue (utxo ᶠᵐ)
 
   cbalance : UTxO → Coin
   cbalance utxo = coin (balance utxo)
```
omelkonian, line 85: Did instance resolution become less useful in Agda 2.6.4, or is there another reason for these changes?

default.nix
```diff
@@ -16,8 +16,8 @@ let
     src = fetchFromGitHub {
       repo = "agda-stdlib";
       owner = "input-output-hk";
-      rev = "bdfab77b179c937856c49d72321ca26e2a27d568";
-      sha256 = "+PMZjmMK5xz+3Va7RN1ErtQghzpUlzsc9JBUoL+iasc=";
+      rev = "cb72ba52dfbd4f83d1034e352eb88550a3e1f681";
+      sha256 = "+OByLIWv+pdHvWt41hniE4oeo2DJZewRyYYmNXvCix0=";
```
omelkonian, line 20: Wouldn't it be more proper to depend on `agda/agda-stdlib#v1.7.3` and just add the changes in the IOHK fork as `.../Ext.agda` files, namely:
- [Unique seq equality = bag equality](https://github.com/agda/agda-stdlib/commit/668f0bbd498cfae253605fd7132c3b9ed52cc4ac) as file `Data/List/Relation/Unary/Unique/Propositional/Properties/WithK.agda`
- [Non-normalising monoid solver](https://github.com/agda/agda-stdlib/commit/bdfab77b179c937856c49d72321ca26e2a27d568) as file `Tactic.MonoidSolver.Ext` which also re-exports the standard `Tactic.MonoidSolver` as `solve-normalised` and `solve-normalised-macro`.

src/Ledger/Utxo.lagda
WhatisRT: This might be your fault 😂
https://github.com/agda/agda/issues/6364

default.nix
```diff
@@ -27,8 +27,8 @@ let
     src = fetchFromGitHub {
       repo = "stdlib-meta";
       owner = "input-output-hk";
-      rev = "897555aefd77c83e88492255641da3bcc4879461";
-      sha256 = "EIKcLjGdnI/6nSCI18v7kPxoojAWsz8O5d9wwol2M0w=";
+      rev = "f434542c4baf667805161eeb35e5ec772233e180";
+      sha256 = "e+gb3z+cTFW4QS0c/SQqnNVBxf9hGHKOZa/vSMkHDvw=";
```
omelkonian, line 31: I am hoping to release a 2.6.4-compatible stdlib-meta soon (also requested for another in-house product that uses stdlib-meta), but let's not block this now; I will simplify when the time comes.

default.nix
WhatisRT: I think it depends on the viewpoint. I'd like to get these in the actual stdlib at some point and in that case it's easier to do it like this.

src/Ledger/Utxo.lagda
WhatisRT: The real problem is of course that using bundled monoids as a typeclass was a bad idea from the start. I think we should switch this to an unbundled version and then I'd expect this to work again.

omelkonian (approved):
LGTM

src/Ledger/Utxo.lagda
UlfNorell: Yes, that's the very issue. I think switching to unbundled representations is the way to go.

PR #257, opened 2023-10-20
PDF cleanup: make PParams more compact

williamdemeo:
# Description

This addresses an item of issue #146 

Specifically, we reformat the `PParams` record slightly, making it more compact, so that it fits nicely on one page of the PDF.

# Checklist

- [X] Self-reviewed the diff


src/Ledger/PParams.lagda
```diff
@@ -21,14 +21,13 @@ module Ledger.PParams
   (es     : _) (open EpochStructure es)
   (ss     : ScriptStructure crypto es) (open ScriptStructure ss)
   where
+ProtVer : Set
+ProtVer = ℕ × ℕ
```
williamdemeo, line 25: `ProtVer` is retired, so it seems there's no reason to include it in the figure.

src/Ledger/PParams.lagda
```diff
@@ -47,65 +46,60 @@ record PParams : Set where
 \emph{Network group}
 \AgdaTarget{maxBlockSize, maxTxSize, maxHeaderSize, maxValSize, pv}
 \begin{code}
-        maxBlockSize       : ℕ
-        maxTxSize          : ℕ
-        maxHeaderSize      : ℕ
-        maxValSize         : ℕ
-        pv                 : ProtVer -- retired, keep for now
+        maxBlockSize maxTxSize        : ℕ
```
williamdemeo, line 49: Put multiple fields of the same type on a single line to reduce the length of the record, but not too aggressively (i.e., don't make lines too long and don't change the ordering of fields).

src/Ledger/PParams.lagda
```diff
@@ -47,65 +46,60 @@ record PParams : Set where
...
+        -- collateralPercent          : ℕ
+        maxCollateralInputs           : ℕ
 
 paramsWellFormed : PParams → Set
-paramsWellFormed pp = 0 ∉ fromList
-    ( maxBlockSize ∷ maxTxSize ∷ maxHeaderSize ∷ maxValSize ∷ minUTxOValue ∷ poolDeposit
-    ∷ collateralPercent ∷ ccMaxTermLength ∷ govActionLifetime ∷ govActionDeposit
-    ∷ drepDeposit ∷ [] )
-  × ℕtoEpoch govActionLifetime ≤ drepActivity
+paramsWellFormed pp =
```
williamdemeo, line 87: Imho, this formatting of the proof of `paramsWellFormed` is much nicer (and it would look even nicer to me if `×` were aligned with `0`, but that seems to be an unpopular style around here).

src/Ledger/PParams.lagda
WhatisRT: The type `ProtVer` isn't retired, just the protocol parameter. It's still used in the `TriggerHF` GA and `EnactState`.

src/Ledger/PParams.lagda
```diff
@@ -47,65 +46,60 @@ record PParams : Set where
...
-        -- Script
-        -- costmdls            : Language →/⇀ CostModel (Does not work with DecEq)
-        costmdls            : CostModel
-        prices              : Prices
-        maxTxExUnits        : ExUnits
-        maxBlockExUnits     : ExUnits
-        coinsPerUTxOWord    : Coin
-        -- collateralPercent   : ℕ
-        maxCollateralInputs : ℕ
+  -- Script group
```
WhatisRT, line 76: There's no such thing as a script group, see https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694#protocol-parameter-groups

WhatisRT (changes requested):

src/Ledger/PParams.lagda
williamdemeo: Yes, I know there was no such thing as a script group in the cip, but since recently there were a whole bunch of script parameters added, so...

src/Ledger/PParams.lagda
williamdemeo: ...but is it okay to leave `ProtVer` out of the figure, as long as it's still documented in the prose?

src/Ledger/PParams.lagda
```diff
@@ -44,68 +44,64 @@ record PoolThresholds : Set where
 record PParams : Set where
   field
 \end{code}
-\emph{Network group}
+\emph{Network group}\vskip-3mm
```
williamdemeo, line 47: The formatting mod `\vskip-3mm` (here and a few places below) merely puts the parameter group name closer to its parameters and helps to ensure that the entire figure fits on a single page.

src/Ledger/PParams.lagda
WhatisRT: The point is that these parameters already have groups, that's why I've linked the CIP. You should sort them into the groups they belong.

src/Ledger/PParams.lagda
WhatisRT: At some point I'd like to move `ProtVer` to some other figure where it makes more sense, but you don't have to do it now

src/Ledger/PParams.lagda
williamdemeo: @WhatisRT Thanks for the clarification.  I've put the new parameters in their rightful group.  As none of these was in the Governance group, am I correct to assume they don't need new documentation, or should I add the documentation of these parameters (from the CIP) to the pdf?

WhatisRT (approved):
LGTM

PR #259, opened 2023-10-24
README: fix links to HTML + add links to Haskell tests

omelkonian:
- HTML links changed after #255, which this hotfix remedies.
- Also include links to generated Haskell.


teodanciu (approved):
Looks good to me

PR #260, opened 2023-10-24
Conway cert features

WhatisRT:
# Description

- Check that pools that we delegate to exist. This is part of Shelley, but somewhat hidden (in the `DELEGS` inductive rule, which is analogous to `CERT` in Conway). This adds it to `DELEG`, where it belongs.
- Implement (Shelley-era) withdrawal logic and the [reward gate](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694#incentives-for-ada-holders-to-delegate-voting-stake), which was still missing. Withdrawals aren't included in `consumed` and and rewards pot isn't included in `HasCoin` of `LState`, so they don't actually add money back into the UTxO set (and it doesn't break POV).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Deleg.lagda
+++ b/src/Ledger/Deleg.lagda
@@ -24,10 +24,11 @@ data DCert : Set where
   ccreghot    : Credential → Maybe Credential → DCert
 
 record CertEnv : Set where
-  constructor ⟦_,_,_⟧ᶜ
+  constructor ⟦_,_,_,_⟧ᶜ
   field epoch  : Epoch
         pp     : PParams
         votes  : List GovVote
+        wdrls  : RwdAddr ⇀ Coin
 
 record DState : Set where
   constructor ⟦_,_,_⟧ᵈ
@@ -47,7 +48,7 @@ record GState : Set where
         ccHotKeys  : Credential ⇀ Maybe Credential
 
 record CertState : Set where
-  constructor ⟦_,_,_⟧ᶜ
+  constructor ⟦_,_,_⟧ᶜˢ
   field dState : DState
         pState : PState
         gState : GState
@@ -86,6 +87,7 @@ private variable
   pp : PParams
   vs : List GovVote
   poolParams : PoolParams
+  wdrls  : RwdAddr ⇀ Coin
 
 module _ (open PParams) where
 \end{code}
@@ -133,8 +135,8 @@ data _⊢_⇀⦇_,GOVCERT⦈_ : GovCertEnv → GState → DCert → GState → S
   GOVCERT-regdrep : let open PParams pp in
     (d ≡ drepDeposit × c ∉ dom dReps) ⊎ (d ≡ 0 × c ∈ dom dReps)
     ────────────────────────────────
-    ⟦ e , pp , vs ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
-                        ⟦ ❴ c , e + drepActivity ❵ᵐ ∪ˡ dReps , ccKeys ⟧ᵛ
+    ⟦ e , pp , vs , wdrls ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
+                               ⟦ ❴ c , e + drepActivity ❵ᵐ ∪ˡ dReps , ccKeys ⟧ᵛ
 
   GOVCERT-deregdrep :
     c ∈ dom dReps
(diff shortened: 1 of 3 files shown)


src/Ledger/Deleg.lagda
```diff
@@ -86,6 +91,7 @@ private variable
   pp : PParams
   vs : List GovVote
   poolParams : PoolParams
+  wdrls  : RwdAddr ⇀ Coin
```
williamdemeo, line 94: This looks fine.  My only concern is probably not a real concern... but I'll mention in case it's worth considering at some point.

Are long lists of private variables (hidden in the pdf doc) good practice, or at least innocuous?  Probably there's no risk that private variables could confuse the pdf reader, but I seem to recall reading in some Agda style recommendation document that, generally speaking, private variables are best limited to simple or very general type or level variables; e.g.,

```agda
private variable
  ℓ : Level
  a : A
  b : B
```
...and inhabitants of more special types should be explicit (as function arguments or module parameters).  On the other hand, I suppose if we are trying to make the pdf as concise and readable as possible, then I suppose using private variables very liberally like this makes more sense for us than for the typical Agda project.

src/Ledger/Deleg.lagda
```diff
@@ -114,9 +120,10 @@ module _ (open PParams) where
 data _⊢_⇀⦇_,DELEG⦈_ : DelegEnv → DState → DCert → DState → Set where
   DELEG-delegate :
     d ≡ requiredDeposit pp mv ⊔ requiredDeposit pp mc
+    → mc ∈ mapˢ just (dom pools)
     ────────────────────────────────
-    pp ⊢  ⟦ vDelegs , sDelegs , rwds ⟧ᵈ ⇀⦇ delegate c mv mc d ,DELEG⦈
-          ⟦ insertIfJust c mv vDelegs , insertIfJust c mc sDelegs , rwds ⟧ᵈ
+    ⟦ pp , pools ⟧ᵈᵉ ⊢  ⟦ vDelegs , sDelegs , rwds ⟧ᵈ ⇀⦇ delegate c mv mc d ,DELEG⦈
+                       ⟦ insertIfJust c mv vDelegs , insertIfJust c mc sDelegs , rwds ⟧ᵈ
```
williamdemeo, line 126:
```suggestion
                        ⟦ insertIfJust c mv vDelegs , insertIfJust c mc sDelegs , rwds ⟧ᵈ
```

(alignment)

src/Ledger/Deleg.lagda
```diff
@@ -133,8 +140,8 @@ data _⊢_⇀⦇_,GOVCERT⦈_ : GovCertEnv → GState → DCert → GState → S
   GOVCERT-regdrep : let open PParams pp in
     (d ≡ drepDeposit × c ∉ dom dReps) ⊎ (d ≡ 0 × c ∈ dom dReps)
     ────────────────────────────────
-    ⟦ e , pp , vs ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
-                        ⟦ ❴ c , e + drepActivity ❵ᵐ ∪ˡ dReps , ccKeys ⟧ᵛ
+    ⟦ e , pp , vs , wdrls ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
+                               ⟦ ❴ c , e + drepActivity ❵ᵐ ∪ˡ dReps , ccKeys ⟧ᵛ
```
williamdemeo, line 144:
```suggestion
                                ⟦ ❴ c , e + drepActivity ❵ᵐ ∪ˡ dReps , ccKeys ⟧ᵛ
```

(alignment)

src/Ledger/Deleg.lagda
```diff
@@ -245,43 +254,51 @@ instance
...
-  Computational-CERT .completeness Γ ⟦ stᵈ , stᵖ , stᵍ ⟧ᶜ
-    dCert@(deregdrep c) ⟦ stᵈ , stᵖ , stᵍ' ⟧ᶜ (CERT-vdel h)
+  Computational-CERT .completeness Γ ⟦ stᵈ , stᵖ , stᵍ ⟧ᶜˢ
+    dCert@(deregdrep c) ⟦ stᵈ , stᵖ , stᵍ' ⟧ᶜˢ (CERT-vdel h)
     with computeProof Γ stᵍ dCert | completeness _ _ _ _ h
   ... | just _ | refl = refl
-  Computational-CERT .completeness Γ ⟦ stᵈ , stᵖ , stᵍ ⟧ᶜ
-    dCert@(ccreghot c mkh) ⟦ stᵈ , stᵖ , stᵍ' ⟧ᶜ (CERT-vdel h)
+  Computational-CERT .completeness Γ ⟦ stᵈ , stᵖ , stᵍ ⟧ᶜˢ
+    dCert@(ccreghot c mkh) ⟦ stᵈ , stᵖ , stᵍ' ⟧ᶜˢ (CERT-vdel h)
```
williamdemeo, line 287: I'm just learning how to do metaprogramming in Agda, so I'm not yet sure it's possible, but it would be nice to have type classes like `HasTernaryConstructor` (and `HasQuaternaryConstructor`).  Members of this type class would be any record type with three (resp., four) fields, and presumably we could then use instance resolution for the constructor, which we would call `⟦_,_,_⟧` (resp., `⟦_,_,_,_⟧`).  Then we could get rid of the superscripts and check off one more item in the PDF cleanup issue #146 .

If you have any ideas about whether this is possible and how we might accomplish this, please share.  Otherwise, I'll continue learning about Agda's metaprogramming capabilities (probably somewhat aimlessly, until someone points me in the right direction 😆).

williamdemeo: On the other hand, there's the default constructor `mk`.  Could we be using that in place of the bold brackets and superscripts?  Of course that would be less visually appealing than our brackets, especially for the pdf. Is it possible to rename `mk` generally, instead of on a case-by-case basis (when defining a particular record)?

src/Ledger/Deleg.lagda
```diff
@@ -195,16 +204,16 @@ open Computational ⦃...⦄
...
   Computational-DELEG : Computational _⊢_⇀⦇_,DELEG⦈_
-  Computational-DELEG .computeProof pp _ = λ where
+  Computational-DELEG .computeProof ⟦ pp , pools ⟧ᵈᵉ _ = λ where
     (delegate c mv mc d) →
-      case d ≟ requiredDeposit pp mv ⊔ requiredDeposit pp mc of λ where
-        (yes p) → just (-, DELEG-delegate p)
-        (no _)  → nothing
+      case ¿ d ≡ requiredDeposit pp mv ⊔ requiredDeposit pp mc × mc ∈ mapˢ just (dom pools) ¿ of λ where
+        (yes (p₁ , p₂)) → just (-, DELEG-delegate p₁ p₂)
+        _  → nothing
```
williamdemeo, line 211:
```suggestion
        _               → nothing
```

src/Ledger/Deleg.lagda
```diff
@@ -195,16 +204,16 @@ open Computational ⦃...⦄
...
     _ → nothing
```
williamdemeo, line 212:
```suggestion
    _                    → nothing
```

src/Ledger/Ledger/Properties.agda
```diff
@@ -34,7 +34,7 @@ instance
       (tx : Tx)    (let open Tx tx renaming (body to txb); open TxBody txb)
       where
       utxoΓ = UTxOEnv ∋ record{ LEnv Γ }
```
williamdemeo, line 36:
```suggestion
      utxoΓ = UTxOEnv ∋ record { LEnv Γ }
```

williamdemeo (changes requested):
A few general comments/ideas/questions and some minor alignment suggestions.

src/Ledger/Deleg.lagda
WhatisRT: Yeah, that's the typeclass(es) that I've had in mind for #146. What you've described above should work fine. 

Metaprogramming isn't super helpful here, the only thing it could do would be to automatically generate the instances. But since those definitions should be simple one-liners anyway, there isn't much gained.

WhatisRT: Ah, but one problem with this is that it doesn't work for patterns, only for expressions. So this wouldn't actually help in this particular case, but since the STS constructors only pretend to pattern-match (but really are expressions) it works there. But Agda has some overloading with patterns, so maybe we could use overloaded patterns + some type classes to get the same effect.

src/Ledger/Deleg.lagda
WhatisRT: Thanks for catching these, my font isn't perfectly monospace so I sometimes get it off by one. Probably should fix that some time...

src/Ledger/Deleg.lagda
WhatisRT: So the issue is that we're doing pretend pattern-matching in the STS constructors. `e ⊢  s ⇀⦇ sig ⦈ s'` is meant to be read as a partial pattern, so it introduces the variables `e`, `s`, `sig`, while `s'` is an expression. This isn't a thing in Agda, so what we do instead is to introduce those variables elsewhere and then use them here. We could have done other things, where the variables are properly introduced by a pattern, but then we don't get this nice presentation for the PDF.

A cleaner alternative (in terms of the Agda code) would be to put a `∀` on each constructor that introduces all those variables - then you don't have those large `variable` blocks that affect things for the entire file. However, then we need to put that `∀` on its own line with a `[hide]` block around it, which adds a few lines for each constructor.

One other approach might be to put everything in a `module _ where` block. Then the `variable` block only affects the STS constructors and nothing after it. Maybe that's the cleanest solution...

src/Ledger/Deleg.lagda
omelkonian: Indeed doing this generically for patterns probably is not possible with the current machinery, but for expressions meta-programming comes in handy for sure:
```agda
mkRecordⁿ : List Term → Hole → TC ⊤
mkRecordⁿ xs hole = do
  def n [] ← inferType hole
    where _ → error $ "[mkRecordⁿ] can only construct values of a defined name"
  record-type recCon _ ← getDefinition n
    where _ → error $ "[mkRecordⁿ] can only construct records types"
  unify hole $ con recCon (vArg <$> xs)
macro
  ⟦_⟧ = mkRecordⁿ

  _`∷_ : Term → List Term → Hole → TC ⊤
  (x `∷ y) hole = do
    `x ← quoteTC x
    `y ← quoteTC y
    unify hole $ con (quote _∷_) (vArg `x ∷ vArg `y ∷ [])
infixr 4 _`∷_

private
  record R : Set where
    field x : ℕ; s : String
  _ = R ∋ ⟦ 0 `∷ "zero" `∷ [] ⟧
  record Q : Set where
    field x : ℕ; s : String; y : ℕ; w : String
  _ = Q ∋ ⟦ 0 `∷ "zero" `∷ 1 `∷ "one" `∷ [] ⟧
```

src/Ledger/Deleg.lagda
omelkonian: I still find the `variable` approach by far the cleanest, assuming that consistent naming is already followed properly of course.

williamdemeo (approved):

Issue #258, opened 2023-10-21
PDF cleanup (cont'd from issue #146)


polinavino:
- [ ] explain Agda notations/concepts/syntax
   - [x] `record` definitions
   - [x] `open` for records
   - [ ] `Type` versus `Set`
   - [ ] implicit arguments
   - [x] `Σ`
   - [ ] `module`
   - [ ] `data`
   - [ ] `λ where` syntax
   - [ ] type definitions that end in `-> Set` or `-> Type`; they are actually constructing a proposition where the proof is relevant. Should we be more consistent about using our `Type` alias throughout?  Should we sometimes use another alias, like `Prop`?  Would this make the document easier to read or would it create even more confusion?
- [ ] define/explain different notions of equality, including but not limited to,
   - [ ] `_≡_`
   - [ ] `_≗_`
- [ ] Replace `≡ᵉ` with `≡` (we may not be able to actually make this replacement in the Agda code, so we may have to do some preprocessing of the LaTeX)
- [ ] explain the following syntax/notation
   - [ ] `⟨_⟩`
   - [ ] `Σᵐᵛ` (fig 12)
   - [ ] `_⊎_` (fig 12)
   - [ ] `_∷ʳ_` (fig 14)
   - [ ] `_⊢_⇀⦇_,GOV⦈_ = SS⇒BSi _⊢_⇀⦇_,GOV'⦈_` (fig 15)
   - [ ] `_↾_` (unnumbered fig after fig 20)
   - [ ] `_↾'_`
   - [ ] `ifᵈ_then_else_` (unnumbered fig after fig 20)
   - [ ] `⌊_⌋`
   - [ ] `_∣_` (fig 23)
   - [ ] `_⊖_` (fig 25)
   - [ ] `_∘₂_` (fig 25)
   - [ ] `_⁻¹` (fig 35)
   - [ ] `ℚ.0ℚ` (fig 36)
   - [ ] `_ℚ.≥_` (fig 36)
- [ ] Remove/rename the following
   - [ ] `mapˢ` - since we now have a `Functor` class, we can make an instance and write `map` instead
   - [ ] Remove `_ᶠᵐ` notation from sums by introducing a typeclass
   - [ ] `∅ᵐ` -> `∅`
   - [ ] `❴_❵ᵐ` -> `❴_❵`
   - [ ] `foldr` (fig 5)
   - [ ] `≗` (fig 5)
   - [ ] `_⟪$⟫_` (fig 25)
[...]

PR #267, opened 2023-10-25
Add style guide for the PDF

WhatisRT:
# Description

# Checklist

- [x] Self-reviewed the diff


PDF-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+  usage to where they would naturally be defined
+
+## Alignment
+
+For better visual presentation, we should align figures properly. This
+might sometimes not be possible while also maintaining other
+constraints. Two cases where we usually want to align things are:
+- multiple lines of the form `a = b`, where those lines are short enough
+- a long expression of the form `x = a1 + ... + an`, broken into
+  multiple lines and align the `=` and `+` signs (also applies to any
```
williamdemeo, line 36:
```suggestion
  multiple lines and the `=` and `+` symbols are aligned (also applies to any
```

(minor grammar fix)

williamdemeo (approved):
I made one minor suggestion in the PR's "Conversation," but otherwise it looks good.

PDF-style-guide.md
```diff
@@ -0,0 +1,74 @@
+# PDF spec guidelines
+
+The purpose of the PDF specs is to be readable by a wide range of people. They should not assume
+- knowledge of common (functional) programming syntax and idioms or
```
omelkonian, line 4: This seems unrealistic, as the current PDF is littered with FP-isms; algebraic datatypes, polymorphism, pattern matching, parametrised records, and the list goes on..
And honestly, I don't even see a way forward to display what are essentially Agda programs without appealing to an FP background, at least this scale of a formalisation.

I would propose to focus on writing simple **Agda** that FP-savvy people can digest, which is itself a difficult task that pushes existing tooling to its limits and  significantly hinders proper software development practices, maintainability, and so forth.

omelkonian (changes requested):
Looks good overall, except the non-FP perspective which does not sit well with me.

PDF-style-guide.md
WhatisRT: I think we agree on the substance, just not on the wording. I'd say the previous PDF specs follow this guideline, and they do a lot of the examples you've mentioned. Note that this isn't saying: don't use any such things, it's just saying don't assume the reader is familiar with it. So for example we can use records, but then we need to say 'a record is like a product, but it's also assigning names to the individual components' somewhere.

I hope this sounds more sensible to you. Do you want to suggest a different wording that makes this clearer?

PDF-style-guide.md
WhatisRT: Actually, I've reformatted this list slightly in a different way

PDF-style-guide.md
WhatisRT: (also, none of this is new, this document is just putting how I've always written/reviewed the PDF into a cohesive document)

omelkonian: But the current document does not explain all these (e.g. there is no 'a record is like a product, but it's also assigning names to the individual components'), and I believe rightly so; if we take this to its extreme, we would need to first have a sizable FP tutorial before saying anything of interest.

WhatisRT: If you look at the Notation section of the [Shelley spec](https://github.com/input-output-hk/cardano-ledger/releases/latest/download/shelley-ledger.pdf), it's 1.5 pages of notation. I think we need to do a little more than that, but not drastically so. And especially in the case of records, I do think we should have that explanation. In fact, when I wanted to mention it like that at the Italy Agda workshop, Phil insisted on giving a proper introduction to records, just so I could use them in my presentation (while I (still) think that kind of short explanation would have been enough).

williamdemeo: @omelkonian you're right that the PDF doesn't currently explain Agda concepts/syntax/notation for things like records but there is an issue addressing that #258 ...but you also say "I believe rightly so," so I encourage you to write in the comments of that issue your feelings about this and we can discuss/argue about which Agda concepts should be explained and which ones we should leave hidden/unexplained.

Also, please, if you have time, jump in and tackle some of the items in #258 yourself! 😃 
 ...especially if you have strong opinions about how the concepts should be presented/explained. 🚀

williamdemeo: Back to the question at hand, what should we do about the claim that we don't assume familiarity with basic FP concepts?  I'd be in favor of removing that claim as I agree with @omelkonian that it is a bit unrealistic.  On the other hand, someone without any FP background could probably read the prose and ignore the code and possibly get something out of it.

When it comes to what we assume of the reader, are we thinking of a reader who could potentially understand everything in the document?

omelkonian: OK, that clears thing up: the plan is to have such a tutorial, but there has been no time to write it yet.
(@williamdemeo here still seems the appropriate place to discuss these requirements though, not #258)

What worries me is the burden on the developer; one has to always keep in mind that every programming construct/pattern they use is explained in said tutorial. Especially at the current state -- where there is no tutorial essentially (just the #258 issue to track its absence) -- that means the only choice one has if they were to follow the style-guide is to write a full-blown FP tutorial + some Agda-specific stuff.

Here's another alternative proposal: do not include this in the style guide and have a separate issue that tracks how complete the 'Tutorial' chapter is. That way, a developer/mechanizer does not this additional burden and there is a more appropriate division of labour IMO.

williamdemeo: I agree that a division of labor would be nice, especially if we had a documentation team.

WhatisRT: I wouldn't call the section on notation a FP tutorial. It should be more of a reference - if you haven't seen records before but otherwise are familiar with the concepts it relates to you should be fine. But if you don't know what a product is then you're out of luck.

A good benchmark for now will be the audit. The auditors don't know Agda, but they've reviewed the previous specs, so they're familiar with the notations there. So we need to put any Agda specifics we're using in the notation section, but we probably don't need to explain much else. Then we can use that feedback to refine this approach: if they don't understand something, we need to explain it or rewrite it in a more natural way.

Also, note that there are currently many violations of this guide, and this specifically applies to what goes into the PDF. So if you write a new piece of code, I think it's perfectly acceptable to use whatever you like, just put it in a `[hide]` block. It can potentially stay there until we need to produce a PDF spec that includes that code.

I'm happy to weaken this sentence, but I don't know to what. For example, we need to be explain `_$_` and `fold` if we use them, and I think that falls under 'common notation'. Maybe we need to make sure that this sentence is understood exclusively: we don't assume familiarity with things that exclusively relate to functional programming. I.e. functions are clearly an important concept in FP, but we still assume that the reader knows what a function is, since that's also relevant elsewhere.

omelkonian: Let's separate the issues we are facing into distinct categories:
- **general concepts:** e.g. do we need to explain what a function is? definitely not.
- **FP concepts:** e.g. do we need to explain what pattern matching is? might be worth a sentence here and there, but I would argue against trying to be thorough/complete and end up with a full-blown FP tutorial
- **FP-sms:** e.g. do we need to explain `_$_`? definitely yes 
  + NB: `fold` is probably much more fundamental that this, i.e. an FP concept
- **Agda-isms:** e.g. do we need to explain `syntax`? probably so, but always take a moment to contemplate if it's worth presenting in the first place
- **notation:** do we need to explain notation of any other kind (e.g. inference rule syntax)? sure, that makes sense

PS: It's funny that gallais' paper that just came out has many paragraphs on *Agda-isms*, which we can definitely reuse: https://arxiv.org/pdf/2310.13413.pdf

WhatisRT: Ok, I think we agree here. That list makes sense to be included in the style guide, maybe it could cannibalise the `Haskell/Agda-isms` section. I'd still like to include (potentially a variation of) the original line in question though, as motivation for that list. Do you think with those more detailed/clarified expectations we still need to modify the line, or would it then be fine?

omelkonian: That sounds reasonable, where the variation of the line tries to sum up our discussion here and all the subtle points we raised in as complete but at the same time precise fashion.

WhatisRT: I've added a table and adjusted the sentence.

omelkonian (approved):
LGTM

PR #268, opened 2023-10-27
Extract rule hypotheses meta-programatically

omelkonian:
- [new] Tactic `genPromises` in Tactic.Premises.
- Always use uncurried rules (_∙_ = _×_)
  + introduce patterns for curried versions
- Meta-programming utilities in Generics.Utils.Ext
  * e.g. `freeVars` to collect all free variables of a term


src/Generics/Utils/Ext.agda
```diff
@@ -0,0 +1,104 @@
...
+  [] → id
+  ((x , t) ∷ tel) → extendContext x t ∘ extendContextTel tel
+
+-- Reducing only definitions.
+reduceDef : Type → TC Type
+reduceDef = λ where
+  t@(def _ _) → reduce t
+  t           → return t
+
+-- Constructing list of quoted names without having to use `quote`.
```
WhatisRT, line 31: Neat!

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+  reduceRuleSyntax ty = do
+    let whitelist = Interface.STS._∙_
+                 `∷ Interface.STS.∙_
+                 `∷ Interface.STS._────────────────────────────────_
+                 `∷ Interface.STS.────────────────────────────────_
+                 `∷ Interface.STS._───────────────────────────────────────_
+                 `∷ Interface.STS.───────────────────────────────────────_
+                 -- ^ include all syntactic sugar for inference rules here
+                 `∷ []
+    onlyReduceDefs whitelist $ go GAS_LIMIT ty
```
WhatisRT, line 101: `onlyReduceDefs whitelist $ normalise ty` works too, is there a reason you didn't just do that?

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+  unbundleHypotheses = concatMap go where go = λ where
+    (def (quote _×_) (hArg _ ∷ hArg _ ∷ vArg x ∷ vArg ys ∷ [])) →
+      x ∷ go ys
+    t → [ t ]
+
+  -- (4) select only the hypotheses that are decidable
+  isDecidable? : AbsTelescope → Type → TC Bool
+  isDecidable? tel ty = runSpeculative $ (_, false) <$>
+    catchTC (checkType (quote it ∙) (tyView (tel , quote Dec ∙⟦ ty ⟧)) >> return true)
+            (return false)
```
WhatisRT, line 146: If you use the `MonadTC` typeclass from `stdlib-meta` you have the functions `runAndReset` and `isSuccessful`, and then you can write `runAndReset $ isSuccessful $ checkType (quote it ∙) (tyView (tel , quote Dec ∙⟦ ty ⟧))`. This suggests defining a function `hasInstance = runAndReset ∘ isSuccessful ∘ checkType (quote it ∙)`

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+
+  -- (5) bundle the (decidable) hypotheses back together in a single product
+  bundleHypotheses : List Type → Type
+  bundleHypotheses = λ where
```
WhatisRT, line 150: There's a function `generatePred` in `Tactic.DeriveComp` which has that function as a helper function. I think that suggests that this might be a more generally useful function.

src/Generics/Utils/Ext.agda
```diff
@@ -0,0 +1,104 @@
...
+module _ {A B : Set} {M : Type↑} ⦃ _ : Functor M ⦄ ⦃ _ : Monad M ⦄ where
+  mapM : (A → M B) → List A → M (List B)
+  mapM f []       = return []
+  mapM f (x ∷ xs) = do y ← f x; y ∷_ <$> mapM f xs
+
+  forM : List A → (A → M B) → M (List B)
+  forM = flip mapM
+
+-- Finding all instance candidates of a given type.
+findInstances : Type → TC (List Term)
```
WhatisRT, line 93: This is unused

src/Generics/Utils/Ext.agda
```diff
@@ -0,0 +1,104 @@
...
+
+open import Interface.Functor using (map; _<$>_; Type↑; Functor)
+
+open import Generics
+open import Agda.Builtin.Reflection using (getInstances)
+open import Reflection hiding (_>>=_; _>>_)
+open import Reflection.AST.Term
+
+-- N-ary extension of the TC context.
+extendContextTel : ∀ {A : Set} → Telescope → TC A → TC A
```
WhatisRT, line 20: This is in stdlib-meta too

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+    unquoteDecl ℚ-step-premises = genPremises ℚ-step-premises (quote ℚ.step)
+
+    _ : ℚ-base-premises .proj₁
+    _ = tt
+
+    module _ {A} ⦃ _ : Monoid A ⦄ (n m n′ m′ : A) (𝕢ˡ : ℚ n m) (𝕢ʳ : ℚ n′ m′) where
+      _ : ℚ-step-premises .proj₁
+      _ = 𝕢ˡ , 𝕢ʳ
+
+  -- * The tactic omits *undecidable* hypotheses.
```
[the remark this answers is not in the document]
omelkonian: If we require that all hypotheses are decidable, we then cannot use it in our inference rules (UTXO,UTXOW,etc..), the reason being the recursive call to other rules (e.g. UTXOW has a UTXO step as the last hypothesis).

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+    _ : 𝕎-base-premises .proj₁
+    _ = tt
+
+    unquoteDecl 𝕎-step-premises = genPremises 𝕎-step-premises (quote 𝕎.step)
+
+    module _ (n n′ : ℕ) (x y : A) (𝕣 : ℝ n n′) (𝕢 : ℚ x y) where
+      _ : 𝕎-step-premises .proj₁
+      _ = 𝕣 , 𝕢
+
+  -- * Irrelevant instance `⦃ Show A ⦄` is not included in the premise type.
```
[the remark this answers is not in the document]
omelkonian: This is crucial if we want to select the *minimal relevant context* w.r.t. the subset of (decidable) hypotheses, since there might be variables that only involve the undecidable fragment or even the conclusion.

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+```
+This meta-program then generates the type of the (decidable) hypotheses of the rule,
+as well as a proof that it is in fact decidable, i.e.
+```
+rule-premises : ∀ ⋯ → Set
+rule-premises ⋯ =
+  ∙ <HYPOTHESIS#1>
+  ∙ <HYPOTHESIS#3>
+  ⋮
+  ∙ <HYPOTHESIS#n-1>
```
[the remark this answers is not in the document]
omelkonian: It is actually deliberate, so as to demonstrate that not all of the given hypotheses make it through (i.e. depending on their decidability).

src/Tactic/Premises.agda
omelkonian: That was my first attempt, but complicated rules like UTXOW take too much time/RAM to completely normalise, so it is necessary to have a limit here.

src/Tactic/Premises.agda
omelkonian: I'm having trouble using this interface:
1. I import the MonadTC interface to use `isSuccessful` instead:
```agda
open import Interface.MonadTC.Instance using (isSuccessful)
```
But I immediately get an error:
```bash
No instance of type MonadError (List ErrorPart) (λ {v} v₁ → TC v₁)
was found in scope.
when checking that the expression isSuccessful has type
(x : TC Term) → TC Bool
```

2. I then import the MonadError interface in hopes of fixing this, but to no avail:
```agda
open import Interface.MonadError.Instance
```

What else should I import?

src/Tactic/Premises.agda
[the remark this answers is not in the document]
WhatisRT: Ah, but the recursive calls are decidable as well! See `Computational⇒Dec` in `Interface.ComputationalRelation`. I think if we turned that into an instance it should work.

src/Tactic/Premises.agda
omelkonian: Hmm, printing must have intervened when I was noticing the slow behaviour, so great! that simplified the code a lot.

src/Tactic/Premises.agda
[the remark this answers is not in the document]
omelkonian: Are you sure that we would be able to write the computational proofs then as one-liners? Sure it will work for UTXO, maybe for UTXOW, but will it work for everything else?

I am worried that we are replaying the unfortunate scenario of the `deriveComp` tactic, i.e. trying to be too general but doomed to fail in complex cases (like the proof that RATIFY is computational).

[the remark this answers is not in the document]
WhatisRT: I'm playing around with it a bit and this does indeed work, but the pattern needs to be changed a bit. I'll investigate more tomorrow.

What we could do is leave the behaviour like it is for now, and if we discover that this works indeed nicely for all computational proofs we can then modify the behaviour to give the stricter guarantees. This works just fine, since if you do make a `Dec` instance for a STS it then doesn't skip it.

src/Tactic/Premises.agda
WhatisRT: Yeah, the library design is sub-par here. The minimal thing would be this:
```
open import Interface.MonadTC using (Monad-TC; MonadTC-TC; module MonadTC)
open import Interface.MonadError using (MonadError-TC)

instance
  _ = Monad-TC
  _ = MonadTC-TC
  _ = MonadError-TC

open MonadTC ⦃...⦄ using (isSuccessful; runAndReset)
```

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+-}
+genPremises : Name → Name → TC ⊤
+genPremises f n = do
+  print $ "*** Generating premises for constructor:" <+> show n <+> "***"
+  ty ← reduceRuleSyntax =<< getType n                  -- (1)
+  let is , hs = breakImplicits $ viewTy ty .proj₁      -- (2)
+  hs′ ← map unbundleHypotheses                         -- (3)
+      $ extendContextTel (absTelescope is)
+      $ mapM reduceRuleSyntax hs
+  dhs ← forM hs′ $ λ h → (_, h) <$> isDecidable? is h  -- (4)
```
WhatisRT, line 74:
```suggestion
  dhs ← forM (unbundleHypotheses hs)
           $ λ h → (_, h) <$> isDecidable? is h        -- (4)
```

I was confused about why `reduceRuleSyntax` was used here again, since all it does is to reduce away notation that was already reduced in `(1)`. This is indeed unnecessary, and you can simplify away the whole monad dance. If you prefer not merging it into step `(4)`, you can also do `let hs = unbundleHypotheses hs`.

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+    (h ∷ hs@(_ ∷ _)) → quote _×_ ∙⟦ h ∣ bundleHypotheses hs ⟧
+
+  -- (6) minimize the implicit arguments to only the ones that are actually used in (5)
+  removeUnusedImplicits : AbsTelescope → Type → TC (AbsTelescope × Type)
+  removeUnusedImplicits is ty = go (pred $ length is) is ty
+    where
+    go : ℕ → AbsTelescope → Type → TC (AbsTelescope × Type)
+    go _ []       t = return ([] , t)
+    go n (i ∷ is) t =
+      if L.any (_≡ᵇ n) (freeVars ty) then -- relevant hypothesis
```
WhatisRT, line 162: I feel like `_∈ᵇ_` needs to be defined somewhere

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,294 @@
...
+genPremises : Name → Name → TC ⊤
+genPremises f n = do
+  print $ "*** Generating premises for constructor:" <+> show n <+> "***"
+  ty ← reduceRuleSyntax =<< getType n                  -- (1)
+  let is , hs = breakImplicits $ viewTy ty .proj₁      -- (2)
+  hs′ ← map unbundleHypotheses                         -- (3)
+      $ extendContextTel (absTelescope is)
+      $ mapM reduceRuleSyntax hs
+  dhs ← forM hs′ $ λ h → (_, h) <$> isDecidable? is h  -- (4)
+  let dhs′ = proj₂ <$> filter ((_≟ true) ∘ proj₁) dhs
```
WhatisRT, line 75: This line and the one above are really just a monadic `filter` function:
```agda

module _ {A : Set} {M : Type↑} ⦃ _ : Functor M ⦄ ⦃ _ : Monad M ⦄ where
  filterM : (A → M Bool) → List A → M (List A)
  filterM P? [] = return []
  filterM P? (x ∷ xs) = do
    b ← P? x
    xs ← filterM P? xs
    return $ if b then x ∷ xs else xs
```

If you use that, then you can just write `let dhs′ = filterM (isDecidable? is) hs`, which I think is a lot clearer.

src/Generics/Utils/Ext.agda
omelkonian: It turns out that `MonadTC.extendTel'` does not include the names which is unfortunate; I propose we fix it upstream and then remove this when we bump stdlib-meta version.

src/Tactic/Premises.agda
omelkonian: It turns out this is necessary for some reason I don't completely understand, but if you look at the generated types without the second `reduceRuleSyntax` you see an artifact of typechecking that grabs all the context (hence all variables will be considered relevant, breaking our Utxow rule for instance), namely there is a term `mkGeneraliseTel (var n) ... (var 0)` which somehow gets generated during typechecking in `src/full/Agda/TypeChecking/Generalize.createGenRecordType` if that helps.

omelkonian:
Just rebased and getting a [CI error](https://github.com/input-output-hk/formal-ledger-specifications/actions/runs/6734894682/job/18306790968?pr=268#step:10:2978) I cannot understand.

@WhatisRT any idea what causes this? probably something to do with the Agda version change and IOG caching?

PS: the error is not displayed due to unicode rendering:
```
<stdout>: commitBuffer: invalid argument (cannot encode character
'\8212')
```
which I hoped to have already fixed in the Makefile by calling agda with `--transliterate`, but I only did it for the Haskell compilation phase, maybe it is needed at all invocations?
```
-- ./src/Makefile
...
    $(AGDA) --transliterate -c --ghc-dont-call-ghc --compile-dir $(HS_DIST) $<
...
```

WhatisRT:
I think I understand what's happening: I built it locally, and since I don't have these locale issues I can see the message that triggered this. It's this one:
```
agda --html --html-dir /nix/store/0if4wyg9cvaidrp669k5pbd5646sp3zy-html-0.1/html Ledger/PDF.lagda

———— All done; warnings encountered ————————————————————————

/private/tmp/nix-build-formal-ledger-0.1.drv-0/src/Tactic/Premises.agda:98,5-19
warning: -W[no]UserWarning
DEPRECATED: Use `withReduceDefs` instead of `onlyReduceDefs`
when scope checking onlyReduceDefs
Warning: HTML is currently generated for ALL files which can be
reached from the given module, including library files.
```
Now what's the bad character? It's the em-dash `—` of course! This is the `\u8212` in the CI output. Maybe you want to define `AGDA` in the makefile to be `agda --transliterate` instead of adding it manually every time?

omelkonian:
@WhatisRT so a deprecation warning kills the CI?

WhatisRT:
Yeah, no idea why. Maybe the encoding failure makes `agda` exit with a non-zero exit code?

omelkonian:
> Maybe the encoding failure makes `agda` exit with a non-zero exit code? If `nix-build -A ledger.html` fails for you locally with the same issue, you can test it locally.

That makes sense. It is weird that if I do `nix-build -A agdaWithStdLibMeta -o ~/IOHK/agdaWithStdLibMeta` I get 2.6.3 instead of 2.6.4, do I need to change anything nix-related on my side?

WhatisRT:
Are you sure you're on the right branch? For me it uses 2.6.4 if I do that.

omelkonian:
Unrelated question, does a successful check on `ci/eval` also mean that our `Formal Ledger Specs` CI build will succeed necessarily?

src/Tactic/Premises.agda
omelkonian: @UlfNorell any idea what these `mkGeneraliseTel` terms are doing there?

UlfNorell: I think it's this: agda/agda#6951

omelkonian: Turns out it's a reported bug: https://github.com/agda/agda/issues/6951

src/Tactic/Premises.agda
```diff
@@ -0,0 +1,312 @@
...
+    t → [ t ]
+
+  -- (4) select only the hypotheses that are decidable
+  isDecidable? : AbsTelescope → Type → TC Bool
+  isDecidable? tel ty =
+    isSuccessful $ checkType (quote it ∙) (tyView (tel , quote Dec ∙⟦ ty ⟧))
+
+  -- (5) bundle the (decidable) hypotheses back together in a single product
+  bundleHypotheses : List Type → Type
+  bundleHypotheses = fold where instance _ = Monoid-Term-×
```
WhatisRT, line 120: Neat!

WhatisRT (approved):
I found one small thing, otherwise LGTM

src/Reflection/Ext.agda
```diff
@@ -0,0 +1,146 @@
...
+    `x ← quoteTC x
+    `y ← quoteTC y
+    unify hole $ con (quote _∷_) (vArg `x ∷ vArg `y ∷ [])
+
+-- Calculating all free variables of a term.
+freeVars : Term → List ℕ
+freeVars = go 0 where mutual
+  go : ℕ → Term → _
+  go bound = λ where
+    (var x as) → if bound ≤ᵇ x then [ x ] else []
```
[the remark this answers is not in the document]
omelkonian: Good catch!

PR #270, opened 2023-11-01
Added an API module to the executable spec, removed Lib

Soupstraw:
# Description

`Lib` is a bit too ambiguous to use in `cardano-ledger` conformance tests, so I exposed the `Foreign` modules directly instead. 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -71,7 +71,7 @@ rec {
 
   mkSpecDerivation = { project, main }: rec {
     docs = stdenv.mkDerivation {
-      pname = "docs";
+      pname = "${project}-docs";
       version = "0.1";
       src = "${formalLedger}";
       meta = { };
@@ -87,7 +87,7 @@ rec {
     };
 
     html = stdenv.mkDerivation {
-      pname = "html";
+      pname = "${project}-html";
       version = "0.1";
       src = "${formalLedger}";
       meta = { };
@@ -103,7 +103,7 @@ rec {
     };
 
     hsSrc = stdenv.mkDerivation {
-      pname = "hs-src";
+      pname = "${project}-hs-src";
       version = "0.1";
       src = "${formalLedger}";
       meta = { };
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -119,10 +119,13 @@ rec {
       '';
       doCheck = true;
       checkPhase = ''
-        test -n "$(find $out/haskell/ -type f -name '*.hs')"
+        test -n "$(find $out/ -type f -name '*.hs')"
```
omelkonian, line 122: I see that the Makefile still produces the Haskell source code under `$out/haskell/`, so does that mean that checking happens after `install`? In which case this block (`doCheck`+`checkPhase`) should be moved right after `installPhase`.

default.nix
```diff
@@ -119,10 +119,13 @@ rec {
...
         # OUT_DIR=$out make "${project}".hsTest
       '';
-      dontInstall = true;
+      installPhase = ''
+        mv $out/haskell/${main}/* $out
+        rm -rf $out/haskell
+      '';
```
omelkonian, line 128: NB: we now have to be careful not to generate both Ledger and MidnightExample under the same directory.

omelkonian (changes requested):

default.nix
Soupstraw: I think checking happens after building and before installing.

Soupstraw: Nix is declarative, so rearranging doesn't affect anything here.

omelkonian: So then this change seems redundant right? (since `$out/haskell` will indeed contain the generated `*.hs` files)

Soupstraw: yeah, you're right, I must've accidentally left this in

default.nix
Soupstraw: I'm not sure what you mean by that, Nix has different outputs for midnight and ledger, so that shouldn't ever happen

omelkonian: c.f. `build.yml`, where we call `nix-build` twice (once for Ledger and once for MidnightExample` but then `rsync` their output folders under a common `docs/` folder.

Soupstraw: oh, yeah, that's going to cause problems

WhatisRT:
@Soupstraw do you still plan on continuing this?

WhatisRT:
Closing this for now

PR #272, opened 2023-11-01
Expand introduction

### LINKED ISSUE #186: Flesh out state machine explanation
There's a short description about the state machine formalism in the introduction that I wrote for the workshop in Parma. We should have a full section about it, and in particular:

- explain the small-step to big-step transformer (probably with a figure)
- say a bit more about `Computational`, that we can derive it, and some of the properties it has

WhatisRT:
# Description

Closes #186 

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Introduction.lagda
```diff
@@ -38,20 +47,55 @@ initial state \(s\), a signal \(b\) and a final state \(s'\). The ledger consist
 25-ish (depending on the version) such relations that depend on each
 other, forming a directed graph that is almost a tree.
 
+\subsection{Small-step to big-step}
```
omelkonian, line 50: I am worried that the terminology here is a bit misused, what this section refers to is the *reflexive, transitive closure of the small-step semantics* but is not directly connected to an alternative presentation of the rules in a natural/big-step style of semantics.

I propose renaming things here to make the above observation explicit (maybe introduce these constructs as *traces* that represent a sequence of single steps)

WhatisRT: Taking the reflexive transitive closure is the usual way of turning a small-step semantics into a big-step semantics. We don't give a direct big-step definition of the relations, but I'm not sure we need one just to call this big-step.

Also, this isn't necessarily reflexive because of the base-case. We could just call this a 'closure' operator, and then say it's called `SStoBS` because this it resembles how you would get a big-step semantics from a small-step one.

omelkonian: You still use the term "big-step" to describe the length of a small-step derivation; this is not how it's used in the literature, namely to distinguish between two popular ways to present your operational semantics/rules either between a transition system between configurations (small-step semantics) or a direct value-centric relation that resembles a recursive interpreter (big-step/natural semantics).

Taking the closure of the small-step relation does not justify calling it a derived big-step relation; in fact it is often the case that you have both kind of semantics and you want to prove that they are equivalent (up to effects not accounted for by big-steps, e.g. convergence) and you do so by verifying that the *small-step ref.trans. closure* agrees with the *big-step relation*.

To sum up, we should never mention the term *big-step semantics* since there are no big-step-style rules that describe the ledger semantics.

WhatisRT: So is the difference here that for you small/big step is simply a style of presentation not actual definitions? Because for me, there's a category of small-step relations, a category of big-step relations and a functor which takes the reflexive-transitive closure to go from small-step to big-step (denotational semantics also fits in that picture nicely, and I completely fail to understand what people mean by denotational semantics if not that). This is just formally defined and completely independent of any style of presentation. But maybe this is indeed a niche view on the meaning of these terms.

I don't like the term _traces_ here, since it sounds historical. If we just name it _transitive closure with base case_, what would be the symbol for it?

omelkonian: Indeed that's a very specific categorical take on small-step VS big-step, probably worked out by people in the literature since there are tons of papers relating different kind of semantics, bidirectional translations, and so forth. Anyway, I would steer clear of this approach for the purposes of this document.

I don't get your last question; currently we use `_⊢_⇀⟦_⟧*_` which is indeed confusing if we have a non-reflexive base (btw, what are the actual use cases for this base relation). Do you seek a more appropriate notation for this relation?

omelkonian (approved):
Looks good for now, but I still support the idea of having explicit constructs from Kleene algebra (*, +, ;, ε, maybe also ∪ for choice to have the full regex) to express this; e.g. for CERTS:
```agda
_⊢_⇀⦇_,CERTS⦈_ = (_⊢_⇀⦇_,CERT⦈_ ⋆) ⨾ _⊢_⇀⦇_,CERTBASE⦈_
```

I think the PDF will look much nicer too and more understandable to people with the relevant background knowledge.

WhatisRT:
Yeah, it's definitely worth exploring, I just wanted to go with something safe for now. I made an issue.

Issue #273, opened 2023-11-01
Clean up and expose PParamsDiff


WhatisRT:
We should say something about the type of protocol parameter updates in the PDF, so we should expose `PParamsDiff` in some manner. In particular:

- [x] `ppdWellFormed` should return a decidable type, not a `Bool`
- [x] We're currently assuming `ppdWellFormed` and requiring it to
  have some properties, but there is no reason those properties can't
  just be its definition.

PR #274, opened 2023-11-01
Require `ppdWellFormed` to imply non-empty `updateGroups`

### LINKED ISSUE #271: Prevent empty PParam update proposals
See https://github.com/input-output-hk/cardano-ledger/pull/3836

We should restrict the well-formedness to only hold when we can actually assign a group.

WhatisRT:
# Description

Closes #271. We should probably clean this thing up and expose it in the PDF soon (#273)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -156,11 +156,12 @@ HsGovParams : GovParams
 HsGovParams = record
   { Implementation
   ; ppUpd = let open PParamsDiff in λ where
-      .UpdateT          → ⊤
-      .updateGroups     → λ _ → ∅
-      .applyUpdate      → λ p _ → p
-      .ppdWellFormed    → λ _ → true
-      .ppdWellFormed⇒WF → λ _ _ x → x
+      .UpdateT                → ⊤
+      .updateGroups           → λ _ → ∅
+      .applyUpdate            → λ p _ → p
+      .ppdWellFormed          → λ _ → false
+      .ppdWellFormed⇒hasGroup → λ ()
+      .ppdWellFormed⇒WF       → λ _ _ x → x
   ; ppHashingScheme = it
   }
 
--- a/src/Ledger/PParams.lagda
+++ b/src/Ledger/PParams.lagda
@@ -135,9 +135,10 @@ record PParamsDiff : Set₁ where
         updateGroups : UpdateT → ℙ PParamGroup
         applyUpdate : PParams → UpdateT → PParams
         ppdWellFormed : UpdateT → Bool
-        ppdWellFormed⇒WF : ∀ {u} → ppdWellFormed u ≡ true → ∀ pp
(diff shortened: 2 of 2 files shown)


src/Ledger/Foreign/HSLedger.agda
```diff
@@ -156,11 +156,12 @@ HsGovParams : GovParams
 HsGovParams = record
```
williamdemeo, line 156: Why is `HsGovParams` not called `HSGovParams`?

None of the other types starts with `Hs` (i.e. `H` and *lower-case* `s`).

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -156,11 +156,12 @@ HsGovParams : GovParams
...
   { Implementation
```
williamdemeo, line 157: Oh, I wasn't familiar with this syntax.  So, does this open up the `Implementation` module inside the record definition?  ...cool!

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -156,11 +156,12 @@ HsGovParams : GovParams
...
-      .UpdateT          → ⊤
-      .updateGroups     → λ _ → ∅
-      .applyUpdate      → λ p _ → p
-      .ppdWellFormed    → λ _ → true
-      .ppdWellFormed⇒WF → λ _ _ x → x
+      .UpdateT                → ⊤
+      .updateGroups           → λ _ → ∅
+      .applyUpdate            → λ p _ → p
+      .ppdWellFormed          → λ _ → false
+      .ppdWellFormed⇒hasGroup → λ ()
```
williamdemeo, line 163: Question: I see that Agda fills in the proof (`λ ()`) for us, as the antecedent (`false ≡ true`) is false.  However, I thought we could also prove it explicitly using `⊥-elim`, but somehow that's not working for me.    I would have thought something like `λ f≡t → ⊥-elim f≡t` (which is obviously just the function `⊥-elim`) would work.  The proof `λ ()` works here and I think that means the antecedent is equivalent to `⊥`.  What am I missing?

williamdemeo (approved):
Just a couple of minor questions/comments... but nothing crucial.  LGTM!

src/Ledger/Foreign/HSLedger.agda
WhatisRT: This is simply notation for an empty pattern lambda, so this is almost no magic. If you pattern-match on the proof that `false = true`, `refl` doesn't work so you don't need a body in the pattern lambda.

src/Ledger/Foreign/HSLedger.agda
WhatisRT: I didn't touch that part in this PR. Since we'll probably be having a bunch of changes in that file soon I'm not going to bother to fix that inconsistency now.

PR #275, opened 2023-11-01
PDF cleanup: explain Agda records

williamdemeo:
# Description

This addresses one item of #258.

# Checklist

- [X] Self-reviewed the diff


src/Ledger/EssentialAgda.lagda
```diff
@@ -0,0 +1,75 @@
...
+p23 : Pair ℕ ℕ
+p23 = record { fst = 2; snd = 3 }
+\end{code}
+%
+Here's an alternative syntax for defining the above record.  (Note: we must \AgdaKeyword{open}
+the record type to bring its components---in this case, the field names \AgdaField{fst} and
+\AgdaField{snd}---into scope.)
+%
+\begin{code}
+open Pair
```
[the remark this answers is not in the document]
williamdemeo: One of the items in the issue (#258) is to explain `open` for records.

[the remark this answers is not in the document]
WhatisRT: Yes, but this isn't what that's referring to. What we use in a bunch of places (and I'm not sure if it's a good idea to keep) is the equivalent of `open Pair p23`, i.e. `open` combined with a module application. That's what we have to explain. `open` without module application is just noise.

src/Ledger/PDF.lagda
```diff
@@ -68,4 +70,11 @@ open import Ledger.Chain
 \section{Properties}
 \input{Ledger/Utxo/Properties}
 
+\clearpage
+
+\appendix
+
+\section{Appendix: Agda essentials}
```
WhatisRT, line 77: I think this should go into the `EssentialAgda` file (I know that `Properties` above has the same issue, and we should fix that at some point as well).

src/latex/agda-latex-macros.sty
```diff
@@ -3,6 +3,8 @@
 \RequirePackage{xspace}
 \RequirePackage{pifont}  %% needed for checkmark (\ding{51}).
 
+\newcommand\agdartdlang[1]{\href{https://agda.readthedocs.io/en/v2.6.4/language/#1}{agda.readthedocs.io\slash...\slash#1}\xspace}
```
WhatisRT, line 6: This looks broken in the PDF (I only know that the look is intended by reading this line). Can you make a proper bibliography that includes this instead?

WhatisRT (changes requested):
Just a few minor things, but I'd rather have them included here so they don't get forgotten later.

src/Ledger/PDF.lagda
williamdemeo: I'm assuming you meant that, of the four lines 74--77, only line 77 should be moved to `EssentialAgda.lagda`.  That's what will take effect in my next commit to this PR.

src/latex/agda-latex-macros.sty
williamdemeo: It doesn't seem broken to me.  Perhaps the way I had intended it to look appears broken to you.  I can change the style to conform to that of the link in the first section of the pdf.  I think a proper bibliography should be a new issue/PR since it involves touching other files as well.

williamdemeo:
@WhatisRT I think this PR is good to go now.

WhatisRT (approved):
LGTM

PR #284, opened 2023-11-07
Switch to stdlib-2.0

### LINKED ISSUE #262: Agda stdlib 2.0
Now that we'll soon be on Agda 2.6.4, and the stdlib 2.0 is getting close to being released, it makes sense to port our code to the new stdlib. This has some benefits:

- Some of the code we rely on is currently not in any released version of the stdlib. I haven't checked, but I think this should change.
- We could use this as an opportunity to try and get as much of our 'should have been in the stdlib' code into the stdlib as possible. Then we could potentially drop our custom patched version.

UlfNorell:
This updates our stdlib dependency to the 2.0 release candidate.

Fixes #262


omelkonian (approved):
Yaay, fork exterminated!

WhatisRT (approved):
LGTM

Issue #288, opened 2023-11-07
Various cleanup items for the Conway spec


WhatisRT:
This issue is for things that are necessary specifically for the Conway spec and don't fit elsewhere.

- [x] ~Fig. 8~ Fig. 13 (Protocol parameter declarations)
      Can we move `Acnt` elsewhere?
- [x] ~Fig. 14~ Fig. 12 has a lot of rules that don't have any
      preconditions. Explain that not all rules have a horizontal line
      and preconditions (or just make one so it looks consistent).
- [x] ~Fig. 15~ Fig. 13 - format
- [x] ~Fig. 17~ Fig. 15 - rules would be easier to read if all the `let`
      statements are above the line
- [ ] ~Fig. 40~ Fig. 43 (EPOCH transition system)
      The module shenanigans and record updates make this one
      especially hard to read

williamdemeo:
Please correct me if I'm wrong, but I believe Fig 14 is now Fig 12 ("ENACT transition system")  (There are no horizontal lines or rules in Fig 14.)

williamdemeo:
I think other Fig numbers might be wrong, too.

williamdemeo:
@WhatisRT do you have any ideas re. where we might move `Acnt`?

WhatisRT:
Most items I just copied over from the old issue, and I readjusted the numbers to fit with the PDF I had (which I thought was the most recent, but maybe I was wrong). I think the item about fig. 13 probably refers to the figure being too long - removing `Acnt` fixes that (it may still be a little too long, but at least it doesn't overlap with the page number anymore).

Ideally, I'd move `Acnt` to some place where we actually use it. I wouldn't mind it being next to the definition of `NewEpochState`.

williamdemeo:
We need to clarify the item "Fig 13 format."

williamdemeo:
> Ideally, I'd move `Acnt` to some place where we actually use it. I wouldn't mind it being next to the definition of `NewEpochState`.

@WhatisRT `Acnt` was moved into the hidden code block just above where it previously stood.  This was done in the (merged) PR #292, where I added the following explanation to the text of the PDF (so that hiding the definition doesn't confuse the reader):

"The \AgdaRecord{Acnt} record has two fields, \AgdaField{treasury} and \AgdaField{reserves}, so the \AgdaBound{acnt} field in \AgdaRecord{NewEpochState} keeps track of the total assets that remain in treasury and reserves."

[...]

PR #290, opened 2023-11-07
Various cleanup items: Fig 12 -- axiomatic rules

williamdemeo:
# Description

This addresses one issue of #288.

We could explain that some rules have no premise (they're essentially axioms). Instead, for consistency, we simply define a horizontal line that takes no premise, introducing syntax to represent such "axiomatic" rules.

# Checklist

- [X] Self-reviewed the diff


omelkonian:
This is somewhat overlapping with #268 , which also introduces the "axiom" syntax and a slightly different represent of premises (= uncurried = single argument as a tuple of premises).

omelkonian:
@williamdemeo Since #268 will be merged today probably, I would suggest rebasing now so that you don't spend time repeating the same definitions or resolving merge conflicts.

WhatisRT:
Indeed, I think this is the correct approach, but it's probably best to get in #268 first

williamdemeo:
Okay, I think we're good to go.  @WhatisRT please take a look so we can GTM.  Thanks!

WhatisRT (approved):
LGTM

Issue #289, opened 2023-11-07
Other PDF cleanup items


WhatisRT:
This issue contains various items that don't really fit anywhere else. I've divided this into smaller and bigger tasks. None of this is really urgent ATM, so we can do these once we have some more time.

Bigger tasks:
- [x] Remove `_ᶠᵐ` notation from sums by introducing a typeclass
- [x] #652
- [x] Can we remove `+ᵉ` somehow? We could change `HasAdd` so that `+`
      can take to different types, and then for a ring define some
      instances where you can always add natural numbers/integers. But
      maybe we can do something easier?
- [x] #656

Smaller tasks:
- [x] Fig. 5 (Token algebras, used for multi-assets): 
   - [x] missing type or explanation of `PolicyID`
   - [x] the two `field` keywords are misaligned, so may be unclear that all fields belong to `TokenAlgebra` record type.
   - [ ] We have some levels in the figure that we should get rid of
   - [x] `MemoryEstimate` doesn't need to be defined inside a `TokenAlgebra`, so we should move it outside of it (otherwise people may think there's a reason why it's defined right there)
- [ ] Fig. 6 (Definitions used in Addresses):
   - [ ] `isVKey` should be defined
   - [ ] `isScript` should be defined
- [ ] why is `Γ ⊢` used instead of say, `_ ⊢` when environment is irrelevant?
- [ ] Fig. 22 (Definitions used in the UTxO transition system):
   - [ ] different font for abstract types than Fig. 6
   - [x] figure doesn't fit on the page
   - [x] `mapPartial` needs to be defined
   - [x] `isInj_1` needs to be defined
- [ ] Fig. 23 (Functions used in UTxO rules): define `mapKeys`
- [ ] Fig. 28 (Functions used for witnessing):
- [ ] Fig. 30 (UTXOW inference rules):
   - [ ] `txidBytes` should be defined
   - [ ] `hash` should be defined
- [x] Fig. 21 (Functions used in UTxO rules) is too big
- [x] Change notation for summation from `Σ` to `∑`

WhatisRT:
@williamdemeo Please work on #282 and #285 first. These are critical, whereas the things in here are really more nice-to-have.

williamdemeo:
I checked as completed the items
- [X] `MemoryEstimate` doesn't need to be defined inside a `TokenAlgebra`
- [X] `mapPartial` needs to be defined

[...]

PR #293, opened 2023-11-07
Other pdf cleanup items: change summation symbol

williamdemeo:
# Description

This addresses one item of issue #289

Hopefully this change will make it less likely that a reader misinterprets the symbols we use for summations as dependent pair symbols.

# Checklist

- [X] Self-reviewed the diff


src/Axiom/Set/Sum.agda
```diff
@@ -156,5 +156,5 @@ module _ ⦃ _ : DecEq A ⦄ ⦃ _ : DecEq B ⦄ where
             helper : toRel m ≡ᵉ toRel (m₁ ∪ˡᶠ m₂)
             helper = ≡ᵉ.trans (proj₁ m≡m₁∪m₂) (≡ᵉ.sym $ disjoint-∪ˡ-∪ disj-dom')
 
-  syntax indexedSumᵐ  (λ a → x) m = Σᵐ[ a ← m ] x
-  syntax indexedSumᵐᵛ (λ a → x) m = Σᵐᵛ[ a ← m ] x
+  syntax indexedSumᵐ  (λ a → x) m = ∑ᵐ[ a ← m ] x
+  syntax indexedSumᵐᵛ (λ a → x) m = ∑ᵐᵛ[ a ← m ] x
```
omelkonian, line 160: Grepping reveals one last remaining occurrence of summation using "sigma", namely in the syntactic sugar of `indexedSumL`.

omelkonian (approved):
LGTM, but please also replace the last occurrence in `Axiom.Set.Sum`.

src/Axiom/Set/Sum.agda
williamdemeo: Thanks for catching that... not sure how I missed it. 😅

PR #298, opened 2023-11-10
Other PDF cleanup items: Fig 5 (Token algebras)

williamdemeo:
# Description

Address items in issue #289.

- [X] align the `field` keywords in the `TokenAlgebra` record type.
- [X] say what `PolicyID` is for.
- [X] add `[h!]` to figure environment so the figure does not appear above the section heading ("4 Token algebras").

# Checklist

- [X] Self-reviewed the diff


src/Ledger/TokenAlgebra.lagda
```diff
@@ -53,5 +54,6 @@ record TokenAlgebra : Set₁ where
   sumᵛ : List Value → Value
   sumᵛ = foldr _+ᵛ_ (inject 0)
 \end{code}
-\caption{Token algebras, used for multi-assets}
+\end{AgdaAlign}
+\caption{Token algebras, used for multi-assets; \AgdaBound{PolicyId} specifies a type (e.g., with natural numbers) used to represent policy identifiers.}
```
WhatisRT, line 58: We already have a pattern for this, see the figure right below (Definitions used in Addresses). Maybe that pattern isn't that great (if you don't like it, make an issue to change it everywhere), but this should still follow it otherwise it's confusing (is `PolicyID` not an abstract type?).

WhatisRT (changes requested):

src/Ledger/TokenAlgebra.lagda
williamdemeo: Okay, I'll check out the existing pattern and conform.

No, `PolicyID : Type` is a parameter passed into this (the `TokenAlgebra`) module.

williamdemeo:
@WhatisRT I think this PR is good to go now.

PR #300, opened 2023-11-10
Remove custom caching

WhatisRT:
# Description

Since we can just fetch our stuff from hydra now, we don't need to maintain a cache ourselves anymore. This should also speed up the action a bunch.

This removes freeing up space, fetching the cache and adding to the cache, which took around 5:30 min. This also reveals that our current HTML derivations aren't cached by Hydra, which eats up 2:30 min of that gained time, but only on PRs that don't change any Agda code. So this should almost always save us the full 5ish minutes.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - run: sudo apt-get update
-
       - uses: actions/checkout@v3
 
       - uses: cachix/install-nix-action@v20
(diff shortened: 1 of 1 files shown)


WhatisRT:
I wonder if we could have this action maybe check if everything is cached, and only continue after 10 (or some other number of) minutes or if the derivations exist in the cache. That way, if Hydra works, this action would always finish shortly after Hydra instead of looking there once, and then doing all the work itself if the cache lookup fails. Might be a nice future improvement.

teodanciu (approved):
Good bye, baby cache!  🥲
Looks good to me

omelkonian (approved):
Sorry for missing this, LGTM.

WhatisRT:
I looked at the diff between the two branches (before rebasing), and the only real difference was that I'm also removing `sudo apt-get update`. I think that also saves a little extra time, since we now shouldn't have a reason to do that anymore.

Issue #282, opened 2023-11-07
Explain/remove concepts/notations


WhatisRT:
## Unavoidable (Do later)

Things that we can either not avoid for technical reasons or because they would make things massively more difficult. These need to be explained.

- [x] `record` definitions, `record` syntax for (de)constructing values
- [ ] #444

## Potentially avoidable (Do later)

Things that can be avoided/hidden at some cost. That cost needs to be weighed against the cost of adding an explanation (which also makes the document less accessible).

- [x] `open` for records
- [x] implicit arguments
- [x] `Σ`
- [x] `data`
- [x] `λ where` syntax
- [x]  Copatterns

## Should be avoided (Do soon)

These are things which aren't too difficult to avoid, are completely custom or only used in an FP context. Those need to have a very good reason to pay the cost of adding an explanation/increasing the barrier to entry.

- [x] `module` syntax
- [x] `_≗_` (fig 5)
- [x] ~`_⟪$⟫_` (fig 25)~ (afaik, this no longer appears in fig 25)
- [x] `_<$>_` (fig 25)
- [x] `sp-∘`
- [x] `to-sp`
- [x] `uncurry`
- [x] `foldr`

## Need explanations/cleanup (Do soon)

Things that definitely we want to use but still should have a (potentially very short) explanation. Some things in here may also be improved by being renamed. Some renames may require type classes or other mechanisms.

Name is probably good:
- [x] `_⊎_`
- [x] `_∷ʳ_`
- [x] `_⁻¹`
- [x] `_∣_`

May benefit from a rename:
- [ ] `_↾_`
- [ ] `_↾'_`
- [x] `_⊖_`
- [x] `_∘₂_`

Needs rename:
- [x] `Σᵐᵛ` -> `∑`, maybe `∑ᵛ`?
- [x] ~`mapˢ` - since we now have a `Functor` class, we can make an
      instance and write `map` instead~ (this is difficult, see #322)
- [x] `∅ᵐ` -> `∅`
- [x] `❴_❵ᵐ` -> `❴_❵`
- [x] `ℚ.0ℚ` -> `0ℚ` (`0ℚ` -> `0` via literal overloading later)
- [x] `_ℚ.≥_` -> `_≥_` or `_≤_`
- [x] `filterᵐ` -> `filter` (or maybe something else?)
- [x] `sucᵉ` -> `... + 1`
- [ ] #446

WhatisRT:
[...]

PR #305, opened 2023-11-14
Rename ∅ᵐ → ∅

williamdemeo:
# Description

Addresses one item of #282.

The task is to rename `∅ᵐ` as `∅` using type classes.

# Checklist

- [X] Self-reviewed the diff


src/Axiom/Set.agda
```diff
@@ -167,13 +167,13 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-unions : {a : A} {U : Set (Set A)} → (∃[ T ] T ∈ U × a ∈ T) ⇔ a ∈ proj₁ (unions U)
   ∈-unions = proj₂ $ unions _
 
-  ∅ : Set A
-  ∅ = fromList []
+  ﹛﹜ : Set A
```
williamdemeo, line 170: These are small curly braces obtained from `\(` (or `\)`) and then hitting `C-f` a bunch of times.

WhatisRT (changes requested):
Sorry, the intent here was that both the empty set and empty map would be denoted by `∅`, via a typeclass.

williamdemeo:
@WhatisRT the renaming now uses type classes so that both empty sets and empty maps can be called `∅`. It seems acceptable for now, but I suspect I could have arranged it so that I don't need to import the type classes whenever I use them; e.g., by putting the instances inside the `Set` or `Map` module.  (I did try that initially, but I couldn't get it to work because of circular dependencies.  When more of the items in the issue #282 are done, I'll revisit this.)

src/Axiom/Set.agda
```diff
@@ -167,13 +167,13 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-unions : {a : A} {U : Set (Set A)} → (∃[ T ] T ∈ U × a ∈ T) ⇔ a ∈ proj₁ (unions U)
   ∈-unions = proj₂ $ unions _
 
-  ∅ : Set A
-  ∅ = fromList []
+  ∅ˢ : Set A
+  ∅ˢ = fromList []
 
-  ∅-strongly-finite : strongly-finite {A} ∅
+  ∅-strongly-finite : strongly-finite {A} ∅ˢ
```
williamdemeo, line 173: The goal of this PR is to rename `∅ ` to `∅`.  To that end, I had to change `∅` to `∅ˢ` in this file (and in a few other places) to accommodate the new type classes `HasEmptySet` and `HasEmptyMap`.  Fortunately, I am able to limit the superscripts to blocks of code in `.agda` files or *hidden* blocks in `.lagda` files.

src/Interface/HasEmptySet.agda
```diff
@@ -0,0 +1,11 @@
...
+
+open import Axiom.Set using (Theory)
+
+module Interface.HasEmptySet {ℓ} (th : Theory {ℓ}) where
+
+open Theory th using () renaming (Set to ℙ_)
+
+record HasEmptySet (A : Set ℓ) : Set ℓ where
+  field
+    ∅ : ℙ A
```
WhatisRT, line 11:
```suggestion
record HasEmptySet (A : Set ℓ) : Set ℓ where
  field
    ∅ : A

instance
  _ : {A : Set} → HasEmptySet (ℙ A)
  _ = record { ∅ = ∅ˢ }

  _ : {A B : Set} → HasEmptySet (Map A B)
  _ = record { ∅ = ∅ᵐ }
```

WhatisRT (changes requested):
This is much better, but you don't need two type classes. I've made a suggestion below, and then you can just use `∅` if you reexport `Interface.HasEmptySet` from `Ledger.Prelude`

src/Interface/HasEmptySet.agda
williamdemeo: Okay, thanks, I'll try it.

williamdemeo: @WhatisRT This commit creates a lot of cyclic module dependencies in the set theory files, but since those are `.agda` files and won't show up in the pdf, I can choose to not import `HasEmptySet` there and simply add superscripts to the emptyset symbol in those modules.  (HasEmptySet now needs to import both `Axiom/Set` *and* `Axiom/Map`, so these cannot use `HasEmptySet`.)

That's fine, but now the problem is that `Set` has type `Type l -> Type l` whereas `Map` has type `Type -> Type -> Type`, so the levels don't line up in your unified module `HasEmptySet {l}`.  I'll keep working on it and try to resolve this.

williamdemeo: I think I figured it out.  Will push the fix soon.

src/Axiom/Set/Map.agda
```diff
@@ -58,7 +58,7 @@ record IsLeftUnique (R : Rel A B) : Type where
   field isLeftUnique : left-unique R
 
 instance
-  ∅-left-unique : IsLeftUnique {A = A} {B = B} ∅
+  ∅-left-unique : IsLeftUnique {A = A} {B = B} ∅ˢ
```
williamdemeo, line 61: It should be okay to leave superscripts on the emptyset symbol in `.agda` files, since they won't appear in the pdf.

williamdemeo: Also, in some cases---namely, in the submodules of `Axiom.Set`---we *must* use superscripts; we can't use the type classes there because that causes a cyclic module dependency since the type class imports `Axiom.Set` and `Axiom.Map`.

src/Ledger/Deleg.lagda
```diff
@@ -10,7 +10,7 @@ module Ledger.Deleg (gs : _) (open GovStructure gs) where
 
 open import Ledger.GovernanceActions gs
 \end{code}
-\begin{figure*}[h]
+\begin{figure*}[h!]
```
williamdemeo, line 13: Fixing figure placement; I don't think this deserves its own PR.

src/Ledger/Transaction.lagda
```diff
@@ -151,17 +151,14 @@ the transaction body are:
...
-\emph{Abstract functions}
-\begin{code}
 \end{code}
 \end{AgdaSuppressSpace}
 \caption{Definitions used in the UTxO transition system}
 \label{fig:defs:utxo-shelley}
 \end{figure*}
 
 \begin{figure*}[h]
+\emph{Abstract functions}
```
williamdemeo, line 161: Fixing figures; I don't think this deserves a separate PR.

WhatisRT (approved):
I've added a cleanup commit on top of this

PR #306, opened 2023-11-14
Rename ❴_❵ᵐ → ❴_❵

williamdemeo:
# Description

Addresses one item of #282: rename `❴_❵ᵐ` → `❴_❵`. 

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Axiom/Set/Map.agda
+++ b/src/Axiom/Set/Map.agda
@@ -146,7 +146,7 @@ filterKeys : {P : A → Type} → specProperty P → Map A B → Map A B
 filterKeys sp-P = filterᵐ (sp-∘ sp-P proj₁)
 
 singletonᵐ : A → B → Map A B
-singletonᵐ a b = singleton (a , b)
+singletonᵐ a b = ❴ (a , b) ❵
                , (from ∈-singleton -⟨ (λ where refl refl → refl) ⟩- from ∈-singleton)
 
 ❴_❵ᵐ : A × B → Map A B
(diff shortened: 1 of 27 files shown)


WhatisRT (changes requested):
Same here as with #305, this should be done with a typeclass.

src/Axiom/Set/Map.agda
```diff
@@ -146,7 +146,7 @@ filterKeys : {P : A → Type} → specProperty P → Map A B → Map A B
 filterKeys sp-P = filterᵐ (sp-∘ sp-P proj₁)
 
 singletonᵐ : A → B → Map A B
-singletonᵐ a b = ❴ (a , b) ❵
+singletonᵐ a b = singleton (a , b)
```
williamdemeo, line 149: The goal of this PR is to rename the curly braces used for singletons using a type class (called `HasSingleton`).  However, in the `.agda` files that are used by that type class, we will not use the undecorated curley braces because that would cause a cyclic dependency.  So we use `singleton` for what previously was called `❴_❵`.

WhatisRT (changes requested):
Similar to the other PR, you don't need to change anything in `Axiom.*` and the instances should just be available together with the class. See this commit: b48799f79221e315397e11d48db727977d119bf5

williamdemeo:
@WhatisRT I think this one is good to go. ⚡

WhatisRT (approved):
LGTM

PR #307, opened 2023-11-14
Rename `ℚ.0ℚ` → `0ℚ` and `_ℚ.≥_` → `_≥_`

williamdemeo:
# Description

Partial addresses one item, fully addresses another item, of #282

- [ ] rename `ℚ.0ℚ` → `0`
- [X] rename `_ℚ.≥_` → `_≥_`

The second is easy because we can avoid geq for natural numbers by hiding it in the `Ledger.Prelude` import.

The first task is harder because we use the natural zero and the rational zero in the same line.  It's easy to change from `ℚ.0ℚ` to `0ℚ`, however, so that's what I did.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Ratify.lagda
+++ b/src/Ledger/Ratify.lagda
@@ -80,11 +80,11 @@ above the threshold.
 {-# OPTIONS --safe #-}
 
 import Data.Integer as ℤ
-open import Data.Rational as ℚ using (ℚ; 0ℚ)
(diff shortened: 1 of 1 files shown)


WhatisRT (approved):
Yeah, for changing `0ℚ` to `0` we need literal overloading. We can do that later.

PR #308, opened 2023-11-14
add syntax for a map that models a property

williamdemeo:
# Description

Addresses one item of #282 

In particular, instead of renaming `filter^m`, we add syntax so that `filter^m sp m` can be denoted by `m \vDash sp`.

# Checklist

- [ ] Self-reviewed the diff


WhatisRT:
I think this notation is more confusing than just writing `filterᵐ`. We need to have less uncommon notations, not more.

PR #301, opened 2023-11-10
[À bas with forks] depend on tagged versions

omelkonian:
- Depends on version v2.0-rc1 of everything:
  + agda-stdlib (already from PR #284)
  + agda-stdlib-classes
  + agda-stdlib-meta
- [nix] rename `agdaWithStdlibMeta` to `agdaWithDeps`
- [.agda-lib] remove unnecessary flag `-WnoUnsupportedIndexedMatch`

- [decidability] completely remove the ad-hoc `Interface.Decidable.Instance` and uses the provided `Class.Decidable`; the breaking change here is that decidability proofs are now wrapped around the newtype _⁇

  Hence this PR goes a long way w.r.t. to issues #36 and #281, but still need a separate PR to completely close these.

# Description

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages
- [ ] Code is formatted according to [CONTRIBUTING.md](https://github.com/input-output-hk/formal-ledger-specifications/blob/master/CONTRIBUTING.md)
- [ ] Self-reviewed the diff

Issue #281, opened 2023-11-07
Decidability in the PDF


WhatisRT:
We have various different notations for decidability and related functions. We need to figure out a consistent approach to deal with it and make sure whatever we do is clear in terms of presentation in the PDF. I.e. if we do pick an approach that relies on custom notation, that needs to be explained. If our approach is completely transparent in terms of notation, we obviously don't need to explain what isn't there.

Here are some of the notations currently in use:
- `¿_¿`
- `¿_¿ᵇ`
- `_≟_`
- `_≡ᵇ_`
- `ifᵈ_then_else_`
- `⌊_⌋`
- `filterᵐᵇ`
- `filterᵐ?`
- suffix `?` (e.g. `isVKeyAddr?`)

There are two tasks here:
- [x] Figure out an acceptable approach
- [x] Implement said approach

PR #309, opened 2023-11-14
[fix #36, #281] consistent use of decision procedures

### LINKED ISSUE #36: Organize decidability instances
Instances for deciding various properties are scattered all over the files that contain proofs. We should find a good way of dealing with decidability consistently.

omelkonian:
*Depends on #301*

---

- Introduces generalised branching on boolean-like things such as
  Bool/Maybe/Dec in `Interface.ToBool` and replaces all branching
variations with this approach uniformly.

- [Gov] More standard predicates isCC/isDRep/isSPO as inductive datatypes

- [Set/Map] hide decision procedures from the user-facing API

- [Script] Significant refactoring of `Ledger.Script.evalTimelock`:
  + Heal boolean blindness by replacing function (... -> Bool) with the
    proper decision procedure for `evalTimelock`
  + Extra lemma for List.Sublist in `Data.List.Relation.Binary.Sublist.Ext`
  + Factored out m-of-n predicate (+ its dec.proc.) in `Data.List.Relation.Unary.MOf`

- [Prelude] introduce polymorphic unit as ⊤↑/tt↑

### DIFF
--- a/src/Data/List/Relation/Binary/Sublist/Ext.agda
+++ b/src/Data/List/Relation/Binary/Sublist/Ext.agda
@@ -1,0 +1,10 @@
+{-# OPTIONS --safe #-}
+module Data.List.Relation.Binary.Sublist.Ext where
+
+open import Data.List.Relation.Binary.Sublist.Propositional public
+  using (_⊆_; []; _∷_; _∷ʳ_)
+open import Data.List using (List; []; _∷_)
+
+[]⊆ : ∀ {ℓ}{A : Set ℓ} {xs : List A} → [] ⊆ xs
+[]⊆ {xs = []} = []
+[]⊆ {xs = _ ∷ _} = _ ∷ʳ []⊆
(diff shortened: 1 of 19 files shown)


omelkonian:
Since this is based on current PR #301, reviewers can use these changes until it's merged: https://github.com/input-output-hk/formal-ledger-specifications/compare/orestis/no-fork-stdlib-meta...orestis/dec-refactor

WhatisRT (commented):
Thanks, this is great! The only thing I dislike is `_？_∶_`. I've programmed enough C to be fine with that syntax, but anyone who didn't will be super confused.

I assume you still don't like rebinding `if_then_else_` to this? Is there a reason you didn't call it `ifᵈ_then_else_`? We could also use some other variation of that name, which would make this a lot easier to read.

WhatisRT (approved):

WhatisRT:
Can you not use the `mutual` trick that I've used for deriving `DecEq`?

omelkonian:
@WhatisRT I am not sure how the `mutual` trick (i.e. deriving multiple mutually-recursive instances of the same typeclass together / simultaneously) applies to the termination problems of `evalTimelock`'s decision procedure; can you elaborate?

PR #310, opened 2023-11-15
CI: record typechecking times

omelkonian:
This PR introduces a `typeCheck.sh` script that measure the total time of typechecking `Everything.agda` and then records individual module times under `typecheck.time` (by looking at the modification dates of `.agdai` files).

That way, we can check that our changes do not terribly slow down typechecking by looking at the generated file in the `gh-pages` branch (or the CI artifact attached to the current PR).

Moreover, we can run the script through our whole (git) history to discover previously introduced regressions, but that can be done independently offline.

@WhatisRT has mentioned that the [ledger team](https://github.com/input-output-hk/cardano-ledger) actually creates nice graphs from such time durations, which we should definitely investigate to track typechecking performance across history. (@WhatisRT  could you provide more concrete details/links and open a new issue to track this feature?)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,17 +62,18 @@ jobs:
       - name: Build formalLedger
         id: formalLedger
         run: |
-          d=$(nix-build -A formalLedger)
+          mkdir -p outputs
+          d=$(nix-build -A formalLedger -j1 -o outputs/formalLedger)
           dRet=$(echo "$d" | tr '\n' ' ')
           closure=$(nix-store --query --requisites --include-outputs $dRet)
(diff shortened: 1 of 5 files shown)


WhatisRT:
Here's the action the ledger team uses: https://github.com/input-output-hk/cardano-ledger/blob/master/.github/workflows/bench.yml I tried locating the page, but it appears to be broken ATM. But basically, it is this action: https://github.com/benchmark-action/github-action-benchmark

omelkonian:
@WhatisRT I think it's good to go, we now get the artifact in the PR build and should have a `docs/typechecking.time` in the `gh-pages` branch once merged.

WhatisRT (approved):
LGTM

PR #313, opened 2023-11-15
CI: remove Github cache entirely

omelkonian:
Since we are importing IOG's nix cache, there is no longer need for managing our own cache on Github.

@WhatisRT I just deleted both the save and load actions; please check that it does not mess up with the IOG cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,17 +26,6 @@ jobs:
             trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
             substituters = https://cache.iog.io https://cache.nixos.org/
 
-      - name: Restore store
-        id: cache-store-restore
-        uses: actions/cache/restore@v3
-        with:
-          path: |
-              store
-              derivations
-          key: cache-store-${{ runner.os }}-${{github.run_id}}
-          restore-keys: |
-            cache-store-${{ runner.os }}
-
       - name: Import cached nix store
         continue-on-error: true
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -36,17 +36,6 @@ jobs:
...
-        uses: actions/cache/restore@v3
-        with:
-          path: |
-              store
-              derivations
-          key: cache-store-${{ runner.os }}-${{github.run_id}}
-          restore-keys: |
-            cache-store-${{ runner.os }}
-
       - name: Import cached nix store
```
teodanciu, line 39: Shouldn't we also remove this?

omelkonian: I presumed this is what actually utilises the IOG cache, but I might be wrong?

teodanciu:
I think a bit more can be cleaned, like all the noise of adding the names of the derivation to GITHUB_OUTPUT,  for example here: https://github.com/input-output-hk/formal-ledger-specifications/blob/e31c45243b793be67f675eb8937524637dcd7d62/.github/workflows/build.yml#L58

And these calls to nix-store: https://github.com/input-output-hk/formal-ledger-specifications/blob/e31c45243b793be67f675eb8937524637dcd7d62/.github/workflows/build.yml#L56

.github/workflows/build.yml
teodanciu: Maybe I misunderstood the intention.
Do we want to get rid of the mechanism that : exports the store to a local file, uploads it to github cache, so that it another job it can be downloaded and imported   ?

omelkonian: I don't know enough of how this whole thing works, so I cannot really answer.

My question with this PR is: 
> Is there any point in having these cache-save/-restore steps in the CI build?

Or another way to put it:
> If we completely get rid of the cache-export/import mechanism for Github, do we lose anything or do we simply get faster build times?

omelkonian:
@teodanciu indeed the build is still fast (2m) after removing all the Nix-related lines and just leaving the `nix-build` ones.

@WhatisRT please review this as well and confirm that all of these commands were actually redundant, then merge at your own convenience.

.github/workflows/build.yml
teodanciu: The whole import/export was done just for faster build times.  If we get that without it, then we can safely remove it .

WhatisRT:
I want to point out that I already did this in #300, which never got reviews 😅

teodanciu:
Oops, sorry @WhatisRT, I missed that completely!

omelkonian:
Closing in favour of #300 .

PR #315, opened 2023-11-17
Rename `sucᵉ` → `_+ 1`

williamdemeo:
# Description

Addresses one item of #282 

- [X] Give `HasAdd` a more general interface: `_+_ : A → B → A`; this makes it an aggregation as opposed to a sum.
- [X] Rename `HasAdd` to `HasAggregate`.
- [X] Rename some of the `sucᵉ` instances to `_+ 1` (it doesn't work for all of them).

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Interface/HasAggregate.agda
+++ b/src/Interface/HasAggregate.agda
@@ -1,0 +1,9 @@
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Interface.HasAggregate where
+
+record HasAggregate (A : Set){B : Set} : Set where
+  infixl 7 _+_
+  field _+_ : A → B → A
+
+open HasAggregate ⦃ ... ⦄ public
--- a/src/Interface/HasAdd.agda
+++ b/src/Interface/HasAdd.agda
@@ -1,8 +1,0 @@
-{-# OPTIONS --safe --cubical-compatible #-}
-module Interface.HasAdd where
-
-record HasAdd (A : Set) : Set where
-  infixl 7 _+_
-  field _+_ : A → A → A
-
-open HasAdd ⦃ ... ⦄ public
--- a/src/Interface/HasAggregate/Instance.agda
+++ b/src/Interface/HasAggregate/Instance.agda
@@ -1,13 +1,14 @@
 {-# OPTIONS --safe #-}
-module Interface.HasAdd.Instance where
 
-open import Interface.HasAdd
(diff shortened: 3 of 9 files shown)


WhatisRT (approved):
LGTM, but I'd like @omelkonian's feedback as well

src/Interface/HasAggregate.agda
```diff
@@ -0,0 +1,9 @@
+{-# OPTIONS --safe --cubical-compatible #-}
+
+module Interface.HasAggregate where
+
+record HasAggregate (A : Set){B : Set} : Set where
+  infixl 7 _+_
+  field _+_ : A → B → A
```
omelkonian, line 7: This feels quite unintuitive to me and also could easily create ugly type-inference behaviour. 

What about making `Epoch` simply definitionally equal to `ℕ` and thus only requiring the more intuitive `HasAdd` typeclass, which btw I would model as follows so as to reuse Semigroup:
```agda
record HasAdd {a} (A : Set a) where
  field {{ sg }} : Semigroup A
  infixl 7
  _+_ = _<>_
open HasAdd ⦃...⦄ public
```

WhatisRT: One problem about this is that `Epoch` is actually a `Maybe Natural` (I think?) in the Haskell code, since the genesis epoch is special. I personally think that's a bit stupid, but apparently Haskellers like that sort of stuff.

So in that sense, `Epoch` is more of an `ℕ`-module, with this addition being the `ℕ`-action. Depending on your viewpoint, you can also see `Epoch` as being `ℕ ∪ {-1}`. But the standard symbol for an action is the multiplication symbol, which might be more confusing... 🤔

omelkonian:
> LGTM, but I'd like @omelkonian's feedback as well

Still don't like it I am afraid :)

A better solution would be to use literal overloading (where `1` means the epoch number 1), so that we regain symmetry of `+` at least and have no longer need for the un-intuitive `HasAggregate` typeclass.

@WhatisRT seems to agree here in Taiwan, so I am not alone in this.

WhatisRT:
Yeah, I remembered that we already use literal overloading in `Ledger.PPUp` (which uses `Algebra.Literals`, which lets you write numbers to be any element in a semiring). We can also use that for `ℚ`.

WhatisRT:
Closing this, since we're going with a different approach and it'll probably be easier to start from scratch.

Issue #236, opened 2023-10-05
Only allow proposals that have a path to be enacted


WhatisRT:
The hash protection scheme means we can only enact actions that were designed to be applied on top of the current one. We now want to additionally guarantee that all current proposals have a sequence of enactments that will get them enacted. This is what we need to do:
- [x] Add `allEnactable` predicate and its decision procedure.
- [x] Check that new proposals point to an existing proposal or the current state.
- [ ] Remove proposals that become impossible to enact at the epoch boundary.
- [ ] Prove that 'all proposals point at some previous proposal or the current state' is an invariant.

williamdemeo:
@WhatisRT I have completed the first two tasks.  I have questions about the other two.

PR #317, opened 2023-11-20
Add `addEnactable` predicate

WhatisRT:
# Description

This is a requirement for #236 

# Checklist

- [x] Self-reviewed the diff


src/Axiom/Set.agda
```diff
@@ -167,9 +167,18 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-unions : {a : A} {U : Set (Set A)} → (∃[ T ] T ∈ U × a ∈ T) ⇔ a ∈ proj₁ (unions U)
   ∈-unions = proj₂ $ unions _
 
+  fromList-finite : ∀ {l : List A} → finite (fromList l)
+  fromList-finite {l = l} = l , mk⇔ (from ∈-fromList) (to ∈-fromList)
+
   ∅ : Set A
   ∅ = fromList []
 
+  ¬a∈[] : ∀ {a : A} → a ∈ˡ [] → ⊥
```
williamdemeo, line 176: These are just useful to have around (so I don't have to figure out how to prove them again next time).

src/Axiom/Set.agda
```diff
@@ -167,9 +167,18 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-unions : {a : A} {U : Set (Set A)} → (∃[ T ] T ∈ U × a ∈ T) ⇔ a ∈ proj₁ (unions U)
   ∈-unions = proj₂ $ unions _
 
+  fromList-finite : ∀ {l : List A} → finite (fromList l)
```
williamdemeo, line 170: This is an easy proof (that a set created from a list is finite), but I found it helpful at one point.  After moving to a more list-based approach in this PR, it's no longer needed.  I wonder if I should delete it or leave it for future reference/use.

williamdemeo:
Note: this PR was type-checking fine until I merged the latest master (with the newly merged "consistent use of decision procedures" PR).  Now type-checking hangs on `Ledger.Utxo.Properties` again. 😞

omelkonian:
> Note: this PR was type-checking fine until I merged the latest master (with the newly merged "consistent use of decision procedures" PR). Now type-checking hangs on `Ledger.Utxo.Properties` again. 😞

The typechecking durations recorded on `gh-pages` seem unaffected: https://github.com/input-output-hk/formal-ledger-specifications/blob/gh-pages/docs/typecheck.time. 

It might be the case that the code *does not currently* typechecks, but instead has unsolved metas that slow things down a lot, i.e. the code after merging does not typecheck but Agda is too slow to tell you that.

WhatisRT:
The hydra errors look like something is broken there - they contain nix error messages. But the github action succeeds so it should be fine.

src/Ledger/Gov.lagda
```diff
@@ -29,6 +33,41 @@ record GovEnv : Set where
...
+
+connects? : ∀ l aidPair → Dec (connects l aidPair)
+connects? [] (aid , aidₚ) = aid ≟ aidₚ
+connects? ((a , aₚ) ∷ s) (aid , aidₚ) with connects? s (a , aid) | (aₚ ≟ aidₚ)
+...| yes p  | yes q  = yes (p , q)
+...| _      | no ¬q  = no λ (_ , q) → ¬q q
+...| no ¬p  | _      = no λ (p , _) → ¬p p
+
+enactable : EnactState → GovActionID × GovActionState → List (GovActionID × GovActionID) → Set
+enactable eState (aid , record { action = action }) aidPairs =
```
WhatisRT, line 56: There are two issues with this definition:
- I made a mistake in the original definition, we need to make sure that (the ID inside) `getHash (GovActionState.prevAction as)` connects to `aidₚ`, instead of `aid`. `enactable` should not depend on `aid` at all.
- Using `subpermutations` makes the definition of `enactable?` easier, but it makes any other proofs about it way harder, since they now need to unfold the definition of `subpermutations`. Also, this definition will show up in the PDF, so the reader would need to know the definition of `subpermutations`. It would be much better to revert this to 'there exists a subset such that ...'.

williamdemeo: I'm okay with the first point, and all the comments above.  I'll fix those.

Re. the second point, [we did discuss](https://input-output-rnd.slack.com/archives/D04JE60R5T3/p1701949721662799) using the collection of all permutations of all sublists.  If I revert to a set-based approach, the proofs will be much harder, and we will still need a function for getting all subsets of a given set and testing equality of each of these sets with each candidate.  Later proofs will still need to unfold the definitions of these functions.

I would like to fix the issues mentioned in the first point and above comments and then I would suggest we get this list-based solution merged, even if it's not the ideal solution.  It works, and, from my experience over the past week or so, a set-based solution will be much harder to get working and to reason about.

williamdemeo: Okay, I'll spend a few more hours on the set-based proof and see if I can get it working.  If I get stuck, I'll ping you (@WhatisRT) on slack.  Thanks!

williamdemeo:
@WhatisRT actually, I don't understand the first bullet point.  Here is your original version:

```agda
enactable : EnactState → GovActionID × GovActionState → ℙ (GovActionID × GovActionID) → Set
enactable e (aid , as) s = case getHashES e (GovActionState.action as) of λ where
  nothing → ⊤
  (just aid') → ∃[ t ] (fromList t ⊆ s × connects t aid' aid)
```

Here's my latest version:

```agda
enactable : EnactState → GovActionID × GovActionState → List (GovActionID × GovActionID) → Set
enactable eState (aid , record { action = action }) aidPairs =
  case getHashES eState action of λ where
  nothing      → ⊤
  (just aidₚ)  → Any (λ l → connects l (aid , aidₚ)) (subpermutations aidPairs)
```

In the first version the id returned by `getHashES` is `aid'` and in the second version it's `aidₚ`.

Are you saying that `aidₚ` should connect to itself?

src/Data/List/Ext/Subperm.agda
```diff
@@ -0,0 +1,37 @@
...
+
+  open import Relation.Binary.Core using (REL)
+
+  data Subperm : REL (List A) (List A) 0ℓ where
+    []   : Subperm [] []
+    _∷ₚʳ_ : ∀ {xs ys} → ∀ y → Subperm xs ys → Subperm xs (y ∷ ys)
+    _∷ₚ_  : ∀ {x xs y ys} → x ∈ (y ∷ ys) → Subperm xs (y ∷ ys) → Subperm (x ∷ xs) (y ∷ ys)
+
+  Subperm⁻¹ : {x : A}{xs l : List A} → Subperm (x ∷ xs) l → x ∈ l × Subperm xs l
+  Subperm⁻¹ (y ∷ₚʳ Spxxs) = (there (proj₁ (Subperm⁻¹ Spxxs))) , (y ∷ₚʳ proj₂ (Subperm⁻¹ Spxxs))
```
omelkonian, line 22: Instead of repeating the recursive call, I would share via `let`:
```agda
Subperm⁻¹ (y ∷ₚʳ Spxxs) = let x∈, p = Subperm⁻¹ Spxxs in there x∈ , (y ∷ₚʳ p)
```
(unfortunately, Agda will later inline this let and lose this sharing when actually computing with this, but you would at least have written the code in the appropriate way and the fault now lies entirely with the implementation of Agda :P)

src/Data/List/Ext/Subperm.agda
```diff
@@ -0,0 +1,37 @@
...
+  Subperm⁻¹ (x∈yys ∷ₚ Spxsyys) = x∈yys , Spxsyys
+
+  Subperm[] : {l : List A} → Subperm [] l
+  Subperm[] {[]} = []
+  Subperm[] {x ∷ _} = x ∷ₚʳ Subperm[]
```
omelkonian, line 27: 1. Note that the first argument of `_∷ₚʳ_` (i.e. the element being "dropped" from the sub-permutation) can (almost) always be inferred from the second type-level index of the `Subperm` relation (i.e. the big list). So you will (almost) never need to explicitly give it in terms, which suggests a "cons" infix notation might not be the best option here.

2. Another confusing aspect of using two "cons" operators for the two (recursive) constructors is that the element being prepended is different in each case: for `_∷ₚʳ_` we are "dropping" a single element; for `_∷ₚ_` we are "keeping" a proof of membership (of a single element).

3. Constructor overloading is one of the biggest conveniences we have in Agda, so I would not be afraid to use the more idiomatic (in the context of stdlib) `keep` and `drop` for the (recursive) constructors instead. Coupled with points (1) and (2), here's the definition I would write to express the same thing:
```agda
  data Subperm : Rel (List A) 0ℓ where
    []   : Subperm [] []
    drop : ∀ {xs ys y} → Subperm xs ys → Subperm xs (y ∷ ys)
    keep : ∀ {x xs ys} → x ∈ ys → Subperm xs ys → Subperm (x ∷ xs) ys
```
The resulting definitions you get become simpler and more intuitive IMO:
```agda
  Subperm⁻¹ : {x : A}{xs l : List A} → Subperm (x ∷ xs) l → x ∈ l × Subperm xs l
  Subperm⁻¹ (drop p)    = bimap there drop (Subperm⁻¹ p)
  Subperm⁻¹ (keep x∈ p) = x∈ , p

  Subperm[] : {l : List A} → Subperm [] l
  Subperm[] {[]} = []
  Subperm[] {_ ∷ _} = drop Subperm[]

  subperm? : ⦃ DecEq A ⦄ → Decidable (Subperm)
  subperm? [] _ = yes Subperm[]
  subperm? (_ ∷ _) [] = no λ where (keep () _)
  subperm? (x ∷ xs) (y ∷ ys)
    with ¿ x ∈ (y ∷ ys) ¿
  ... | no  x∉ = no (x∉ ∘ proj₁ ∘ Subperm⁻¹)
  ... | yes x∈
    with subperm? xs (y ∷ ys)
  ... | no ¬p = no (¬p ∘ proj₂ ∘ Subperm⁻¹)
  ... | yes p = yes (keep x∈ p)
```

omelkonian (changes requested):
Some interim comments that I found important to address in this intermediate stage.

src/Data/List/Ext/Subperm.agda
```diff
@@ -0,0 +1,37 @@
...
+open import Data.List
+open import Data.List.Relation.Unary.Any using (here; there; any?)
+open import Relation.Binary.Definitions using (Decidable)
+open import Data.List.Membership.Propositional using (_∈_)
+
+module _ {A : Set} where
+
+  open import Relation.Binary.Core using (REL)
+
+  data Subperm : REL (List A) (List A) 0ℓ where
```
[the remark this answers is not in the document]
williamdemeo: Ah, yes, I was following the example in the Standard Library too closely... you're right, here we can use `Rel`.  (In the standard library, the type in question is `REL (List A) (List B)`, so they need `REL`.)

src/Data/List/Ext/Properties.agda
```diff
@@ -59,6 +62,96 @@ AllPairs⇒≡∨R∨Rᵒᵖ : ∀ {ℓ ℓ'} {A : Set ℓ} {R : A → A → Set
...
+
+
+
+
+
+
+
+
+
+-- TESTS --
```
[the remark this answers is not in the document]
williamdemeo: @omelkonian do you mean, "uncomment the tests, do the type checking, and then comment out or delete the tests again"? ...or do you suggest I uncomment the tests and leave them in the code (uncommented) in my next commit?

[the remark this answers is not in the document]
omelkonian: The latter: leave the unit tests uncommented (or a select significant subset of the unit tests).

src/Data/List/Ext.agda
```diff
@@ -1,12 +1,66 @@
 {-# OPTIONS --safe #-}
 module Data.List.Ext where
 
-open import Agda.Builtin.List using (List; []; _∷_)
-open import Agda.Builtin.Maybe using (Maybe; nothing; just)
-open import Agda.Builtin.Nat using (Nat; zero; suc)
```
williamdemeo, line 6: Not sure why we're using Builtins here.  It's generally preferred to use standard library definitions where possible.

src/Ledger/Gov.lagda
```diff
@@ -41,6 +51,57 @@ _⊢_⇀⦇_,GOV⦈_     : GovEnv → GovState → List (GovVote ⊎ GovProposal
...
+enactable e aidPairs = λ (aidNew , as) → case getHashES e (GovActionState.action as) of λ where
+  nothing → ⊤
+  (just aidOld) → ∃[ t ](fromList t ⊆ fromList aidPairs × t connects aidNew to aidOld)
+
+open Equivalence
+
+∃?-connecting-subset : ∀ L {u}{v} → Dec (∃[ l ](l ⊆ˡ L) × (l connects u to v))
+∃?-connecting-subset L {u}{v} with [ L connects u to v ?]
+... | yes p = yes (L , (⊆ˡ-id , p))
+... | no ¬q = {!!}
```
williamdemeo, line 91: This is the last (and hardest) hole that's left to fill.  I believe a recursive algorithm would work---one that looks at each subpermutation of L of length |L|-1 to find one that connects u to v.

WhatisRT: The whole point of adding the proofs about subpermutations is to make this easy, since `∃[ l ](l ⊆ˡ L) × (l connects u to v)` is equivalent to `Any (_connects u to v) (subpermutations L)` and decidability of that is just `any? [_connects u to v ?] (subpermutations L)`.

williamdemeo: Yes, I know that, but I wasn't able to prove the equivalence `[ l ](l ⊆ˡ L) × (l connects u to v) ⇔ Any (_connects u to v) (subpermutations L)` so I was trying to find an alternative route... but I'll give the equivalence proof another try today.

williamdemeo: For example, one of the issues is that `[] ⊆ˡ []` but `[] ∉ subpermutations []` (since `subpermutations []` is `[]` and `[] ∉ []`).  Of course, `[]` will not be a "connecting list" (unless the two action ids in question are equal), so I should be able to find a way around this issue... still working on it.

williamdemeo: I finally managed to prove the equivalence `l ⊆ˡ L ⇔ Subperm l L` for the inductive `Subperm` type I had defined earlier, so now I just need to get the "Any" function working for that type and I think we're in business!

WhatisRT: I thought the whole point of `subpermutations` was that it'd be a function that returns all the sublists. If it doesn't do that then you should probably fix that, otherwise `subpermutations` will just be completely unused. And note that the `subpermutations` function is the one doing all the work, so replacing it means you have to write another function constructing subpermutations.

Also, this reveals that you've had a bug in your changed definition: if `subpermutations [] = []`, then if the list of proposals is empty, `enactable` is always the empty type!

williamdemeo: Yes, I agree with all of that.  Actually, `subpermutations` is fine.  I've tested it.  It gives the expected results on small lists, empty list, etc.  However (as you observed early on) it's hard to prove that it is indeed the list of all sublists, and permutations thereof, of the given list.

I agree that if I could prove that `(l ⊆ˡ L) ⇔ l ∈ subpermutations L`, then I could use `any?` on the list `subpermutations aidPairs` to complete the decidability proof (`∃?-connecting-subset`)... and I'll be done!

Thanks for your comments and helping me clarify what I need to do. 👍🏼

williamdemeo: Update: I think proving `Subperm l L ⇔ l ∈ subpermutations L` may be easier.  I just pushed an outline of the proof to `Data.List.Ext.Subperm.Properties`.  I'll try to complete the proof today.

WhatisRT: I'm really confused why you're still going ahead with this. The theorem you eventually need to prove is
```
l ⊆ L ⇔ l ∈ subpermutations L
```
but your `subpermutations` function still doesn't have that property. You seem to be proving that equivalence in the case that `l` isn't the empty list, but why? The whole point of the `subpermutations` function is to be a function that gives you a list having exactly this property. Just change `subpermutations` to have the property you need!

Most likely this will also make a bunch of proofs substantially easier. You can even write it in a correct-by-construction manner, where you first define `subpermutationsᶜ : (L : List A) → List (∃[ l ] Subperm l L)` and then `subpermutations` just projects out the lists. Then the first half of the equivalence is free.

williamdemeo: I don't entirely agree with all of the points above.  In particular, the property `l ⊆ L ⇔ l ∈ subpermutations L` will not (and should not) hold in general.  It will hold for lists without repeated entries... and I am indeed trying to prove the equivalence under those assumptions. I hope to complete the proof today.  I am very close.

src/Data/List/Ext.agda
```diff
@@ -1,12 +1,51 @@
...
 
 -- Looking up an index into the list; fails when out-of-bounds.
-_⁉_ : ∀ {a}{A : Set a} → List A → Nat → Maybe A
+_⁉_ : ∀ {a}{A : Set a} → List A → ℕ → Maybe A
 []       ⁉ _     = nothing
 (x ∷ _)  ⁉ zero  = just x
 (_ ∷ xs) ⁉ suc n = xs ⁉ n
+
+-- LIST OPERATIONS --
+flatMap : (A → List B) → List A → List B
```
WhatisRT, line 25: This is called `concatMap` in the stdlib

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +36,173 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+  All++intro {ll = .(_ ∷ _)} (px All.∷ h1) h2 = px All.∷ (All++intro h1 h2)
+
+  ∈All-def : {ls : List (List A)}{y : A} → All (y ∈_ ) ls → (∀ l → l ∈ ls → y ∈ l)
+  ∈All-def {.(l ∷ _)} (y∈l All.∷ _) l (here refl) = y∈l
+  ∈All-def {.(_ ∷ _)} (_ All.∷ y∈all) l (there l∈ls) = ∈All-def y∈all l l∈ls
+
+-------------------------------
+------ properties of Any ------
+-------------------------------
+module _ {a}{A : Set a} where
```
WhatisRT, line 125: Everything in here should be in `Data.List.Relation.Unary.Any.Properties`

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +36,173 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+
+  map∷distr++ʳ : {ll lr : List (List A)}{x : A} → ∀ l → l ∈ map(x ∷_) lr → l ∈ map(x ∷_) (ll ++ lr)
+  map∷distr++ʳ {[]} _ = id
+  map∷distr++ʳ {_ ∷ _} l l∈ = there (map∷distr++ʳ l l∈)
+
+-------------------------------
+------ properties of All ------
+-------------------------------
+module _ {a}{A : Set a} where
+  -- TODO: check whether any of these already exist in the Agda std lib
```
WhatisRT, line 87: Most of this is in `Data.List.Relation.Unary.All.Properties`

src/Data/List/Ext.agda
```diff
@@ -1,12 +1,51 @@
...
+flatMap f [] = []
+flatMap f (x ∷ xs) = f x ++ flatMap f xs
+
+-- every element in the first list occurs in the second list
+_⊆_ : (l l' : List A) → Set _
```
WhatisRT, line 30: This is defined in `Data.List.Relation.Binary.Subset.Propositional` with a different but equivalent definition. If you use that, you may be able to use some theorems about it rather than proving them yourself.

src/Data/List/Ext.agda
```diff
@@ -1,12 +1,51 @@
...
+l ⊆ l' = All (_∈ l') l
+
+-- list of all sublists of the given list
+sublists : List A → List (List A)
```
WhatisRT, line 34: This function is unused (except for the properties that are proven about it)

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +231,440 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
-                 → AllPairs R l → a ∈ˡ l → b ∈ˡ l → a ≡ b ⊎ R a b ⊎ R b a
-AllPairs⇒≡∨R∨Rᵒᵖ [] = λ ()
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (here refl) (here refl) = inj₁ refl
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (here refl) (there b∈l) = inj₂ (inj₁ (All.lookup x b∈l))
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (there a∈l) (here refl) = inj₂ (inj₂ (All.lookup x a∈l))
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (there a∈l) (there b∈l) = AllPairs⇒≡∨R∨Rᵒᵖ h a∈l b∈l
+-----------------------------------------------------------
+-------- properties of concatenation and union ------------
+-----------------------------------------------------------
+module _ {a} {A : Set a} where
```
WhatisRT, line 237: Most things in here are in `Data.List.Relation.Unary.Any.Properties` (note that `_∈_` is defined via `Any`)

WhatisRT (commented):
There's lots of duplication with things that already exist in the stdlib. Also, I see lots of parentheses that are stacked right next to each other without spacing, which is inconsistent with the rest of our code base.

I still need to review a bunch of stuff, this is just what I saw first. Also, please rebase on master.

williamdemeo:
> There's lots of duplication with things that already exist in the stdlib. 

Okay, I'm working on a major revision to eliminate the duplication.

> Also, I see lots of parentheses that are stacked right next to each other without spacing, which is inconsistent with the rest of our code base.

Oh, are we spacing out parentheses?  I hadn't noticed.  So, `) )` instead of `))`?

> I still need to review a bunch of stuff, this is just what I saw first. Also, please rebase on master.

Okay, I've reabased and I'm now doing a major rewrite/overhaul to eliminate duplicates and use the standard library's list inclusion relation (instead of `All`).  I'll let you know when it's done and ready for review.

williamdemeo:
@WhatisRT I've finished the rewrite and cleanup.  Some of the proofs could be simplified with some more time/effort.  I'll continue polishing, but otherwise it's probably good to go... it's complete and correct at least! 😅

Todo: function to find largest enactable subset. (I'll start on that as well.)

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +161,382 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+  ...| inj₂ v with insert-decomp∈ xs v
+  ...| l , l∈ls , xs∈yl = l , there l∈ls , xs∈yl
+
+  insert-decomp⁻¹ : {ls : List (List A)}{y : A} → ∀ xs → ∃[ l ] (l ∈ ls × xs ∈ (insert y l))
+                         → xs ∈ (concatMap (insert y) ls)
+  insert-decomp⁻¹ {l ∷ ls} {y} xs (.l , here refl , xs∈) = xs⊆xs++ys _ _ xs∈
+  insert-decomp⁻¹ {l ∷ ls} {y} xs (l' , there l'∈ , xs∈) = xs⊆ys++xs _ _ (insert-decomp⁻¹ xs (l' , (l'∈ , xs∈)))
+
+  insert++distr : {ll lr : List (List A)}{xs : List A}{x : A} → xs ∈ concatMap (insert x) (ll ++ lr)
+                     → xs ∈ concatMap (insert x) ll ++ concatMap (insert x) lr
```
WhatisRT, line 246: This proof and the ones below get way easier if you have
```
concatMap-++ : {f : A → List B} {l l' : List A} → concatMap f (l ++ l') ≡ concatMap f l ++ concatMap f l'
```

omelkonian: https://omelkonian.github.io/formal-prelude/Prelude.Lists.Concat.html#701

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +161,382 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+  l∈map∷→l∈fm∘insert {(_ ∷ _) ∷ _} l (there l∈) = xs⊆ys++xs _ _ (l∈map∷→l∈fm∘insert l l∈)
+
+  l∈map³→l∈fm∘insert∘map² : {ls : List (List A)}{y' y z : A} → ∀ l
+                            → l ∈ map (y' ∷_) (map (y ∷_) (map (z ∷_) ls))
+                            → l ∈ concatMap (insert y) (map (y' ∷_) (map (z ∷_) ls))
+  l∈map³→l∈fm∘insert∘map² {_ ∷ _} l (here px) = there (here px)
+  l∈map³→l∈fm∘insert∘map² {_ ∷ _} l (there l∈) = xs⊆ys++xs _ _ (l∈map³→l∈fm∘insert∘map² l l∈)
+
+  insert-decomp : {ls : List (List A)}{x : A} → ∀ ys → ys ∈ (concatMap (insert x) ls)
+                  → ∃[ l ] (l ∈ ls × ys ⊆ (x ∷ l))
```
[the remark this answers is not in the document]
omelkonian: https://omelkonian.github.io/formal-prelude/Prelude.Lists.Concat.html#1632

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +39,100 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+  -- TODO: check whether any of these already exist in the Agda std lib
+  x∈All→∈Allmapy∷ : {ls : List (List A)}{y x : A} → All (x ∈_) ls → All (x ∈_) (map (y ∷_) ls)
+  x∈All→∈Allmapy∷ {[]} x∈all = x∈all
+  x∈All→∈Allmapy∷ {_ ∷ _} (px All.∷ x∈all) = (there px) All.∷ (x∈All→∈Allmapy∷ x∈all)
+
+  x∈Allmapx∷ : {ls : List (List A)}{x : A} → All (x ∈_) (map (x ∷_) ls)
+  x∈Allmapx∷ {[]} = All.[]
+  x∈Allmapx∷ {_ ∷ _} = here refl All.∷ x∈Allmapx∷
+
+  ∈×All∈→∈ : {ys xs : List A}{z : A} → z ∈ ys → All(_∈ xs) ys → z ∈ xs
```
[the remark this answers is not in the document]
williamdemeo: (removed)

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +39,100 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+  ∈×All∈→∈ (here refl) (px All.∷ ys∈) = px
+  ∈×All∈→∈ (there z∈) (px All.∷ ys∈) = ∈×All∈→∈ z∈ ys∈
+
+  ∈All-def : {ls : List (List A)}{y : A} → All (y ∈_ ) ls → (∀ l → l ∈ ls → y ∈ l)
+  ∈All-def (y∈l All.∷ _) l (here refl) = y∈l
+  ∈All-def (_ All.∷ y∈all) l (there l∈ls) = ∈All-def y∈all l l∈ls
```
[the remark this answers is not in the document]
williamdemeo: Actually, the definition of this helper function changed (when I changed everything to use the library's definition of set inclusion).  It's now:

```agda
  ∈All-def : {ls : List (List A)}{y : A} → All (y ∈_ ) ls → (∀ l → l ∈ ls → y ∈ l)
  ∈All-def (y∈l All.∷ _) l (here refl) = y∈l
  ∈All-def (_ All.∷ y∈all) l (there l∈ls) = ∈All-def y∈all l l∈ls
```

and it helped me transition from the old definition of set inclusion to the library's, but I'll see if I can remove it without too much extra work.

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +39,100 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+  ∃→Any {l' ∷ ls} (.l' , here refl , Pl) = here Pl
+  ∃→Any {l' ∷ ls} (l , there l∈ , Pl) = there (∃→Any (l , l∈ , Pl))
+
+  Any→∃ : {ls : List (List A)} → Any P ls → ∃[ l ](l ∈ ls × P l)
+  Any→∃ {.(l ∷ _)} (here {l} px) = l , here refl , px
+  Any→∃ {.(l ∷ ls)} (there {l} {ls} h) with Any→∃ h
+  ...| l' , l'∈ls , Pl' = l' , there l'∈ls , Pl'
+
+  ∃⇔Any : {ls : List (List A)} → (∃[ l ](l ∈ ls × P l)) ⇔ Any P ls
+  ∃⇔Any = mk⇔ ∃→Any Any→∃
```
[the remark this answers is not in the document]
williamdemeo: I removed the three lemmas relating Any to exists and used your suggestion `R.↔⇒ Any↔` directly where it is needed in `Ledger.Gov`.

src/Data/List/Ext/Properties.agda
williamdemeo: Yes, good points.  I'll try to use them to simplify things.

By the way, as I realized just yesterday, I could probably clean up a bunch of stuff by using the following minor embellishment of the library's `concat⁺`:

```agda
  concatMap⁺ : {ys xs : List A}{f : A → List A} → xs ⊆ ys → concatMap f xs ⊆ concatMap f ys
  concatMap⁺ = concat⁺ ∘ (map⁺ _)
```

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +161,382 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+                  → l ∈ concatMap (insert y) (insert x ys)
+  insert²-swap {[]}  l (here px) = there (here px)
+  insert²-swap {[]}  l (there (here px)) = here px
+  insert²-swap {_ ∷ _} l (here px) = there (here px)
+  insert²-swap {_ ∷ _} l (there (here px)) = here px
+  insert²-swap {_ ∷ _} l (there (there l∈)) with ++→⊎ l∈
+  ...| inj₁ v = xs⊆ys++xs _ _ (l∈map∷→l∈fm∘insert l v)
+  ...| inj₂ v = insert-map-swap l v
+
+  insert²-swap' : {ls : List (List A)}{y x : A} → ∀ l → l ∈ concatMap (insert x) (concatMap (insert y) ls)
```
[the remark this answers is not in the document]
williamdemeo: Yes, I can simplify things a lot now... but please keep in mind that it is only after working so hard on it---trying to prove it from many different directions, including many failed attempts---that what we really need becomes "obvious."  I will not be surprised if, now that the proof is complete, the end result may look very easy after it is cleaned up and reduced to its essence. 😅

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +161,382 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+  ⊎→++ : {ll lr : List A}{x : A} → x ∈ ll ⊎ x ∈ lr → x ∈ ll ++ lr
+  ⊎→++ {ll} {lr} (inj₁ x∈ll) = xs⊆xs++ys ll lr x∈ll
+  ⊎→++ {ll} {lr} (inj₂ x∈lr) = xs⊆ys++xs lr ll x∈lr
+
+  ++→⊎ : {ll lr : List A}{x : A} → x ∈ ll ++ lr → x ∈ ll ⊎ x ∈ lr
+  ++→⊎ {[]} x∈ll++lr = inj₂ x∈ll++lr
+  ++→⊎ {_ ∷ _} (here refl) = inj₁ (here refl)
+  ++→⊎ {_ ∷ _} (there x∈ll++lr) with (++→⊎ x∈ll++lr)
+  ...| inj₁ y∈ys = inj₁ (there y∈ys)
+  ...| inj₂ x∈ll = inj₂ x∈ll
```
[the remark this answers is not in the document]
williamdemeo: (removed the two lemmas relating `++` and `⊎`)

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +161,382 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
-AllPairs⇒≡∨R∨Rᵒᵖ [] = λ ()
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (here refl) (here refl) = inj₁ refl
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (here refl) (there b∈l) = inj₂ (inj₁ (All.lookup x b∈l))
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (there a∈l) (here refl) = inj₂ (inj₂ (All.lookup x a∈l))
-AllPairs⇒≡∨R∨Rᵒᵖ (x ∷ h) (there a∈l) (there b∈l) = AllPairs⇒≡∨R∨Rᵒᵖ h a∈l b∈l
+-----------------------------------------------------------
+-------- properties of concatenation and union ------------
+-----------------------------------------------------------
+module _ {a} {A : Set a} where
+  ⊎→++ : {ll lr : List A}{x : A} → x ∈ ll ⊎ x ∈ lr → x ∈ ll ++ lr
```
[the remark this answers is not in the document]
williamdemeo: (removed)

src/Data/List/Ext/Properties.agda
williamdemeo: Indeed, the following should help simplify things:

```agda
  concatMap⁺ : {ys xs : List A}{f : A → List A} → xs ⊆ ys → concatMap f xs ⊆ concatMap f ys
  concatMap⁺ = concat⁺ ∘ (map⁺ _)

  concatMap-++ : ∀{b}{B : Set b} → ∀ (f : A → List B) (xs ys : List A)
    → concatMap f (xs ++ ys) ≡ concatMap f xs ++ concatMap f ys
  concatMap-++ f xs ys = begin
    concatMap f (xs ++ ys)           ≡⟨⟩
    concat (map f (xs ++ ys))        ≡⟨ cong concat $ map-++ f xs ys ⟩
    concat (map f xs ++ map f ys)    ≡˘⟨ concat-++ (map f xs) (map f ys) ⟩
    concatMap f xs ++ concatMap f ys ∎ where open ≡-Reasoning
```

The former is the one I mentioned above.  The latter is from the link @omelkonian shared.  Thanks Orestis!

src/Data/List/Ext/Properties.agda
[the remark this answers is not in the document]
WhatisRT: I'm confused, that's exactly the definition when I made this comment

src/Data/List/Ext/Properties.agda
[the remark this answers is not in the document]
williamdemeo: > Yes, that special case you've mentioned isn't true, 

I just quoted the result that you suggested.  From your reply it seems you had quite a different result in mind.

> ```agda
> insert→ : {l l' : List A} {x : A} → l ∈ insert x l' → ∃[ l₁ ] ∃[ l₂ ] (l ≡ l₁ ++ [ x ] ++ l₂ × l' ≡ l₁ ++ l₂)
> insert→ {l' = []} (here px) = [] , [] , px , refl
> insert→ {l' = x ∷ l'} (here px) = [] , x ∷ l' , px , refl
> insert→ {l = []} {l' = x ∷ l'} (there h) = ⊥-elim {!!}
> insert→ {l = y ∷ l} {l' = x ∷ l'} (there {x₁} h) = case insert→ (proj₁ $ helper h) , proj₂ (helper h) of λ where
>     ((l₁ , l₂ , pf₁ , pf₂) , refl) → x ∷ l₁ , l₂ , cong (y ∷_) pf₁ , cong (y ∷_) pf₂
>   where
>     helper : {l : List (List A)} {x y : A} {xs : List A} → x ∷ xs ∈ map (y ∷_) l → xs ∈ l × x ≡ y
>     helper = {!!}
> ```

Okay, that's better.  I do already have such decomposition lemmas, for `insert` and `++` separately... but I suppose it's possible that combining them in a single decomposition may yield a cleaner proof...  Well, I'll consider using it while I continue refining and cleaning up the whole PR.

src/Data/List/Ext/Properties.agda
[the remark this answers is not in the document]
williamdemeo: Oh, okay, my mistake.

src/Data/List/Ext/Properties.agda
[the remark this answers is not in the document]
williamdemeo: Btw, your "helper" is almost exactly one of my decomposition lemmas (`map∷decomp∈`) 😄

williamdemeo:
@WhatisRT I think it's looking much better overall now.  Lmkwyt.  In the meantime, I'll work on finding the maximal subset of enactable actions.

src/Ledger/Gov.lagda
```diff
@@ -39,6 +54,64 @@ _⊢_⇀⦇_,GOV⦈_     : GovEnv → GovState → List (GovVote ⊎ GovProposal
...
+
+allEnactable : EnactState → GovState → Set
+allEnactable e aid×states = All (λ p → enactable e (getAidPairsList aid×states) p) aid×states
+
+allEnactable? : (eState : EnactState)(aid×states : GovState) → Dec (allEnactable eState aid×states)
+allEnactable? eState aid×states =
+  all? (λ aid×st → enactable? eState (getAidPairsList aid×states) aid×st) aid×states
+
+maxAllEnactable : EnactState → List (GovActionID × GovActionState) → List (List (GovActionID × GovActionState))
+maxAllEnactable e = maxsublists⊧P{P = λ l → allEnactable e l}{λ l → allEnactable? e l}
```
williamdemeo, line 113: @WhatisRT this function is meant to return all sublists of the given list of proposals that (a) are all enactable and (b) are of maximal length among those sublists that satisfy `allEnactable`.

Next I'd like to do the following:

1. check whether you think this or something similar the right approach toward having a function that removes non-enactable proposals
2. assuming a positive answer to 1, test `maxAllEnactable` with some toy examples of lists of proposals; if you have suggestions or guidance for this step, please advise, otherwise, I'll try to make something up.

WhatisRT: Looks fine. Can you prove that the chosen list satisfies the property & no larger intermediate list does?

williamdemeo: @WhatisRT My latest commit contains proofs verifying the assertions made in the comments of `maxAllEnactable`---namely, `maxAllEnactable` returns a list `ls` of sublists of the given `aid×states` such that
1. each sublist `l ∈ ls` satisfies `allEnactable e l` and
2. each sublist `l ∈ ls` is of maximal length among sublists of `aid×states` satisfying `allEnactable`.

Your comment above refers to "the chosen list."  In fact, I didn't defined it so that there is one particular "special" maximal list (in case there is more than one), but if we do want to chose one, then we could simply take, say, the head of the list of maximal-length, `allEnactable` lists returned by `maxAllEnactable`.

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +44,183 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+
+-- ≤-total : Total _≤_
+-- ≤-total zero    _       = inj₁ z≤n
+-- ≤-total _       zero    = inj₂ z≤n
+-- ≤-total (suc m) (suc n) = Sum.map s≤s s≤s (≤-total m n)
+
+-- Total : Rel A ℓ → Set _
+-- Total _∼_ = Connex _∼_ _∼_
+
+-- Connex : REL A B ℓ₁ → REL B A ℓ₂ → Set _
```
WhatisRT, line 121: I don't know why this is here

src/Data/List/Ext/Properties.agda
```diff
@@ -29,6 +44,183 @@ _⊎-cong_ : ∀ {a b c d} {A : Set a} {B : Set b} {C : Set c} {D : Set d} {k} 
...
+    map++distr (_ ∷ _) (here refl) = here refl
+    map++distr (_ ∷ lls) (there h) = there (map++distr lls h)
+
+    map++distrˡ : (ll : List (List A)) {lr : List (List A)} → map f ll ⊆ map f (ll ++ lr)
+    map++distrˡ (_ ∷ _) (here px) = here px
+    map++distrˡ (_ ∷ lls) (there h) = there (map++distrˡ lls h)
+
+    map++distrʳ : (ll : List (List A)) {lr : List (List A)} → map f lr ⊆ map f (ll ++ lr)
+    map++distrʳ [] h = h
+    map++distrʳ (_ ∷ lls) h = there (map++distrʳ lls h)
```
WhatisRT, line 181: All of this is unused

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +249,324 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+                    → l ∈ concatMap (insert x) (ll ++ lr)
+  insert++distr⁻¹ {[]} l l∈++ = l∈++
+  insert++distr⁻¹ {ll ∷ lls} {lr} l l∈++ with ∈-++⁻ (insert _ ll ++ concatMap (insert _) lls) l∈++
+  ...| inj₂ v = xs⊆ys++xs _ _ (insert++ʳ{ll = lls} v)
+  ...| inj₁ v =
+    case (∈-++⁻ _ (subst (λ u → l ∈ u)
+                    (++-assoc (insert _ ll) (concatMap (insert _) lls) (concatMap (insert _) lr))
+                    (xs⊆xs++ys _ _ v))) of λ where
+      (inj₁ w) → xs⊆xs++ys (insert _ ll) _ w
+      (inj₂ w) → xs⊆ys++xs _ _ (insert++distr⁻¹{ll = lls} l w)
```
WhatisRT, line 375: All of this is completely unnecessary, just use `concatMap-++`

src/Data/List/Ext/Properties.agda
```diff
@@ -54,11 +249,324 @@ module _ {a} {A : Set a} ⦃ _ : DecEq A ⦄ where
...
+
+  ∈All→∈Allinserts : {ls : List (List A)} {y x : A} → All (x ∈_) ls → All (x ∈_) (concatMap (insert y) ls)
+  ∈All→∈Allinserts {.[]} All.[] = All.[]
+  ∈All→∈Allinserts {.(_ ∷ _)} (px All.∷ p) = All++intro (∈→∈Allinsert px) (∈All→∈Allinserts p)
+
+  Allinsert⊆∷ : {xs : List A} {x : A} → All(_⊆ (x ∷ xs)) (insert x xs)
+  Allinsert⊆∷ {[]} = id All.∷ All.[]
+  Allinsert⊆∷ {_ ∷ xs} = id All.∷ (All⊆-swap xs (All⊆→Allmap∷⊆∷ Allinsert⊆∷))
+
+  insert-decomp : {ls : List (List A)} {x : A} → ∀ ys → ys ∈ (concatMap (insert x) ls)
```
WhatisRT, line 287: I think there are at least three `insert-decomp` functions per direction, some with `concatMap` and some without. Please just take one for each direction without `concatMap`, and use these to prove what you need, everything will be way simpler.

WhatisRT (commented):
I looked into your usage of `All` in this PR: every theorem that invokes `All` is either used to prove a further theorem that proves an `All` or is eliminated with `All-def`. Most likely, almost everything would get a lot easier if you just prove the theorem you need, not and `All` version of it that you then need to apply `All-def` to.

There are so many lemmas that are stated and proven twice, once with `concatMap` and once without. All of that is unnecessary, just prove it without and then use `concatMap-∈ : x ∈ concatMap f l ⇔ (∃[ y ] (y ∈ l) × (x ∈ f y))` if you ever need the version with `concatMap`.

src/Data/List/Ext/Properties.agda
williamdemeo: oops... I just had it there for reference while I was working on the module. (removed)

src/Data/List/Ext/Properties.agda
williamdemeo: (replaced with `concatMap-++`)

src/Data/List/Ext/Properties.agda
williamdemeo: (removed)

src/Axiom/Set.agda
```diff
@@ -164,6 +164,9 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-fromList : ∀ {a} → a ∈ˡ l ⇔ a ∈ fromList l
   ∈-fromList = proj₂ $ listing _
 
+  ∈-fromList-tail : ∀ {a x} → a ∈ fromList l → a ∈ fromList (x ∷ l)
+  ∈-fromList-tail a∈l = to ∈-fromList (there (from ∈-fromList a∈l))
```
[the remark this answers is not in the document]
williamdemeo: no longer needed (removed)

williamdemeo:
> I looked into your usage of `All` in this PR: every theorem that invokes `All` is either used to prove a further theorem that proves an `All` or is eliminated with `All-def`. Most likely, almost everything would get a lot easier if you just prove the theorem you need, not and `All` version of it that you then need to apply `All-def` to.
> 
> There are so many lemmas that are stated and proven twice, once with `concatMap` and once without. All of that is unnecessary, just prove it without and then use `concatMap-∈ : x ∈ concatMap f l ⇔ (∃[ y ] (y ∈ l) × (x ∈ f y))` if you ever need the version with `concatMap`.

Regarding the first paragraph above, initially I had been proving the more direct lemmas you allude to, but then I realized the cleaner way would be to prove the more general `All` statements and then eliminate to get the "pointwise" versions when needed.  I still think that's the cleaner approach.  Nonetheless, for the latest commit, I've eliminated some of the `All` results and replaced them with the "pointwise" versions.  I've also removed some of the ("so many") lemmas involving `concatMap` and replaced them with the (harder-to-read) proofs involving the `concatMap-∈` lemma, as you suggested.

There are probably still some places where some small lemmas could be eliminated in favor of longer proofs elsewhere (where those lemmas are used), but I think most of the duplication and unused lemmas are now gone... though I may have missed something.

WhatisRT:
@omelkonian could you unblock this PR?

omelkonian (approved):

PR #329, opened 2023-12-08
Add `IsCommutativeMonoid'` typeclass

WhatisRT:
# Description

This sits between `CommutativeMonoid` and `IsCommutativeMonoid`. It doesn't bundle the carrier but everything else. It is a lot better behaved as a typeclass than the other two, and enables a bunch of cleanup.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Interface/IsCommutativeMonoid.agda
+++ b/src/Interface/IsCommutativeMonoid.agda
@@ -2,23 +2,25 @@
 
 module Interface.IsCommutativeMonoid where
 
-open import Algebra
+open import Algebra using (IsCommutativeMonoid; CommutativeMonoid)
 open import Prelude
 
 record IsCommutativeMonoid' c ℓ Carrier : Set (sucˡ (c ⊔ˡ ℓ)) where
-  infixl 7 _∙_
   infix  4 _≈_
   field
     _≈_                 : Carrier → Carrier → Set ℓ
-    _∙_                 : Carrier → Carrier → Carrier
-    ε                   : Carrier
-    isCommutativeMonoid : IsCommutativeMonoid {c} _≈_ _∙_ ε
+    ⦃ semigroup ⦄       : Semigroup Carrier
+    ⦃ monoid ⦄          : Monoid Carrier
+    isCommutativeMonoid : IsCommutativeMonoid {c} _≈_ _◇_ ε
 
 fromCommMonoid' : ∀ {c ℓ Carrier} → IsCommutativeMonoid' c ℓ Carrier → CommutativeMonoid c ℓ
 fromCommMonoid' c = record { IsCommutativeMonoid' c }
 
 toCommMonoid' : ∀ {c ℓ} → (m : CommutativeMonoid c ℓ) → IsCommutativeMonoid' c ℓ (CommutativeMonoid.Carrier m)
-toCommMonoid' c = record { CommutativeMonoid c }
+toCommMonoid' c = record
+  { CommutativeMonoid c using (_≈_; isCommutativeMonoid)
+  ; semigroup = λ where ._◇_ → CommutativeMonoid._∙_ c
+  ; monoid    = λ where .ε   → CommutativeMonoid.ε c }
 
 open import Data.Nat.Properties using (+-0-commutativeMonoid)
 
--- a/src/Interface/IsCommutativeMonoid.agda
+++ b/src/Interface/IsCommutativeMonoid.agda
@@ -1,0 +1,26 @@
+{-# OPTIONS --safe #-}
+
+module Interface.IsCommutativeMonoid where
+
+open import Algebra
+open import Prelude
+
+record IsCommutativeMonoid' c ℓ Carrier : Set (sucˡ (c ⊔ˡ ℓ)) where
+  infixl 7 _∙_
+  infix  4 _≈_
+  field
+    _≈_                 : Carrier → Carrier → Set ℓ
+    _∙_                 : Carrier → Carrier → Carrier
+    ε                   : Carrier
+    isCommutativeMonoid : IsCommutativeMonoid {c} _≈_ _∙_ ε
+
+fromCommMonoid' : ∀ {c ℓ Carrier} → IsCommutativeMonoid' c ℓ Carrier → CommutativeMonoid c ℓ
+fromCommMonoid' c = record { IsCommutativeMonoid' c }
+
+toCommMonoid' : ∀ {c ℓ} → (m : CommutativeMonoid c ℓ) → IsCommutativeMonoid' c ℓ (CommutativeMonoid.Carrier m)
+toCommMonoid' c = record { CommutativeMonoid c }
+
+open import Data.Nat.Properties using (+-0-commutativeMonoid)
+
+instance
+  ℕ-commMonoid' = toCommMonoid' +-0-commutativeMonoid
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -56,13 +56,13 @@ module Implementation where
 
   PlutusScript = ⊤
   ExUnits      = ℕ × ℕ
-  ExUnit-CommutativeMonoid = CommutativeMonoid 0ℓ 0ℓ ∋ record
+  ExUnit-CommutativeMonoid = IsCommutativeMonoid' 0ℓ 0ℓ ExUnits ∋ (toCommMonoid' record
     { Carrier = ExUnits
     ; _≈_ = _≈ᵖ_
     ; _∙_ = _∙ᵖ_
     ; ε = zero , zero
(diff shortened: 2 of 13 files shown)


src/Ledger/Script.lagda
```diff
@@ -40,6 +39,9 @@ record PlutusStructure : Set₁ where
     using ()
     renaming (_≈_ to _≈ᵉ_; ε to εᵉ; Carrier to ExUnits; refl to reflᵉ; _∙_ to _+ᵉˣ_)
 
+  instance
+    CommMonoid'-ExUnits : IsCommutativeMonoid' _ _ ExUnits
+    CommMonoid'-ExUnits = toCommMonoid' ExUnit-CommutativeMonoid
```
omelkonian, line 45: Now it makes more sense to add an `instance` field instead IMO:
```agda
record PlutusStructure : Set₁ where
  field Dataʰ : HashableSet
        Language PlutusScript CostModel Prices LangDepView ExUnit : Set
        ⦃ CommMonoid-ExUnits  ⦄ : IsCommutativeeMonoid 0ℓ 0ℓ ExUnit
        ⦃ Hashable-PlutusScript ⦄ : Hashable PlutusScript ScriptHash
        ⦃ DecEq-PlutusScript    ⦄ : DecEq PlutusScript
        ⦃ DecEq-CostModel       ⦄ : DecEq CostModel
        ⦃ DecEq-LangDepView     ⦄ : DecEq LangDepView
```

src/Ledger/TokenAlgebra.lagda
```diff
@@ -37,6 +37,10 @@ record TokenAlgebra : Set₁ where
     using (_≈_ ; ε ; monoid ; rawMonoid)
     renaming (Carrier to Value ; refl to reflᵛ ; _∙_ to _+ᵛ_)
 
+  instance
+    CommMonoid'-Value : IsCommutativeMonoid' _ _ Value
+    CommMonoid'-Value = toCommMonoid' Value-CommutativeMonoid
+
```
omelkonian, line 43: Same comment as in `Ledger.Script.ExUnits`: immediately declare an instance field instead of all this wrapping/unwrapping.

src/Interface/IsCommutativeMonoid.agda
```diff
@@ -0,0 +1,26 @@
...
+open import Algebra
+open import Prelude
+
+record IsCommutativeMonoid' c ℓ Carrier : Set (sucˡ (c ⊔ˡ ℓ)) where
+  infixl 7 _∙_
+  infix  4 _≈_
+  field
+    _≈_                 : Carrier → Carrier → Set ℓ
+    _∙_                 : Carrier → Carrier → Carrier
+    ε                   : Carrier
```
omelkonian, line 14: Shouldn't these operations come from superclasses, i.e. 
```agda
record IsCommutativeMonoid' c ℓ Carrier : Set (sucˡ (c ⊔ˡ ℓ)) where
  infixl 7 _∙_
  infix  4 _≈_
  field
    ⦃ setoid ⦄ : Setoid Carrier
    ⦃ semigroup ⦄ : Semigroup Carrier
    ⦃ monoid ⦄ : Monoid Carrier
    isCommutativeMonoid : IsCommutativeMonoid {c} _≈_ _◇_ ε
```
(I think we don't yet have `Setoid` so at least the other two for now)

omelkonian (changes requested):
Some comments before merging.

src/Interface/IsCommutativeMonoid.agda
williamdemeo: I like @omelkonian's suggestion.  Maybe less obvious/straightforward to understand/use/apply (at least for novices), but it seems like the "right" way to do things from an abstraction point of view.

src/Ledger/GovernanceActions/Properties.agda
```diff
@@ -22,15 +21,15 @@ instance
...
+               term ≤ s .pparams .proj₁ .PParams.ccMaxTermLength +ᵉ e ¿ of λ where
       (yes p) → just (-, Enact-NewComm p)
       (no ¬p) → nothing
     (NewConstitution dh sh)  → just (-, Enact-NewConst)
     (TriggerHF v)            → just (-, Enact-HF)
     (ChangePParams up)       → just (-, Enact-PParams)
     Info                     → just (-, Enact-Info)
     (TreasuryWdrl wdrl) →
-      case ¿ ∑[ x ← (s .withdrawals ∪⁺ wdrl) ᶠᵐ ] x ≤ t ¿ of λ where
+      case ¿ ∑[ x ← s .withdrawals ∪⁺ wdrl ] x ≤ t ¿ of λ where
```
williamdemeo, line 32: Nice! 😺

src/Ledger/GovernanceActions/Properties.agda
```diff
@@ -39,12 +38,12 @@ instance
   ... | .NewCommittee new rem q | Enact-NewComm p
     rewrite dec-yes
       (¿ ∀[ term ∈ range new ] term
-           ≤ (s .pparams .proj₁ .PParams.ccMaxTermLength +ᵉ e) ¿) p .proj₂
+           ≤ s .pparams .proj₁ .PParams.ccMaxTermLength +ᵉ e ¿) p .proj₂
```
williamdemeo, line 41: We still have to use `_+ᵉ_` around tactics? 😿

williamdemeo (approved):
Looks find to me (though I second @omelkonian's suggestion to favor definitions that build on existing ones).

src/Ledger/GovernanceActions/Properties.agda
WhatisRT: This is unrelated to tactics, it's just that we still don't have `Epoch` in our class structure.

WhatisRT:
@williamdemeo the issue here is that the monoid stuff that's defined in the stdlib is unsuitable for typeclass resolution. For typeclass resolution to behave nicely, we need to have the carrier as an argument to the class, but we don't want to pass the rest of the structure (since that's what the typeclass is supposed to provide). But the stdlib only has fully bundled (no arguments) and fully unbundled (everything except the laws as arguments) versions.

src/Ledger/Foreign/HSLedger.agda
```diff
@@ -72,23 +72,22 @@ module Implementation where
...
-    .specialAsset              → "Ada"
-    .property                  → λ _ → refl
-    .coinIsMonoidHomomorphism → λ where
-      .isMagmaHomomorphism → λ where
-        .isRelHomomorphism → record {cong = id}
-        .homo → λ _ _ → refl
-      .ε-homo → refl
+    .Value                      → ℕ
+    .Value-IsCommutativeMonoid' → it
+      -- ^ Agda bug? Without this line, `coinIsMonoidHomomorphism` doesn't type check anymore
```
WhatisRT, line 77: Here's something really stupid. Typeclass resolution will find this if the line is removed, and I checked, `Value-IsCommutativeMonoid'` normalizes to the same expression (`toCommMonoid' {zeroˡ} {zeroˡ} +-0-commutativeMonoid`) in either case.

But `coinIsMonoidHomomorphism` won't work if the line is removed. The reason is that when we provide the morphism explicitly like in the old version, this line breaks:
```agda
.isRelHomomorphism → record {cong = id}
```
What's the type of `cong`? If you replace `id` with `λ {x} {y} h → ?` and normalize the type, it'll be `h : x ≡ y ⊢ x ≡ y`, and Agda will accept `h` interactively, but when you re-check it it'll claim that `h` doesn't have the type `x ≡ y`. I also tried stuff like `let h' = x ≡ y ∋ h in h` to no avail.

@omelkonian @UlfNorell any ideas?

WhatisRT: My suspicion is that if this line is left out, Agda instantiates the typeclass later than with this line, which leads Agda to give up too early in trying to compare the two equality types.

omelkonian (approved):
LGTM

src/Ledger/Foreign/HSLedger.agda
omelkonian: That's strange, probably your suspicion is correct that it's an issue with the timing of instance resolution (as most often is). 

A quick fix for now is to simply replace the back-and-forth `toCommMonoid'/fromCommMonoid'` when opening `CommutativeMonoid` to get `rawMonoid` with an immediate `open CommutativeMonoid ℕ.+-0-commutativeMonoid using (rawMonoid)` and the problem goes away.

I will continue investigating to report a minimal Agda bug.

omelkonian: Reported as [Agda issue #7038](https://github.com/agda/agda/issues/7038); in this minimal example replacing co-patterns with `record` notation solves the error, but not in our case!

PR #330, opened 2023-12-12
Add a version of the monoid solver that tries first without normalising

UlfNorell:
This gives a significant speedup to the proofs in  `Ledger.Utxo.Properties` (from 30s total to 10s total).


WhatisRT (approved):
Ah, this is the one that we had in the past but got lost? LGTM!

UlfNorell:
Yes, this time around I made it fall back to normalising if it fails without.

PR #331, opened 2023-12-12
Split `NEWEPOCH` into `EPOCH` and `NEWEPOCH`

### LINKED ISSUE #287: `EPOCH` STS
We should add an `EPOCH` STS and split the logic that's currently in `NEWEPOCH` into `EPOCH` and `NEWEPOCH` like the Shelley spec. We should also split up `Chain.lagda` while we're at it.

WhatisRT:
# Description

This matches how it's done in the Shelley spec, and makes things a little simpler. I also went and simplified `EPOCH` as much as I could. Closes #287.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Chain.lagda
+++ b/src/Ledger/Chain.lagda
@@ -5,7 +5,6 @@
 
 open import Algebra
 open import Data.Nat.Properties using (+-0-monoid)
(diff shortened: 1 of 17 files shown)


Soupstraw (approved):
LGTM

PR #332, opened 2023-12-12
Tactic: inlining prototype

omelkonian:
~- `Script.evalTimelock`: works for all? and any?, but not MOf~
now fixed by an improved version of `MOf?` by Ulf (posted on Slack)

### DIFF
--- a/src/Ledger/Script.lagda
+++ b/src/Ledger/Script.lagda
@@ -1,6 +1,8 @@
 \section{Scripts}
 \begin{code}[hide]
 {-# OPTIONS --safe #-}
+{-# OPTIONS -v tc.unquote:10 #-}
+{-# OPTIONS -v tactic.inline:100 #-}
 
 open import Algebra using (CommutativeMonoid)
 open import Algebra.Morphism
@@ -11,8 +13,9 @@ open import Data.Nat.Properties using (+-0-commutativeMonoid; suc-injective)
 open import Data.List.Relation.Unary.MOf
 
 open import Tactic.Derive.DecEq
+open import Tactic.Inline
 
-open import Ledger.Prelude hiding (All; Any; all?; _∷ʳ_; uncons; _⊆_)
+open import Ledger.Prelude hiding (All; Any; all?; any?; _∷ʳ_; uncons; _⊆_)
 open import Ledger.Epoch
 open import Ledger.Crypto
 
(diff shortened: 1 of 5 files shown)


src/Tactic/Inline.agda
```diff
@@ -0,0 +1,279 @@
...
+
+  -- (7) [LIMITATION] does not work well with pattern lambdas
+  refl? : ℕ → Set
+  refl? n = case n ≟ n of λ where
+    (yes p) → ⊤
+    (no ¬p) → ⊥
+
+  -- unquoteDecl refl42 = inlineDecl refl42 (quoteTerm (refl? 42))
+  {-
+  refl42
```
WhatisRT, line 269: This fails because applying pattern-lambdas directly usually fails. If you generate the code
```agda
  refl42
    = case (42 ≟ 42) of
      (λ { (true because ofʸ p) → ⊤ ; (false because ofⁿ ¬p) → ⊥ })

```
it'll work.

This is especially stupid since it appears using `getDefinition` gives you some reduced form of the original definition? So in general, `defineFun n cs >>= getDefinition n` does not return `function cs`...

src/Tactic/Inline.agda
```diff
@@ -0,0 +1,279 @@
...
+open import PreludeMeta hiding (All)
+open import Interface.ToBool
+
+open Debug ("tactic.inline" , 100)
+open import Class.Show
+
+open import Algebra.Core using (Op₁)
+
+private
+  _⁉_ : ∀ {A : Set} → List A → ℕ → Maybe A
```
WhatisRT, line 20: Shouldn't this be somewhere more sensible?

src/Ledger/Script.lagda
```diff
@@ -1,6 +1,8 @@
 \section{Scripts}
 \begin{code}[hide]
 {-# OPTIONS --safe #-}
+{-# OPTIONS -v tc.unquote:10 #-}
+{-# OPTIONS -v tactic.inline:100 #-}
```
WhatisRT, line 5:
```suggestion
```

WhatisRT (approved):
Just a few minor cleanups

src/Tactic/Inline.agda
omelkonian: Nice workaround! I've now added a special case to use `case` for single-argument pat-lams and the failing example now works (+ we can use `case` like before in `MOf?` so no need to depend on our generalised `if_then_else`).

PR #333, opened 2023-12-14
Add a definition of ledger invariants

WhatisRT:
# Description

This is a small addition, just to define the notion of an invariant.

# Checklist

- [x] Self-reviewed the diff


src/Interface/STS.agda
```diff
@@ -73,27 +71,35 @@ data IdSTS {C S} : C → S → ⊤ → S → Set where
...
+ReflexiveTransitiveClosureᵢ-total SS-total {s = s} {x ∷ sig} =
+  case SS-total of λ where
+    (s' , Ps') → map₂′ (BS-ind Ps') $ ReflexiveTransitiveClosureᵢ-total SS-total
 
 -- with a given base case
 ReflexiveTransitiveClosureᵢᵇ = _⊢_⇀⟦_⟧ᵢ*_
 ReflexiveTransitiveClosureᵇ  = _⊢_⇀⟦_⟧*_
+
+LedgerInvariant : (C → S → Sig → S → Set) → (S → Set) → Set
+LedgerInvariant STS P = ∀ {c s sig s'} → STS c s sig s' → P s → P s'
```
omelkonian, line 100: I am slightly worried that here the invariant does not necessarily hold for base cases; we only it is preserved through all inductive steps. But I guess this is what we actually want..

omelkonian (approved):
LGTM

src/Interface/STS.agda
WhatisRT: Yeah, you can usually make a state that doesn't satisfy the property. It's just that it's preserved by the state machine.

PR #334, opened 2023-12-14
Add notation section & remove footnote remarks in favor of it

WhatisRT:
# Description

Lots of notations still need to be explained (or removed), but this is a start.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Notation.lagda
+++ b/src/Ledger/Notation.lagda
@@ -1,0 +1,27 @@
+\section{Notation}
+\label{sec:notation}
+\begin{code}[hide]
+{-# OPTIONS --safe #-}
+
+module Ledger.Notation where
+\end{code}
+
+In this section, we introduce some notations used in this document.
+
+\begin{description}
+\item[Propositions, sets and types] This document loosly treats sets and types as the same thing.
+\item[Sums and products] The sum (or disjoint union, coproduct, etc.) of \AgdaBound{A} and
+  \AgdaBound{B} is denoted by \AgdaBound{A} \coproduct \AgdaBound{B}, and their product
+  is denoted by \AgdaBound{A} \agdatimes \AgdaBound{B}.
+\item[Record types] Record types are explained in Appendix~\ref{sec:appendix-agda-essentials}.
+\item[Postfix projections] Projections can be written using postix notation. For example, we may
+  write \AgdaBound{x}\AgdaSpace{}\AgdaSymbol{.}\AgdaField{proj₁} instead of
+  \AgdaField{proj₁}\AgdaSpace{}\AgdaBound{x}.
+\item[Restriction, corestriction and complements] The restriction of a function or map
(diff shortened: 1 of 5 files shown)


src/Ledger/Notation.lagda
```diff
@@ -0,0 +1,24 @@
...
+
+In this section, we introduce some notations used in this document.
+
+\begin{description}
+\item[Propositions, sets and types] This document loosly treats sets and types as the same thing.
+\item[Sums and products] The sum of \AgdaBound{A} and \AgdaBound{B} (also called their disjoint
+  union, coproduct, etc.) is denoted by \AgdaBound{A} \coproduct \AgdaBound{B}, and their product
+  is denoted by \AgdaBound{A} \agdatimes \AgdaBound{B}.
+\item[Record types] Record types are explained in Appendix \ref{sec:appendix-agda-essentials}.
+\item[Postfix projections] TODO, if we want them in the PDF
```
omelkonian, line 16: Add a normal sentence for postfix projections (e.g. "Record fields are simply projection functions from a value of the record to said field and can be conveniently written in postfix form; for instance, getting the e-mail `address` of a `user` would be denoted `user .address`"); we can then remove it if we decide no such projections are ever rendered on PDF.

omelkonian (changes requested):

src/Ledger/Notation.lagda
```diff
@@ -0,0 +1,27 @@
...
+module Ledger.Notation where
+\end{code}
+
+In this section, we introduce some notations used in this document.
+
+\begin{description}
+\item[Propositions, sets and types] This document loosly treats sets and types as the same thing.
+\item[Sums and products] The sum of \AgdaBound{A} and \AgdaBound{B} (also called their disjoint
+  union, coproduct, etc.) is denoted by \AgdaBound{A} \coproduct \AgdaBound{B}, and their product
+  is denoted by \AgdaBound{A} \agdatimes \AgdaBound{B}.
```
omelkonian, line 15: I would rephrase this as "The product of  \AgdaBound{A} and \AgdaBound{B}  is denoted by \AgdaBound{A} \agdatimes \AgdaBound{B}; their sum (a.k.a. disjoint union or coproduct) is denoted \AgdaBound{A} \coproduct \AgdaBound{B}."

src/Ledger/Notation.lagda
```diff
@@ -0,0 +1,27 @@
...
+\item[Record types] Record types are explained in Appendix \ref{sec:appendix-agda-essentials}.
```
omelkonian, line 16: ...in Appendix~\ref...

(`~` should generally be used consistently  with `\ref`)

omelkonian (approved):
Some small changes, otherwise LGTM.

src/Ledger/Notation.lagda
WhatisRT: I prefer the original order, since that's the order of the header as well. But I did slightly rephrase it.

PR #336, opened 2023-12-15
Various PDF fixes

### LINKED ISSUE #303: PDF improvement: add a proper bibliography

WhatisRT:
# Description

- Replace `$` with parentheses
- Add bibliography (Closes #303)
- Aligning & reformatting figures

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -35,9 +35,8 @@ record GovEnv : Set where
 data
 \end{code}
 \begin{code}
-  _⊢_⇀⦇_,GOV'⦈_ : GovEnv × ℕ → GovState → GovVote ⊎ GovProposal → GovState → Set
-
-_⊢_⇀⦇_,GOV⦈_ : GovEnv → GovState → List (GovVote ⊎ GovProposal) → GovState → Set
+  _⊢_⇀⦇_,GOV'⦈_  : GovEnv × ℕ → GovState → GovVote ⊎ GovProposal → GovState → Set
+_⊢_⇀⦇_,GOV⦈_     : GovEnv → GovState → List (GovVote ⊎ GovProposal) → GovState → Set
 \end{code}
 \begin{code}[hide]
 open GovActionState
@@ -105,24 +104,22 @@ compatible version.
 data _⊢_⇀⦇_,GOV'⦈_ where
 \end{code}
 \begin{code}
-  GOV-Vote : ∀ {x ast} →  let  open GovEnv Γ
-                               sig = inj₁ record  { gid = aid ; role = role
-                                                  ; credential = cred
-                                                  ; vote = v ; anchor = x }
-                          in
-    (aid , ast) ∈ fromList s
-    → canVote pparams (action ast) role
+  GOV-Vote : ∀ {x ast} → let
+    open GovEnv Γ
+    sig = inj₁ record
+      { gid = aid ; role = role ; credential = cred ; vote = v ; anchor = x }
+    in
+       (aid , ast) ∈ fromList s
+    →  canVote pparams (action ast) role
     ───────────────────────────────────────
     (Γ , k) ⊢ s ⇀⦇ sig ,GOV'⦈ addVote s aid role cred v
 
-  GOV-Propose : ∀ {x} →  let  open GovEnv Γ
-                              open PParams pparams hiding (a)
-                              prop = record  { returnAddr = addr ; action = a
-                                             ; anchor = x ; deposit = d
-                                             ; prevAction = prev }
-                              s' = addAction  s (govActionLifetime +ᵉ epoch)
-                                              (txid , k) addr a prev
-                         in
+  GOV-Propose : ∀ {x} → let
+    open GovEnv Γ; open PParams pparams hiding (a)
+    prop = record
+      { returnAddr = addr ; action = a ; anchor = x ; deposit = d ; prevAction = prev }
+    s' = addAction s (govActionLifetime +ᵉ epoch) (txid , k) addr a prev
+    in
        actionWellFormed a ≡ true
     →  d ≡ govActionDeposit
     →  (∀ {new rem q} → a ≡ NewCommittee new rem q
--- a/src/Ledger/Introduction.lagda
+++ b/src/Ledger/Introduction.lagda
@@ -19,12 +19,41 @@ private variable
   sigs : List Sig
 \end{code}
 
-Repository: https://github.com/input-output-hk/formal-ledger-specifications
+\begin{center}
+Repository: \url{https://github.com/IntersectMBO/formal-ledger-specifications}
+\end{center}
+This is the work-in-progress specification of the Cardano ledger. The
+current status of each individual era is described in Table \ref{fig:eras-progress}.
 
-This document describes the formalization of the Cardano ledger specification in the Agda programming
-language and proof assistant. The specification formalized here is that of the Conway era, described
-in detail in the Cardano Improvement Proposal (CIP) 1694,
-\href{https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694}{github.com/cardano-foundation/CIPs/CIP-1694}.
+\begin{longtable}[h!]{|l l l l|}
+\hline
+Era  & Figures & Prose & Cleanup \\
+\hline
+\endhead
+Shelley & Partial & Partial & Not started \\
+Shelley-MA & Partial & Partial & Not started \\
+Alonzo & Partial & Partial & Not started \\
+Babbage & Not started & Not started & Not started \\
+Conway \cite{cip1694} & Complete & Partial & Partial \\
+\hline
+\caption{Specification progress}
+\label{fig:eras-progress}
+\end{longtable}
+
+\subsection{A note on Agda}
+
+This specification is written using the Agda programming language and
+proof assistant \cite{agda2023}. We have spent a lot of time on making
+this document readable for people unfamiliar with Agda (or other proof
+assistants, functional programming languages, etc.). However, by the
+nature of working in a formal language we have to play by its rules,
+meaning that some instances of uncommon notation are very difficult or
+impossible to replace. Some are explained in
+Section~\ref{sec:notation}, but there is no guarantee that this
+section is complete. Anyone who is confused by the meaning of an
+expression, please feel free to open an issue in our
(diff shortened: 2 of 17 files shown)


src/Ledger/Introduction.lagda
```diff
@@ -19,12 +19,38 @@ private variable
...
+proof assistant \cite{agda2023}. We have spent a lot of time on making this document
+readable for people unfamiliar with Agda (or other proof assistants,
+functional programming languages, etc.). However, by the nature of
+working in a formal language we have to play by its rules, meaning
+that some instances of uncommon notation are very difficult or
+impossible to replace with other notations. Some of those notations
+are explained in Section \ref{sec:notation}, but there is no guarantee
+that this section is complete. Anyone who is confused by the meaning
+of an expression, please feel free to open an issue in our
+\href{https://github.com/input-output-hk/formal-ledger-specifications/issues}{repository}
```
omelkonian, line 52: With the current Latex config, the only way to discover `hrefs` is to hover the mouse on top. I hope this is not intended and the proper fix is to highlight the URLs (typically in blue-ish colour), e.g. by using the `hyperrefs` package:
```latex
\usepackage{hyperref}
\hypersetup{
    colorlinks=true,
    linkcolor=blue,
    filecolor=magenta,      
    urlcolor=cyan
 }
```

src/Ledger/Gov.lagda
```diff
@@ -105,24 +104,22 @@ compatible version.
...
-                         in
+  GOV-Propose : ∀ {x} → let
+    open GovEnv Γ; open PParams pparams hiding (a)
+    prop = record
+      { returnAddr = addr ; action = a ; anchor = x ; deposit = d ; prevAction = prev }
+    s' = addAction s (govActionLifetime +ᵉ epoch) (txid , k) addr a prev
+    in
        actionWellFormed a ≡ true
     →  d ≡ govActionDeposit
     →  (∀ {new rem q} → a ≡ NewCommittee new rem q
```
omelkonian, line 125: Note that the next section of 'Delegation' starts with an empty page, which is probably not desirable.

src/Ledger/Gov.lagda
```diff
@@ -105,24 +104,22 @@ compatible version.
...
 
-  GOV-Propose : ∀ {x} →  let  open GovEnv Γ
-                              open PParams pparams hiding (a)
-                              prop = record  { returnAddr = addr ; action = a
-                                             ; anchor = x ; deposit = d
-                                             ; prevAction = prev }
-                              s' = addAction  s (govActionLifetime +ᵉ epoch)
-                                              (txid , k) addr a prev
-                         in
+  GOV-Propose : ∀ {x} → let
```
omelkonian, line 117: The formatting of `let` seems weird to me and inconsistent across the document; I would suggest the following two versions to be applied uniformly and consistently:
1. for very small binding, just keep it in the same line as the rule/constructor, e.g.
```agda
RULE: let x = 5 in ...
...
```
2. for longer cases, **always** use the following format:
```agda
RULE: 
  let
    BINDING #1
    BINDING #2
    ...
    BINDING #k
  in
    ...
```
(if you cannot spare the two spaces caused by the suggested indentation, the problem lies elsewhere..)

---

While I'm at it, I saw some occurrences of the function arrow for rule hypotheses, which should be uniformly replaced with `∙` in my opinion and formatted the same way as `LEDGER` currently is, i.e.
```agda
RULE : let ... in
  ∙ HYPOTHESIS #1
  ∙ HYPOTHESIS #2
 
  ∙ HYPOTHESIS #n
   CONCLUSION
```
(notice how the 'dots' lie outside and hypothesis, horizontal rule and conclusion are all lined up.

src/Ledger/Utxo.lagda
```diff
@@ -130,14 +126,15 @@ module _ (let open Tx; open TxBody) where
...
 \begin{code}
--- this has to be a type definition for inference to work
 data inInterval (slot : Slot) : (Maybe Slot × Maybe Slot) → Set where
   both   : ∀ {l r}  → l ≤ slot × slot ≤ r  →  inInterval slot (just l   , just r)
   lower  : ∀ {l}    → l ≤ slot             →  inInterval slot (just l   , nothing)
   upper  : ∀ {r}    → slot ≤ r             →  inInterval slot (nothing  , just r)
   none   :                                    inInterval slot (nothing  , nothing)
 \end{code}
 \begin{code}[hide]
+-- Note: inInterval this has to be a type definition for inference to work
```
omelkonian, line 136: inInterval ~this~ has to be ...

src/Ledger/Utxo.lagda
```diff
@@ -314,17 +311,14 @@ data _⊢_⇀⦇_,UTXO⦈_ where
...
     ∙ ∀[ (a , _) ∈ range txouts ]
         Sum.All (const ⊤) (λ a → a .BootstrapAddr.attrsSize ≤ 64) a
-    ∙ ∀[ (a , _) ∈ range txouts ] netId a        ≡ networkId
-    ∙ ∀[ a ∈ dom  txwdrls ]       a .RwdAddr.net ≡ networkId
-    -- Add deposits
-
+    ∙ ∀[ (a , _) ∈ range txouts ]  netId a         ≡ networkId
+    ∙ ∀[ a ∈ dom  txwdrls ]        a .RwdAddr.net  ≡ networkId
        ────────────────────────────────
        Γ ⊢ s ⇀⦇ tx ,UTXO⦈  ⟦ (utxo ∣ txins ᶜ) ∪ˡ outs txb
```
omelkonian, line 317: The vertical alignment of this whole rule seems off; it should be like so:
```agda
    ∙  txins ≢ ∅                              ∙ txins ⊆ dom utxo
    ∙  inInterval slot txvldt                 ∙ minfee pp tx ≤ txfee
    ∙  consumed pp s txb ≡ produced pp s txb  ∙ coin mint ≡ 0
    ∙  txsize ≤ maxTxSize pp
    ∙  ∀[ (_ , txout) ∈ txouts .proj₁ ]
         inject (utxoEntrySize txout * minUTxOValue pp) ≤ᵗ getValue txout
    ∙  ∀[ (_ , txout) ∈ txouts .proj₁ ]
         serSize (getValue txout) ≤ maxValSize pp
    ∙  ∀[ (a , _) ∈ range txouts ]
         Sum.All (const ⊤) (λ a → a .BootstrapAddr.attrsSize ≤ 64) a
    ∙  ∀[ (a , _) ∈ range txouts ]  netId a         ≡ networkId
    ∙  ∀[ a ∈ dom  txwdrls ]        a .RwdAddr.net  ≡ networkId
       Γ ⊢ s ⇀⦇ tx ,UTXO⦈  ⟦ (utxo ∣ txins ᶜ) ∪ˡ outs txb
                           , fees + txfee
                           , updateDeposits pp txb deposits
                           , donations + txdonation
                           ⟧ᵘ
```

omelkonian (changes requested):

src/Ledger/Introduction.lagda
```diff
@@ -19,12 +19,38 @@ private variable
...
-This document describes the formalization of the Cardano ledger specification in the Agda programming
-language and proof assistant. The specification formalized here is that of the Conway era, described
-in detail in the Cardano Improvement Proposal (CIP) 1694,
-\href{https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694}{github.com/cardano-foundation/CIPs/CIP-1694}.
+This is the work-in-progress specification of the Cardano ledger. The
+current status of each individual era is described in Table \ref{fig:eras-progress}.
+
+\begin{longtable}[h!]{|l l l l|}
+\hline
+Era  & Figures & Prose & Cleanup \\
```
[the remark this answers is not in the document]
WhatisRT: I think we'll need to make a separate PDF just for the Conway spec at this point, because anything this incomplete probably shouldn't be a document to point people at. The point of this table is simply to make it clear where this document is currently at, so when people go looking for Babbage-era stuff they don't accidentally go look here.

src/Ledger/Introduction.lagda
```diff
@@ -19,12 +19,38 @@ private variable
...
+\hline
+\endhead
+Shelley & Partial & Partial & Not started \\
+Shelley-MA & Partial & Partial & Not started \\
+Alonzo & Partial & Partial & Not started \\
+Babbage & Not started & Not started & Not started \\
+Conway \cite{cip1694} & Complete & Partial & Partial \\
```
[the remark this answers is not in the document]
WhatisRT: 1. There are different terminologies for different things. Maybe we can explain them somewhere, but I don't really want to do that now. There was a link to the CIP somewhere in the spec before, and I moved it to the bibliography.
2. See 1. I'll make an issue.

src/Ledger/Gov.lagda
WhatisRT: Yeah, I'm aware. I didn't try to fix everything in this PR.

src/Ledger/Gov.lagda
WhatisRT: Made an issue

src/Ledger/Utxo.lagda
```diff
@@ -85,11 +85,7 @@ module _ (let open Tx; open TxBody) where
   coinPolicies = policies (inject 1)
 
   isAdaOnlyᵇ : Value → Bool
-  isAdaOnlyᵇ v =
-    if (policies v) ≡ᵉ coinPolicies then
-      true
-    else
-      false
+  isAdaOnlyᵇ v = toBool (policies v ≡ᵉ coinPolicies)
```
omelkonian, line 88: I missed this in my first review: the inner workings of `toBool` are prohibitively complex for the PDF readers and  that goes against the PDF code being self-contained, hence why I used the Haskell "anti-pattern" `if b then true else false` which linters like `hlint` would report as a code smell that can be fixed by replacing the whole thing with just `b`.

WhatisRT: Yeah, but the whole property shouldn't be a boolean in the first place. This should all go to `Set` and then this issue will go away (and only `policies v ≡ᵉ coinPolicies` will remain). So this is a step in the right direction, I just didn't want to refactor all that as part of this PR.

omelkonian: Oh, I completely agree with that, issue resolved :)

omelkonian (approved):
LGTM

PR #352, opened 2024-01-11
Add `minFeeRefScriptsCoinsPerByte`

### LINKED ISSUE #345: Add `minFeeRefScript` parameter & add it to `minfee`
We need to add a term `minFeeRefScriptCoinsPerByte * totalRefInputSize` to `minfee`. Since we don't have any Babbage features yet, we cannot compute the total size of reference scripts. One way around this would be to write this part in LaTeX only, and migrate it to Agda later once we have added reference scripts.

WhatisRT:
# Description

This is a small addition to the fee calculation in Conway. Since we don't have Babbage features the `refScripts` function is just the constant empty set ATM. Also, since we don't have a way to compute a size for scripts in general, `scriptSize` is just an abstract function.

One annoyance is that we don't have types & functions to deal with rational numbers in the unit interval. So the new parameter doesn't have that constraint for now, but we should probably add that later.

Closes #345 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Utxo.lagda
+++ b/src/Ledger/Utxo.lagda
@@ -12,6 +12,9 @@ open import Data.Nat.Properties  using (+-0-monoid)
 import Data.Maybe as M
 import Data.Sum.Relation.Unary.All as Sum
 
+import Data.Integer as ℤ
+import Data.Rational as ℚ
+
 open import Tactic.Derive.DecEq
 
 open import Ledger.Prelude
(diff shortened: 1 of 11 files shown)


WhatisRT:
Hmm, seems like Hydra is broken again & the GH runner runs out of memory. Maybe I'll spend a day on using `opaque ... unfolding ...` to improve this.

Soupstraw (approved):
LGTM

src/Ledger/Utxo.lagda
```diff
@@ -49,7 +58,10 @@ utxoEntrySizeWithoutVal : MemoryEstimate
 utxoEntrySizeWithoutVal = 8
 
 utxoEntrySize : TxOut → MemoryEstimate
-utxoEntrySize utxo = utxoEntrySizeWithoutVal + size (getValue utxo)
+utxoEntrySize o = utxoEntrySizeWithoutVal + size (getValue o)
+
+refScripts : Tx → UTxO → ℙ Script
+refScripts tx utxo = ∅ -- TODO: implement when we do Babbage
```
lehins, line 64: I think it would be good adding a documentation saying that this is only the scripts that are actually used by the transaction, i.e. native scripts and plutus scripts that have a script purpose.

PR #356, opened 2024-02-08
PDF cleanup 282: rename `filterᵐ` to `filter`

williamdemeo:
# Description

Create a type class called HasFilter for types with filter operations.

This addresses one item of #282 

# Checklist

- [ ] Self-reviewed the diff


src/Interface/HasFilter.agda
```diff
@@ -0,0 +1,16 @@
+{-# OPTIONS --safe #-}
+
+open import Agda.Primitive using () renaming (Set to Type)
+open import Axiom.Set using (Theory)
+
+module Interface.HasFilter {ℓ} (th : Theory{ℓ}) where
```
WhatisRT, line 6: This module shouldn't be quantified over those two things

src/Interface/HasFilter/Instance.agda
```diff
@@ -0,0 +1,23 @@
...
+
+open import Data.Product using (_×_; _,_)
+open import Data.List using (List) renaming (filter to filterˡ)
+open import Relation.Unary using (Decidable)
+open import Axiom.Set.Map th
+open import Interface.HasFilter th
+open Theory th renaming (filter to filterˢ)
+private variable A B : Type
+
+instance
```
WhatisRT, line 15: There should also be an instance for `List` here, which fixes the problem that the `filter` function for lists needs to be used qualified.

WhatisRT (changes requested):

src/Interface/HasFilter/Instance.agda
williamdemeo: Yeah, what's weird is that I have an instance for List... but it doesn't seem to be working.  I'll keep at it...

src/Interface/HasFilter/InstanceTh.agda
```diff
@@ -0,0 +1,20 @@
...
+open import Interface.HasFilter
+open Theory th renaming (filter to filterˢ)
+private variable A B : Type
+
+instance
+  mapFilter : HasFilter (A × B) (Map A B)
+  mapFilter = record { predCheck = specProperty; filter = filterᵐ}
+
+  setFilter : HasFilter A (Set A)
+  setFilter = record { predCheck = specProperty; filter = filterˢ}
```
williamdemeo, line 20: I'm confused about when I need to have an explicit theory (`th`) in context.  I don't see how to create these filter instances for `Map` and `Set` unless I have a `th` around.  And, assuming this is correct, and I need to parameterize this module with a `th`, how do I get it to work with, e.g., the `Transaction` module which does not have an explicit `th`?

WhatisRT: It comes from `Ledger.Set.Theory`. So if you have a module that's quantified over a set model, you could do `open Interface.HasFilter.InstanceTh {List-Model} public`. Probably the best place to do this is in `Ledger.Prelude`, it already has some instance imports like that at the bottom.

williamdemeo: Okay, I'll give that a try.  Thanks.

src/Axiom/Set.agda
```diff
@@ -4,7 +4,7 @@ open import Agda.Primitive renaming (Set to Type)
 
 module Axiom.Set where
 
-open import Prelude hiding (filter; map)
```
williamdemeo, line 7: `Prelude` no longer imports `Data.List.filter`. Instead, we import the kind of filter we need where we need it, and we name it `filter`.

williamdemeo:
This works now, hides superscripts on `filter`, without the need for type classes.  I think it's ready to be merged (once I rebase/merge on the latest master).

WhatisRT (approved):
I'd have preferred a typeclass, but I'll merge this now since it does improve the PDF.

PR #358, opened 2024-02-09
PDF cleanup 282: remove `uncurry` from code appearing in pdf

williamdemeo:
# Description

This is a requirement of issue #282.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Foreign/HSLedger.agda
+++ b/src/Ledger/Foreign/HSLedger.agda
@@ -238,10 +238,10 @@ instance
 
   Convertible-VDeleg : Convertible VDeleg F.VDeleg
   Convertible-VDeleg = λ where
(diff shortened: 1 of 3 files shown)


WhatisRT (commented):
Changing the type of `credVoter` seems to be a pretty heavy solution for making documentation more accessible. I'm not necessarily against this change, but I'll need to think more about it.

williamdemeo:
> Changing the type of `credVoter` seems to be a pretty heavy solution for making documentation more accessible.

I totally agree and that was exactly my sentiment as I was making that change.  This was just the first approach that came to mind, but if you think of a better one, I'm open to suggestions.  It certainly seems like it shouldn't be very hard to avoid calling `uncurry`; it's such a simple operation.

WhatisRT:
Closing this - the original issue has been closed and I don't know if it's even relevant anymore

PR #361, opened 2024-02-09
PDF cleanup 282: rename `<$>` in Fig 25

williamdemeo:
# Description

This addresses one item of #282 by simply replacing `<$>` in Fig 25 with the `．` symbol.

# Checklist

- [X] Self-reviewed the diff


WhatisRT:
Adding a different symbol that people don't know isn't the way to go here. I'll close this since there isn't anything that's worth keeping.

PR #362, opened 2024-02-09
PDF cleanup 282: rename `_≗_` to `_≡_`

williamdemeo:
# Description

Addresses one item in issue #282 

I'm conflicted about this change since, while it might make the pdf easier for non-Agda/non-TT folks to read, it could cause confusion for Agda/TT folks who know that the usual `_≡_` wouldn't work here.

# Checklist

- [X] Self-reviewed the diff


WhatisRT:
There are a bunch of other solutions here, such as writing that property in prose instead of the figure.

PR #363, opened 2024-02-14
PDF cleanup 282: rename `sucᵉ` to `1 +ᵉ_`

williamdemeo:
# Description

Second attempt to rename `sucᵉ`, this time using a simpler approach---simply replace it with `1 +ᵉ_`.

Maybe this is not idea, but I'm not sure there's a better alternative if we can't identify `Epoch` as one of the simple/standard types of algebraic structures for which there is already support for literals.

It was suggested in the comments of previous PR #315 (since closed) that `Epoch` has a module structure, but I don't see it.  To me `Epoch` looks more like a G-set, where the group acting on `Epoch` is ℕ` ...and I don't know if there's support for G-set literals ...yet.  If there is
 not, then we could develop it, but I'm not sure it's worth doing for the sole purpose of renaming one symbol.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -99,7 +99,7 @@ its results, i.e:
       certState' =
         ⟦ record dState { rewards = rewards ∪⁺ refunds }
         , ⟦ pools ∣ retired ᶜ , retiring ∣ retired ᶜ ⟧ᵖ
-        , ⟦ if null govSt' then mapValues sucᵉ dreps else dreps
+        , ⟦ if null govSt' then mapValues (1 +ᵉ_) dreps else dreps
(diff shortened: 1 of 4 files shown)


WhatisRT:
It'd be better to instead make an epoch `1ᵉ`, so that you can write `1ᵉ +_` for `sucᵉ`. That way we can then just use literal overloading to get rid of the `ᵉ` entirely.

williamdemeo:
> It'd be better to instead make an epoch `1ᵉ`, so that you can write `1ᵉ +_` for `sucᵉ`. That way we can then just use literal overloading to get rid of the `ᵉ` entirely.

Okay, I'll try that. 👍🏼

WhatisRT:
Closing in favour of #387

PR #364, opened 2024-02-14
PDF cleanup 282: hide use of `open` for records in pdf

williamdemeo:
# Description

Hide use of `open` for records in pdf.  This addresses one item of issue #282.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -50,6 +50,8 @@ private variable
 
 instance _ = +-0-monoid; _ = +-0-commutativeMonoid
 
+open Acnt; open DState; open GState; open PState; open RatifyState
+
 data _⊢_⇀⦇_,EPOCH⦈_ : NewEpochEnv → EpochState → Epoch → EpochState → Set where
 \end{code}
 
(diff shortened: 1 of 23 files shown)


WhatisRT (changes requested):
`open` can't just be hidden from view, since it carries a lot of meaning. For example, in `EPOCH` it does some of the actual work. Rather, we should use techniques that avoid using `open` in the code, like for example pattern-matching on a state/environment when defining rules, instead of binding it to a single name and then writing `open State s`. Another option would be to use some kind of deep accessors to avoid nested `open`s and then applying those deep accessors to our variables. What's right needs to be decided on a case-by-case basis.

williamdemeo:
Okay, I tried being less aggressive about hiding open statements, especially when applied to a private variable.  I'm now hiding some open statements without naming the private variable, using accessors to make it more clear what's happening.

src/Ledger/Epoch.lagda
```diff
@@ -74,45 +76,43 @@ its results, i.e:
 \begin{code}
   EPOCH : let
       open EpochState eps hiding (es)
-      open RatifyState fut using (removed) renaming (es to esW)
       -- ^ this rolls over the future enact state into es
       open LState ls; open UTxOState utxoSt; open CertState certState
-      open PState pState; open DState dState; open GState gState
-      open Acnt acnt
```
williamdemeo, line 81: Removed a couple of instances of `open` for records here, at the expense of using accessors below.

src/Ledger/Gov.lagda
```diff
@@ -110,13 +111,13 @@ data _⊢_⇀⦇_,GOV'⦈_ where
       (Γ , k) ⊢ s ⇀⦇ sig ,GOV'⦈ addVote s aid voter v
 
   GOV-Propose : ∀ {x} → let
-      open GovEnv Γ; open PParams pparams hiding (a)
+      open GovEnv Γ
```
williamdemeo, line 114: Remove one instance of `open` for record `pparams` at the expense of having to write `pparams` a couple of times below.

src/Ledger/Ledger.lagda
```diff
@@ -44,6 +44,8 @@ record LState : Set where
 
 txgov : TxBody → List (GovVote ⊎ GovProposal)
 txgov txb = map inj₁ txvote ++ map inj₂ txprop
+\end{code}
+\begin{code}[hide]
   where open TxBody txb
```
williamdemeo, line 49: It doesn't seem necessary to show this `open` directive in the figure.

src/Ledger/Ledger.lagda
```diff
@@ -60,9 +62,7 @@ private variable
 
 \begin{figure*}[h]
 \begin{code}[hide]
-open RwdAddr
-open DState
-open CertState
+open CertState; open DState; open LEnv; open LState; open RwdAddr; open TxBody
```
williamdemeo, line 65: Remove five instances of `open` for records from the figure at the expense of lots explicit use of accessors in the `LEDGER` transition rule below.  Is it worth it?  Is there a better way to avoid the `open` directives here?

src/Ledger/TokenAlgebra/ValueSet.lagda
```diff
@@ -56,37 +57,45 @@ record AdaIdType : Type where
...
 Quantity = ℕ
 
 module _
   {X : ℙ AssetId}
   {⋁A : isMaximal X}
+\end{code}
+\begin{code}[hide]
   ⦃ DecEq-PolicyId  : DecEq PolicyId ⦄
   ⦃ DecEq-AssetName : DecEq AssetName ⦄
   ⦃ DecEq-Tot : DecEq (AssetId ⇒ ℕ) ⦄
```
williamdemeo, line 74: Do these really need to appear in the figure?

williamdemeo:
I'm closing this PR because it's outdated and none of the changes were difficult or time-consuming.  However, I'll copy @WhatisRT's comment about preserving some `open` lines, and include that in the most recent pdf-cleanup issue that mentions the `open` keyword.

PR #365, opened 2024-02-14
Address some comments from the audit

WhatisRT:
# Description

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Introduction.lagda
+++ b/src/Ledger/Introduction.lagda
@@ -78,14 +78,14 @@ other, forming a directed graph that is almost a tree.
 
 \subsection{Reflexive-transitive closure}
 
-Some STS relations need to be applied as many times as they can to
-arrive at a final state. Since we use this pattern multiple times, we
-define a closure operation which takes a STS relation and applies it
-as many times as possible.
+Some STS (state transition system) relations need to be applied as
+many times as they can to arrive at a final state. Since we use this
+pattern multiple times, we define a closure operation which takes a
+STS relation and applies it as many times as possible.
 
-The closure is defined in Figure \ref{fig:rt-closure}, assuming a
-relation \RTCB. In the remainder of the text, the closure operation is
-called \RTC.
+The closure \RTCI of a relation \RTCB is defined in Figure
+\ref{fig:rt-closure}. In the remainder of the text, the closure
+operation is called \RTC.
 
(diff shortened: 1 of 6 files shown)


src/Ledger/Introduction.lagda
```diff
@@ -180,7 +180,8 @@ irrelevant. Additionally, all sets in this specification are finite.
 We use this set theory to define maps as seen below, which are used in
 many places. We usually think of maps as partial functions
 (i.e. functions not defined everywhere), but importantly they are not
-Agda functions.
+Agda functions. We denote the powerset operation by \PowerSet, which
+we use here to form a type of sets with elements in a given type.
```
williamdemeo, line 184: The phrase "type of sets with elements in a give type" sounds a bit odd... like a wordy way of describing the type of sets.  I mean, aren't all sets collections elements of a given type?  Are we using \PowerSet here in the usual (set of all subsets) sense?

src/Ledger/PParams.lagda
```diff
@@ -129,7 +130,10 @@ The first four groups have the property that every protocol parameter
 is associated to precisely one of these groups. The \SecurityGroup is
 special: a protocol parameter may or may not be in the
 \SecurityGroup. So, each protocol parameter belongs to at least one
-and at most two groups.
+and at most two groups. Note that in \cite{cip1694} there is no
+\SecurityGroup, but there is the concept of security-relevant protocol
+parameters. The difference between these notions is only social, so we
```
williamdemeo, line 135: I'm not sure what is meant by a difference being "social."  Of course, I can take a guess at what it means... maybe that's good enough?

williamdemeo (approved):
Just two minor comments.  Otherwise, looks good.

src/Ledger/Introduction.lagda
WhatisRT: In classical set theory there are no types, so this sentence would be nonsensical. `{1, ℕ → ℕ}` is a perfectly fine set. And our set theory could be one where this is allowed, but it isn't.

src/Ledger/PParams.lagda
WhatisRT: Yeah, that's good enough. Basically, we weren't allowed to call it a group in the CIP because the groups have non-technical stuff attached to them.

PR #368, opened 2024-02-15
PDF cleanup 282: hide `Σ`

williamdemeo:
# Description

One way to hide `Σ` is to simply use an alternative `Σ-syntax` definition.  

This PR does this in the two places where `Σ` had appeared in a pdf figure (`Ledger.Introduction` and `Ledger.Address`).

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Address.lagda
+++ b/src/Ledger/Address.lagda
@@ -55,13 +55,19 @@ record RwdAddr : Set where
 \end{code}
 \begin{code}[hide]
 open BaseAddr; open BootstrapAddr; open BaseAddr; open BootstrapAddr
+private variable
+  a b : Level
+  A : Set a
+Σ-syntax' : (A : Set a) → (A → Set b) → Set _
+Σ-syntax' = Σ
+syntax Σ-syntax' A (λ x → B) = x ∈ A ﹐ B
 \end{code}
 \begin{code}
 
-VKeyBaseAddr         = Σ[ addr ∈ BaseAddr       ] isVKey    (addr .pay)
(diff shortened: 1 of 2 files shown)


WhatisRT:
How about using a set builder notation instead? Something like `Σ A P = ｛ a ∈ A ∣ P a ｝`?

williamdemeo:
> How about using a set builder notation instead? Something like `Σ A P = ｛ a ∈ A ∣ P a ｝`?

We could do that.  However, I think that choice, like the choice I made initially, doesn't really look like an inhabitant of a Sigma type---a dependent pair.  I think it would be better if we could just write a pair, like `(a , P a)` in the example you give, but I'm not sure that's possible.  Anyway, I'll see if Agda accepts the set-builder notation you suggest.

omelkonian:
> I think it would be better if we could just write a pair, like `(a , P a)` in the example you give, but I'm not sure that's possible.

Nowadays it is actually [possible](https://agda.readthedocs.io/en/v2.6.4/language/syntax-declarations.html) to do that!

williamdemeo:
> > I think it would be better if we could just write a pair, like `(a , P a)` in the example you give, but I'm not sure that's possible.
> 
> Nowadays it is actually [possible](https://agda.readthedocs.io/en/v2.6.4/language/syntax-declarations.html) to do that!

I was only able to get this to work in the `Ledger.Address` module by making a mess of the code (due to the fact that we're using a generalizable variable `addr` here).  If either of you can see how to make it work and look clean, please advise.

WhatisRT (approved):
LGTM

PR #369, opened 2024-02-15
PDF cleanup 282: hide `data` keyword

williamdemeo:
# Description

One issue: in this PR we typically just hide the `data` keyword; however, there were already some instances in master where (not only the `data` keyword, but also) the name of the inductive type is hidden.  Is this what we want, or do we want to try to be consistent about it?

(Also, I noticed a data type was mentioned in the prose in the `Scripts` module but not typeset in the Agda datatype font. I fixed this in this PR since I don't think a separate PR is warranted for such a minor correction.)

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Interface/HasSubtract.agda
+++ b/src/Interface/HasSubtract.agda
@@ -1,8 +1,8 @@
 {-# OPTIONS --safe --cubical-compatible #-}
 module Interface.HasSubtract where
 
-record HasSubtract (A : Set) : Set where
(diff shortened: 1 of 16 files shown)


williamdemeo:
I believe all the `data` and `where` keywords are now handled as you suggest.

WhatisRT (approved):
LGTM

PR #370, opened 2024-02-15
PDF cleanup 282: hide `module` syntax

williamdemeo:
# Description

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Script.lagda
+++ b/src/Ledger/Script.lagda
@@ -82,30 +82,22 @@ open import Data.List.Relation.Binary.Sublist.Propositional as S
 import Data.Maybe.Relation.Unary.Any as M
 \end{code}
 \begin{code}[hide]
-module _
+data
 \end{code}
 \begin{code}
-  (khs : ℙ KeyHash) (I : Maybe Slot × Maybe Slot)
-\end{code}
-\begin{code}[hide]
-  where
-  data
-\end{code}
-\begin{code}
-
-    evalTimelock : Timelock → Set where
-    evalAll  : All evalTimelock ss
-             → evalTimelock (RequireAllOf ss)
-    evalAny  : Any evalTimelock ss
-             → evalTimelock (RequireAnyOf ss)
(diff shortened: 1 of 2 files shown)


src/Ledger/Script.lagda
```diff
@@ -81,9 +81,19 @@ open import Data.List.Relation.Binary.Sublist.Ext
...
+\end{code}
+\begin{code}
+  (khs : ℙ KeyHash) (I : Maybe Slot × Maybe Slot)
+\end{code}
+\begin{code}[hide]
+  where
+  data
+\end{code}
 \begin{code}
-module _ (khs : ℙ KeyHash) (I : Maybe Slot × Maybe Slot) where
```
WhatisRT, line 85: Since this `module` block contains only one thing, I think it'd be better to just add these as arguments to the datatype.

williamdemeo: (done)

WhatisRT (approved):

src/Ledger/Script.lagda
omelkonian: Oh gosh, the repetition...

Also isn't it more confusing now to have the readers read "indexed constant types"?

WhatisRT: Is there any repetition?

Generally I agree, but I think presenting `evalTimelock` as a datatype is a bad idea in the first place (why isn't it a function?). So when we have a spec that actually wants to present this there's a decent chance that we'll want to refactor this anyway.

omelkonian: The repetition comes from the "constant" indices, `s/evalTimelock/(evalTimelock khs I)`, which is why I used module parameters as the best practice to avoid this, and results in much more readable constructors.

Having said that, I agree that using the "call-graph view" of the function as a datatype might be a bit too much here, I don't know why Ali initially did it this way. I agree you should investigate the alternative sooner rather than later to avoid maintenance overhead.

PR #373, opened 2024-02-21
Vertical vectors in LaTeX: script to postprocess agda generated tex files

williamdemeo:
# Description

When finished, this PR will address issue #7 by providing a Python script `agda2vec.py` which will become part of the make command for docs in `Makefile` so that, whenever the docs are produced (with `nix-shell --command make-docs`), the agda-generated tex files will be processed by `agda2vec.py` and Agda code appearing inside brackets of will be converted to vertical vectors in the following modules: `Chain`, ~`Deleg`~ `Chain`, `Enact`, `Epoch,` `Ratify`, `Ledger`, `NewPP`, ~`Utxo/Properties`~.

# Checklist

- [x] Self-reviewed the diff


WhatisRT:
I think for now, `Enact, Epoch` and `Ledger` should be disabled as they are a bit too broken. If it's only the alignment that breaks that's fine, but enabling it in these modules makes them a bit too hard to read ATM. Otherwise I'm pretty happy!

williamdemeo:
> Oh, for some reason I thought this was already merged. I'll wait for CI to finish and then merge.

I made some changes that broke the CI.  I'll fix the problems and then merge.

PR #377, opened 2024-03-05
Add a collection of (unproven) properties

WhatisRT:
# Description

This PR adds a module containing a bunch of top-level properties that we should prove. Since CI won't pass if we have holes, the actual properties are commented out (and modules that become empty because of this as well, to avoid warnings).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Axiom/Set/Map/Dec.agda
+++ b/src/Axiom/Set/Map/Dec.agda
@@ -16,6 +16,8 @@ open Theory th
 open import Axiom.Set.Rel th using (dom)
 open import Axiom.Set.Map th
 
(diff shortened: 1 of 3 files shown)


jmchapman (commented):
Commenting them out seems susceptible to bit rot. Wouldn't it be better to ensure they are typechecked? E.g., allowing unsolved metas, postulates, giving names to them as types, sticking them in a record etc.

WhatisRT:
These are all things that I'd like to have proven some time in the not-too-distant future, i.e. before we HF into Conway. I'm not sure transforming the code somehow, only to transform it back is worth it.

PR #383, opened 2024-03-12
Deriving tactic for Convertible instances

UlfNorell:
I had to make a small fix to agda-stdlib-meta to work around agda/agda#7187. I put it on our fork here: input-output-hk/stdlib-meta@4fc4b1ed on top of the `v2.0-rc1` tag.


WhatisRT (approved):
LGTM

Issue #367, opened 2024-02-14
Prove that GA deposits match current proposals


WhatisRT:
We should prove that the following property is an invariant, i.e. that if it holds for some state it also holds when we successfully apply a block to that state.

```agda
filterˢ isGADeposit (dom (UTxOState.deposits utxoSt))
    ≡ fromList (map (λ where (id , _) → GovActionDeposit id) govSt)
```

Where `isGADeposit` is a decision procedure whether a deposit was constructed via the `GovActionDeposit` constructor.

jmchapman:
This involves the utxo rule and the gov rule so the proof should live at the level of the ledger rule.

jmchapman:
check what the right notion of the equality is above

jmchapman:
Step 1 should be to encode the statement in Agda and check it makes sense.

PR #384, opened 2024-03-26
Prove GA deposits match current proposals: #367

williamdemeo:
# Description

This PR is to address issue #367.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Ledger/Properties.agda
```diff
@@ -128,145 +133,136 @@ isGADeposit dp = isGADepositᵇ dp ≡ true
 govDepsMatch : LState → Set
 govDepsMatch ⟦ utxoSt , govSt , _ ⟧ˡ =
   filterˢ isGADeposit (dom (utxoSt .deposits ))
-  ≡ᵉ fromList (map (λ (id , _) → GovActionDeposit id) govSt)
+  ≡ᵉ fromList (map f govSt)
```
jmchapman, line 136: why did you change this?

src/Ledger/Ledger/Properties.agda
```diff
@@ -128,145 +133,136 @@ isGADeposit dp = isGADepositᵇ dp ≡ true
...
-  {filterCD : ∀ {c} {pp} → filterˢ isGADeposit (dom ((certDeposit c {pp})ˢ)) ≡ᵉ ∅}
-  {filterCR : (c : DCert) {deps : DepositPurpose ⇀ Coin}
-              → filterˢ isGADeposit (dom ( deps ∣ certRefund c ᶜ ˢ ))
-                ≡ᵉ filterˢ isGADeposit (dom (deps ˢ))}
-  {dom-single : {d : DepositPurpose} {c : Coin} → ❴ d ❵ˢ ≡ dom ❴ d , c ❵ˢ }
-  {ΔgovSt≡ : {txid : TxId} {govSt govSt' : GovState} {txb : TxBody}
-             → fromList (map (λ (id , _) → GovActionDeposit id) govSt')
-               ≡ᵉ fromList (map (λ (id , _) → GovActionDeposit id) govSt)
-                  ∪ fromList (V⊎P∖V→DepPurposes{txid} (txgov txb))}
+  {- 1 -} {dom-single : {d : DepositPurpose} {c : Coin} → ❴ d ❵ˢ ≡ dom ❴ d , c ❵ˢ }
```
jmchapman, line 147: is the intention to leave these ones for this PR?

src/Ledger/Ledger/Properties.agda
```diff
@@ -128,145 +133,136 @@ isGADeposit dp = isGADepositᵇ dp ≡ true
...
+
+  dec-is-inj₂ : ∀ x → Dec (is-inj₂ x)
+  dec-is-inj₂ = λ {(inj₁ _) → no λ (); (inj₂ _) → yes tt}
+
   LEDGER-govDepsMatch : LedgerInvariant _⊢_⇀⦇_,LEDGER⦈_ govDepsMatch
 
   LEDGER-govDepsMatch
-    (LEDGER-I⋯ tx-not-valid (UTXOW-inductive⋯ _ _ _ _ _ (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ ⊢utxo)))
-    aprioriMatch = case ⊢utxo of λ where
+    (LEDGER-I⋯ tx-not-valid (UTXOW-inductive⋯ _ _ _ _ _ (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ UTXOS-sts)))
```
jmchapman, line 167: what's the intention here of doing a pattern match and then a case?

src/Ledger/Ledger/Properties.agda
```diff
@@ -128,145 +133,136 @@ isGADeposit dp = isGADepositᵇ dp ≡ true
...
+    updateGovStates≡ᵉ govSt (inj₂ p ∷ vps) = {!!}
 
-    -- TODO: prove this and remove from list of module assumptions
-    -- ΔgovSt≡ : fromList (map (λ (id , _) → GovActionDeposit id) govSt')
-    --          ≡ᵉ fromList (map (λ (id , _) → GovActionDeposit id) govSt) ∪ fromList (V⊎P∖V→DepPurposes (txgov txb))
-    -- ΔgovSt≡ = {!!}
+    utxo-govst-connex : {votes : List GovVote}{props : List GovProposal}
+      → dom (proposalDepositsΔ props pp txb ˢ)
+        ≡ᵉ fromList (map f (updateGovStatesΔ (map inj₁ votes ++ map inj₂ props)))
+    utxo-govst-connex = {!!}
```
jmchapman, line 240: unsolved 2

src/Ledger/Ledger/Properties.agda
```diff
@@ -128,145 +133,136 @@ isGADeposit dp = isGADepositᵇ dp ≡ true
...
+      begin
+      fromList (map f (updateGovStates govSt (inj₁ v ∷ vps)))
+        ≈⟨ updateGovStates≡ᵉ ((addVote govSt gid voter vote)) vps ⟩
+      fromList (map f (addVote govSt gid voter vote))
+      ∪ fromList (map f (updateGovStatesΔ vps))
+        ≈⟨ ∪-cong (votesDontCount govSt v) ≡ᵉ.refl ⟩
+      fromList (map f govSt)
+      ∪ fromList (map f (updateGovStatesΔ (inj₁ v ∷ vps)))
         ∎
+    updateGovStates≡ᵉ govSt (inj₂ p ∷ vps) = {!!}
```
jmchapman, line 235: unsolved 1

jmchapman (changes requested):
Ok, so I see there are 2 unsolved goals and some cleaning up to do at least at this point

jmchapman:
Hi William, I'm not seeing the latest work on this here

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,535 @@ instance
...
-    LEDGERS-pov (subst (λ s → FreshTxs Γ s l)
-                        (sym $ computational⇒rightUnique Computational-LEDGER x h₁)
-                        h₂) st
+module _ where
+
+  private variable
+    tx : Tx
+    Γ : LEnv
+    s s' : LState
+    l : List Tx
```
williamdemeo, line 96: Moved these inside a module since they clashed with variables in proof of ga deposits invariant below.

src/Ledger/Utxo.lagda
```diff
@@ -130,23 +130,24 @@ instance
...
+certDeposit : DCert → {pp : PParams} → DepositPurpose ⇀ Coin
+certDeposit (delegate c _ _ v)  = ❴ CredentialDeposit c , v                ❵
+certDeposit (regpool c _) {pp}  = ❴ PoolDeposit       c , pp .poolDeposit  ❵
+certDeposit (regdrep c v _)     = ❴ DRepDeposit       c , v                ❵
+certDeposit _                   = ∅
+
+certRefund : DCert → ℙ DepositPurpose
+certRefund (dereg c)      = ❴ CredentialDeposit c ❵
+certRefund (deregdrep c)  = ❴ DRepDeposit c ❵
+certRefund _              = ∅
```
williamdemeo, line 145: Exposing `certDeposit` and `certRefund` to use them in proof of ga deposits invariant.

src/Axiom/Set/Properties.agda
```diff
@@ -189,6 +193,38 @@ card-≡ᵉ (X , lX , lXᵘ , eqX) (Y , lY , lYᵘ , eqY) X≡Y =
...
+
+filter-hom-⊇ : ∀ {P : A → Type} {sp-P : specProperty P} {X Y : Set A}
+               → filter sp-P X ∪ filter sp-P Y ⊆ filter sp-P (X ∪ Y)
+filter-hom-⊇ a∈ = to ∈-filter (case (from ∈-∪ a∈) of λ where
+     (inj₁ v) → proj₁ (from ∈-filter v) , to ∈-∪ (inj₁ (proj₂ (from ∈-filter v)))
+     (inj₂ v) → proj₁ (from ∈-filter v) , to ∈-∪ (inj₂ (proj₂ (from ∈-filter v))) )
+
+filter-hom-∪ : ∀ {P : A → Type} {sp-P : specProperty P} {X Y : Set A}
+               → filter sp-P (X ∪ Y) ≡ᵉ (filter sp-P X) ∪ (filter sp-P Y)
+filter-hom-∪ {A} {P} {sp-P} {X} {Y} = filter-hom-⊆ , filter-hom-⊇
```
williamdemeo, line 226: (used in proof of ga deposits invariant)

src/Axiom/Set/Properties.agda
```diff
@@ -264,14 +300,30 @@ Set-BddSemilattice {A} = record
...
+  ∪-assoc : (X Y Z : Set A) → (X ∪ Y) ∪ Z ≡ᵉ X ∪ (Y ∪ Z)
+  ∪-assoc = ∨-assoc
+
+fromList-∪-singleton : {A : Type ℓ} {x : A} {l : List A} → fromList (x ∷ l) ≡ᵉ ❴ x ❵ ∪ fromList l
+fromList-∪-singleton .proj₁ h with from ∈-fromList h
+... | here refl = ∈-∪⁺ (inj₁ (to ∈-fromList (here refl)))
+... | there q = ∈-∪⁺ (inj₂ (to ∈-fromList q))
+fromList-∪-singleton .proj₂ h with ∈-∪⁻ h
+... | (inj₁ a∈) = to ∈-fromList (here (from ∈-singleton a∈))
+... | (inj₂ a∈) = to ∈-fromList (there (from ∈-fromList a∈))
```
williamdemeo, line 325: (used in proof of ga deposits invariant)

WhatisRT:
There are still two things left to do here, which is to prove the same invariant for `EPOCH` (this should be much easier since `removed` is used to update `govSt` in L100, and also to remove deposits in L108 (and L89) and to lift everything to `CHAIN`, since this is where the invariant is supposed to hold.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+    STS→utxoDeps≡' : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                    → getDeps s' ≡ depUpdate s withIsValid≡ isValid
+
+    STS→utxoDeps≡' {.(⟦ utxoSt' , _ , _ ⟧ˡ)} (LEDGER-V {utxoSt' = utxoSt'}
+      (val , UTXOW-inductive⋯ _ _ _ _ _ _ _ (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ UTXOS-sts) , _ , _)) =
+        trans (⊢utxo-valid UTXOS-sts) (sym (cong (depUpdate s withIsValid≡_) val))
+          where
+          ⊢utxo-valid : record { slot = slot ; pparams = pp } ⊢ LState.utxoSt s ⇀⦇ tx ,UTXOS⦈ utxoSt'
+                        → UTxOState.deposits utxoSt' ≡ updateDeposits pp txb (getDeps s)
+          ⊢utxo-valid (_⊢_⇀⦇_,UTXOS⦈_.Scripts-Yes _) = refl
```
WhatisRT, line 181:
```suggestion
          ⊢utxo-valid (Scripts-Yes _) = refl
```

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+      reflexive (cong dom (STS→utxoDeps≡' utxos-sts))
+
+    STS→utxoDeps≡ : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                    → isValid ≡ true → getDeps s' ≡ updateDeposits pp txb (getDeps s)
+    STS→utxoDeps≡ utxo-sts txvalid = trans (STS→utxoDeps≡' utxo-sts) (cong (depUpdate s withIsValid≡_) txvalid)
+
+    STS→utxoDeps≡ᵉ : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                    → isValid ≡ true → dom (getDeps s') ≡ᵉ dom (updateDeposits pp txb (getDeps s))
+    STS→utxoDeps≡ᵉ utxo-sts txvalid = let open IsEquivalence (≡ᵉ-isEquivalence{DepositPurpose}) in
+      reflexive (cong dom (STS→utxoDeps≡ utxo-sts txvalid))
```
WhatisRT, line 205: This is unused

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+
+    updateVotesOnly : GovState → List (GovVote ⊎ GovProposal) → GovState
+    updateVotesOnly s [] = s
+    updateVotesOnly s (inj₁ v ∷ vps) = updateVotesOnly (updateVote s v) vps
+    updateVotesOnly s (inj₂ _ ∷ vps) = updateVotesOnly s vps
+
+    votesOnly-empty : (vps : List (GovVote ⊎ GovProposal)) → updateVotesOnly [] vps ≡ []
+    votesOnly-empty [] = refl
+    votesOnly-empty (inj₁ v ∷ vps) = votesOnly-empty vps
+    votesOnly-empty (inj₂ p ∷ vps) = votesOnly-empty vps
```
WhatisRT, line 238: Unused

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+      govSt'                             ≡⟨ STS→updateGovSt≡ (txgov txb) (proj₂ (proj₂ (proj₂ x))) ⟩
+      updateGovStates govSt (txgov txb)  ≡˘⟨ cong (govUpdate s withIsValid≡_) (proj₁ x) ⟩
+      govUpdate s withIsValid≡ isValid   ∎
+
+    STS→GovSt≡' {⟦ _ , .govSt , _ ⟧ˡ} (LEDGER-I x) = sym (cong (govUpdate s withIsValid≡_) (proj₁ x))
+
+    STS→GovSt≡ : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                 → isValid ≡ true
+                 → LState.govSt s' ≡ updateGovStates govSt (txgov txb)
+    STS→GovSt≡ utxo-sts txvalid = trans (STS→GovSt≡' utxo-sts) (cong (govUpdate s withIsValid≡_) txvalid)
```
WhatisRT, line 288: You can prove this more easily directly:
```suggestion
    STS→GovSt≡ (LEDGER-V x) refl = STS→updateGovSt≡ (txgov txb) (proj₂ (proj₂ (proj₂ x)))
```
Then, `govUpdate_withIsValid≡_` and `STS→GovSt≡'` become unused.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+
+    ∪++ : {l l' : List DepositPurpose } → fromList l ∪ fromList l' ≡ᵉ fromList (l ++ l')
+    ∪++ {[]} {l'} = ∪-identityˡ (fromList l')
+    ∪++ {x ∷ l} {l'} = let open SetoidReasoning (≡ᵉ-Setoid{DepositPurpose}) in
+      begin
+      fromList (x ∷ l) ∪ fromList l'      ≈⟨ ∪-cong fromList-∪-singleton ≡ᵉ.refl ⟩
+      (❴ x ❵ ∪ fromList l) ∪ fromList l'  ≈⟨ ∪-assoc ❴ x ❵ (fromList l) (fromList l') ⟩
+      ❴ x ❵ ∪ (fromList l ∪ fromList l')  ≈⟨ ∪-cong ≡ᵉ.refl ∪++ ⟩
+      ❴ x ❵ ∪ fromList (l ++ l')          ≈˘⟨ fromList-∪-singleton ⟩
+      fromList (x ∷ (l ++ l'))            ∎
```
WhatisRT, line 530: This should go into `Axiom.Set.Properties`

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+    -- MAIN THEOREM --
+    LEDGER-govDepsMatch : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                          → govDepsMatch s → govDepsMatch s'
+
+    LEDGER-govDepsMatch (LEDGER-I⋯ tx-not-valid
+                          (UTXOW-inductive⋯ _ _ _ _ _ _ _
+                            (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ UTXOS-sts)))
+      aprioriMatch with UTXOS-sts
+    ... | (_⊢_⇀⦇_,UTXOS⦈_.Scripts-Yes u) = contradiction (trans (sym $ proj₂ u) tx-not-valid) (λ ())
+    ... | (_⊢_⇀⦇_,UTXOS⦈_.Scripts-No _) = aprioriMatch
```
WhatisRT, line 583: Like before, assuming you have the pattern defined, you can do this:
```suggestion
    LEDGER-govDepsMatch (LEDGER-I⋯ refl (UTXOW-UTXOS (Scripts-No _))) aprioriMatch = aprioriMatch
```

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+
+    LEDGER-govDepsMatch s'@{⟦ utxoSt' , govSt' , certState' ⟧ˡ}
+      utxosts@(LEDGER-V⋯ tx-valid
+                (UTXOW-inductive⋯ _ _ _ _ _ _ _
+                  (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ UTXOS-sts)) _ GOV-sts) aprioriMatch =
```
WhatisRT, line 588: You're not using any information beyond `LEDGER-V`. This makes me think that you're working too hard to make other things more general. I.e. there's probably a bunch more potential for simplification.
```suggestion
    LEDGER-govDepsMatch s'@{⟦ utxoSt' , govSt' , certState' ⟧ˡ}
      utxosts@(LEDGER-V⋯ tx-valid _ _ _) aprioriMatch =
```

WhatisRT (changes requested):
In addition to the inline comments, there are tons of places where you write `dom (X ˢ)`, but you can usually just write `dom X`.

williamdemeo:
> In addition to the inline comments, there are tons of places where you write `dom (X ˢ)`, but you can usually just write `dom X`.

This wasn't working in a few places so, because I didn't want to bother with it while struggling to get other things to work, I included the superscript... but now that other things are working, I'll remove the superscript where possible.

src/Axiom/Set/Properties.agda
```diff
@@ -196,6 +198,9 @@ mapPartial-∅ : {f : A → Maybe B} → mapPartial f ∅ ≡ᵉ ∅
 mapPartial-∅ {f = f} = ∅-least λ x∈map → case from (∈-mapPartial {f = f}) x∈map of λ where
   (_ , h , _) → ⊥-elim (∉-∅ h)
 
+∈-singleton-pair : ∀ {a x : A} {b y : B} → a ≡ x → b ≡ y → (a , b) ∈ ❴ x , y ❵
+∈-singleton-pair a≡x b≡y = to ∈-singleton $ ×-≡,≡→≡ (a≡x , b≡y)
+
```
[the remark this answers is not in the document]
williamdemeo: Actually, I use `dom-single≡single` in the proof of `allGA-propDepsΔ`, and `dom-single≡single` depends on `∈-dom-singleton-pair` which depends on `∈-dom-singleton-pair`, which depends on `∈-singleton-pair`.  However, if you don't think the latter will be useful, I will delete it and incorporate its proof into the proof of `∈-dom-singleton-pair`.

src/Ledger/Ledger/Properties.agda
williamdemeo: Now that I'm more familiar with the ingredients involved, and have proved a few supporting lemmas about singletons and their domains, I am able to eliminate that assumption.  I will look again at the remaining assumptions and see if I can dispense with those as well; if not, then we'll leave them for a future PR.

src/Ledger/Ledger/Properties.agda
williamdemeo: just to have less duplication of code (less repetition of the patterns)... but much of this will likely change quite a bit when I address Andre's comments re. not needing anything from the expanded UTXOW and UTXO patterns.

src/Ledger/Ledger/Properties.agda
williamdemeo: This is now solved.

src/Ledger/Ledger/Properties.agda
williamdemeo: This is now solved.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+
+
+-- ** Proof that govDepsMatch is a LEDGER invariant.
+
+isGADeposit : DepositPurpose → Set
+isGADeposit dp = isGADepositᵇ dp ≡ true
+  where
+  isGADepositᵇ : DepositPurpose → Bool
+  isGADepositᵇ (GovActionDeposit _) = true
+  isGADepositᵇ _                    = false
```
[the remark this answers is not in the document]
williamdemeo: If I make this change, then I get the error

```
No instance of type (isGADeposit x ⁇) was found in scope.
when checking that (dom (UTxOState.deposits utxoSt)) is a valid
argument to a function of type
⦃ _ : isGADeposit ⁇¹ ⦄ → ℙ DepositPurpose → ℙ DepositPurpose
```

when type-checking the following definition:

```agda
govDepsMatch : LState → Set
govDepsMatch ⟦ utxoSt , govSt , _ ⟧ˡ =
  filterˢ isGADeposit (dom (UTxOState.deposits utxoSt))
  ≡ᵉ fromList (map (λ pr → GovActionDeposit (proj₁ pr)) govSt)
```

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +87,536 @@ instance
...
+
+    depUpdate_withIsValid≡_ : LState → Bool → DepositPurpose ⇀ Coin
+    depUpdate s withIsValid≡ true = updateDeposits pp txb (getDeps s)
+    depUpdate s withIsValid≡ false = getDeps s
+
+    STS→utxoDeps≡' : ∀ {s' : LState} → Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ s'
+                    → getDeps s' ≡ depUpdate s withIsValid≡ isValid
+
+    STS→utxoDeps≡' {.(⟦ utxoSt' , _ , _ ⟧ˡ)} (LEDGER-V {utxoSt' = utxoSt'}
+      (val , UTXOW-inductive⋯ _ _ _ _ _ _ _ (UTXO-inductive⋯ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ UTXOS-sts) , _ , _)) =
```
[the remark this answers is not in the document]
williamdemeo: Nice!

src/Ledger/Ledger/Properties.agda
williamdemeo: No, it's used in the first line of the proof of the main theorem, at or around line 566:

```agda
      begin
      filterˢ isGADeposit (dom utxoDeps')
        ≈⟨ filter-pres-≡ᵉ (STS→utxoDeps≡ᵉ utxosts tx-valid) ⟩
        ...
```

src/Ledger/Ledger/Properties.agda
williamdemeo: Nice! 👍🏼

src/Ledger/Ledger/Properties.agda
williamdemeo: Very nice 👍🏼

src/Ledger/Ledger/Properties.agda
williamdemeo: However, I have since in-lined the proof of `STS→utxoDeps≡ᵉ` in the one place where it was used... so I will now delete `STS→utxoDeps≡ᵉ`.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +97,611 @@ instance
...
+           (Γ : LEnv) (let open LEnv Γ renaming (pparams to pp))
+           (eΓ : NewEpochEnv)(eps : EpochState)
+    where
+    open EpochState eps hiding (es)
+    open LState ls
+    open GovActionState
+    open RatifyState fut using (removed) renaming (es to esW)
+
+    EPOCH-govDepsMatch :
+      {assumption/bug? : mapˢ (GovActionDeposit ∘ proj₁) removed ⊆ mapˢ proj₁ ((UTxOState.deposits utxoSt)ˢ)}
```
WhatisRT, line 577: This isn't a bug, `removed` should have this property if it's been computed by `RATIFY` in the previous `EPOCH` transition. But of course it might be false when you just make up an arbitrary state. So I think we can think about how to correctly state and prove this in the future.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +97,611 @@ instance
...
+    open NewEpochState newEpochState
+
+    ls₀ : LState
+    ls₀ = EpochState.ls epochState
+
+    CHAIN-govDepsMatch : {b : Block} {cs' : ChainState}
+      → _ ⊢ cs ⇀⦇ b ,CHAIN⦈ cs'
+      → govDepsMatch ls₀ → govDepsMatch (EpochState.ls (NewEpochState.epochState (ChainState.newEpochState cs')))
+
+    CHAIN-govDepsMatch (CHAIN _ y) = RTC-preserves-inv (λ {c}{s}{sig} → LEDGER-govDepsMatch sig c s) y
```
WhatisRT, line 707: The fact that this proof doesn't need `EPOCH-govDepsMatch` means that `CHAIN` is incorrect. We actually throw the result of `NEWEPOCH` away. Here's a fixed version:
```
  CHAIN :
    let open Block b; open ChainState s; open NewEpochState nes
        open EpochState epochState; open EnactState es
    in
       record { stakeDistrs = calculateStakeDistrs (EpochState.ls (NewEpochState.epochState newEpochState)) }
         ⊢ newEpochState ⇀⦇ epoch slot ,NEWEPOCH⦈ nes
    →  ⟦ slot , constitution .proj₁ .proj₂ , pparams .proj₁ , es ⟧ˡᵉ
         ⊢ ls ⇀⦇ ts ,LEDGERS⦈ ls'
    _ ⊢ s ⇀⦇ b ,CHAIN⦈
        record s { newEpochState = record nes { epochState = record epochState { ls = ls'} } }
```

src/Ledger/Set/Theory.agda
```diff
@@ -126,3 +126,9 @@ indexedSum' f s = indexedSum ⦃ fromCommMonoid' it ⦄ f (s ᶠˢ)
 
 syntax indexedSumᵛ' (λ a → x) m = ∑[ a ← m ] x
 syntax indexedSum'  (λ a → x) m = ∑ˢ[ a ← m ] x
+
+module _ ⦃ _ : DecEq A ⦄ (P : A → Set) ⦃ _ : P ⁇¹ ⦄ {m : A ⇀ B} {X : ℙ A} where
+
+  ∈-filter-resᶜ-dom⁺ : ∀ {a} → P a × a ∉ X × ∃[ b ] (a , b) ∈ (m ˢ) → a ∈ filterˢ P (dom (m ∣ X ᶜ))
+  ∈-filter-resᶜ-dom⁺ (Pa , a∉X , bab∈m) = to ∈-filter (Pa , (∈-resᶜ-dom⁺ (a∉X , bab∈m)))
```
WhatisRT, line 133: I don't think this should be a separate lemma. This is essentially just a combination of the main properties of `filter` and restriction and there are `n^2` many such lemmas.

src/Axiom/Set/Rel.agda
```diff
@@ -88,6 +90,54 @@ dom∈ {R = R} {a} =
...
+    a ∈ dom (R ∪ R')                           ∼⟨ dom∈ ⟩
+    (∃[ b ] (a , b) ∈ R ∪ R')                  ∼⟨ ∃-cong′ (R.SK-sym ∈-∪) ⟩
+    (∃[ b ] ((a , b) ∈ R ⊎ (a , b) ∈ R'))      ∼⟨ mk⇔ (λ {(b , inj₁ p) → inj₁ (b , p);
+                                                          (b , inj₂ p) → inj₂ (b , p)})
+                                                      (λ {(inj₁ (b , pf)) → (b , inj₁ pf);
+                                                          (inj₂ (b , pf)) → (b , inj₂ pf)}) ⟩
+    (∃[ b ] (a , b) ∈ R ⊎ ∃[ b ] (a , b) ∈ R') ∼⟨ mk⇔ (λ {(inj₁ bpf) → inj₁ (from dom∈ bpf);
+                                                          (inj₂ bpf) → inj₂ (from dom∈ bpf)})
+                                                      (λ {(inj₁ pf) → inj₁ (to dom∈ pf);
+                                                          (inj₂ pf) → inj₂ (to dom∈ pf)}) ⟩
```
WhatisRT, line 131: You can use `∃-cong′` to simplify the rest of these steps as well.

src/Ledger/Ledger/Properties.agda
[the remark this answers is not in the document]
WhatisRT: Ah, yeah, that makes sense. I guess that's a neat hack to avoid this issue.

src/Ledger/Ledger/Properties.agda
williamdemeo: Okay, I'll change the name for now, to something like, ratified-removed.

src/Ledger/Ledger/Properties.agda
williamdemeo: Okay, thanks.  Apparently, I didn't fully understand your earlier explanation about why CHAIN must depend on EPOCH.  Thanks for clarifying.  Hopefully this is enough for me to go on... but I'll let you know if I get stuck trying to prove it.

src/Ledger/Set/Theory.agda
williamdemeo: I thought it would be useful to not have to think about how to prove this again in the future, and I thought this was the sort of thing James was suggesting in our last meeting... but maybe not.  Also, `n^2` many such lemmas? 🤔

src/Axiom/Set/Rel.agda
williamdemeo: I tried to do so last week, but couldn't get anything simpler than what I had to type check... but I'll try again.

src/Axiom/Set/Map/Dec.agda
```diff
@@ -60,3 +61,37 @@ module Lookupᵐᵈ (sp-∈ : spec-∈ A) where
...
+
+    module _ {m m' : Map A V} where
+
+      dom-∪⁺⊆∪ : dom ((m ∪⁺ m') ˢ) ⊆ dom (m ˢ) ∪ dom (m' ˢ)
+      dom-∪⁺⊆∪ p = subst (_∈ dom (m ˢ) ∪ dom (m' ˢ))
+                         (Prelude.sym $ proj₁ (×-≡,≡←≡ $ proj₁ (proj₂ $ ∈-dom-∪⁺ p)))
+                         (proj₂ $ proj₁ (∈-dom-∪⁺ p))
+        where
+        f : Σ A (λ a → a ∈ dom (m ˢ) ∪ dom (m' ˢ)) → A × V
+        f (a , a∈) = a , (fold id id _◇_)(unionThese m m' a a∈)
```
WhatisRT, line 73: This `f` appears twice. How about you lift it out of this `where` block into a `private` block of the parent module and give it a slightly more descriptive name?

src/Axiom/Set/Properties.agda
```diff
@@ -205,8 +220,36 @@ card-≡ᵉ (X , lX , lXᵘ , eqX) (Y , lY , lYᵘ , eqY) X≡Y =
...
-filter-⊆ : ∀ {P} {sp-P : specProperty P} → filter sp-P X ⊆ X
-filter-⊆ = proj₂ ∘′ ∈⇔P
+module _ {P : A → Type} {sp-P : specProperty P} where
+
+  filter-⊆ : filter sp-P X ⊆ X
+  filter-⊆ = proj₂ ∘′ ∈⇔P
+
+  filter-pres-⊆ : X ⊆ Y → filter sp-P X ⊆ filter sp-P Y
+  filter-pres-⊆ xy a∈ = let Pa∈ = from ∈-filter a∈ in
+    to ∈-filter (map₂ (λ u → xy u) Pa∈)
```
WhatisRT, line 230:
```suggestion
    to ∈-filter (map₂ xy Pa∈)
```

src/Axiom/Set/Rel.agda
WhatisRT: Here you go:
```
    (∃[ b ] ((a , b) ∈ R ⊎ (a , b) ∈ R'))      ↔⟨ ∃-distrib-⊎' ⟩
    (∃[ b ] (a , b) ∈ R ⊎ ∃[ b ] (a , b) ∈ R') ∼⟨ R.SK-sym (dom∈ ⊎-cong dom∈) ⟩
```
Only needs the correct imports

WhatisRT (changes requested):
There are some more simplifications than I have pointed out here, but it's too late to finish this today. Basically, you should be able to show something like this:
```agda
    dpMap-update : (ps : List GovProposal) (k : ℕ)
      → dpMap (updateGovStates (map inj₂ ps) k []) ≡ map (λ i → GovActionDeposit (txid , i)) (upTo (length ps))
```
Then you don't need to deal with permutations at all, since the two permuted lists have the same length.

src/Axiom/Set/Properties.agda
```diff
@@ -64,9 +68,28 @@ module _ {l : List A} {a} where
...
-  ∈-fromList⁺ : ∀ {l : List A} {a} → a ∈ˡ l → a ∈ fromList l
+  ∈-fromList⁺ : a ∈ˡ l → a ∈ fromList l
   ∈-fromList⁺ = to ∈-fromList
 
+module _ {f : A → B} {l : List A} {b : B} where
+  ∈-fromList-map⁺ : b ∈ fromList (Data.List.map f l) → ∃[ a ] a ∈ˡ l × b ≡ f a
+  ∈-fromList-map⁺ = (∈-map⁻ f) ∘ ∈-fromList⁻
+
+  ∈-fromList-map⁻ : ∃[ x ] x ∈ˡ l × b ≡ f x → b ∈ fromList (Data.List.map f l)
+  ∈-fromList-map⁻ (x , x∈ , b≡) = ∈-fromList⁺ (∈-map⁺ (x , b≡ , x∈))
```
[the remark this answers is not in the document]
williamdemeo: Okay, I'll remove them again... I reproved them when resuming work on this PR because they were useful when proving things that required keeping a lot of logic in my head at once.  Without these abstractions, the slightly higher complexity somehow made it significantly more difficult to get the proof right.

src/Axiom/Set/Properties.agda
```diff
@@ -283,14 +337,63 @@ Set-BddSemilattice {A} = record
...
+      fromList (mapˡ f lr) ∪ fromList (mapˡ f ll)
+        ≈⟨ ∪-fromList-++ (mapˡ f lr) (mapˡ f ll) ⟩
+      fromList (mapˡ f lr ++ mapˡ f ll)
+        ≈˘⟨ ≡ᵉ.reflexive (cong fromList (map-++ f lr ll)) ⟩
+      fromList (mapˡ f (lr ++ ll))
+        ∎
+    where
+    module ≡ᵉ = IsEquivalence (≡ᵉ-isEquivalence)
+    open import Relation.Binary.Reasoning.Setoid ≡ᵉ-Setoid
+
```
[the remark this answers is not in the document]
williamdemeo: removed

src/Data/List/Ext.agda
```diff
@@ -35,3 +40,23 @@ insert x (y ∷ ys) = (x ∷ y ∷ ys) ∷ map (y ∷_) (insert x ys)
...
+∈-map⁻ {l = _ ∷ _} (here _≡_.refl) = _ , _≡_.refl , here _≡_.refl
+∈-map⁻ {l = _ ∷ _} (there h) = proj₁ (∈-map⁻ h) , proj₁ (proj₂ (∈-map⁻ h)) , there (proj₂ (proj₂ (∈-map⁻ h)))
+
+∈-map⁺ : {f : A → B} {b : B} {l : List A} → ∃[ a ](b ≡ f a × a ∈ l) → b ∈ map f l
+∈-map⁺ (._ , b≡fa , here _≡_.refl) = here b≡fa
+∈-map⁺ (a , b≡fa , there a∈l) = there (∈-map⁺ (a , b≡fa , a∈l))
+
+∈-map : {f : A → B} {b : B} {l : List A} → (∃[ a ] (b ≡ f a × a ∈ l)) ⇔ b ∈ map f l
+∈-map {l = []} = mk⇔ (λ ()) (λ ())
+∈-map {l = _ ∷ _} = mk⇔ ∈-map⁺ ∈-map⁻
```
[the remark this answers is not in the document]
williamdemeo: They're not quite the same.  The first two differ in signatures and `∈-map` inhabits `Equivalence`, whereas `map-∈↔` (in `Data.List.Membership.Propositional.Properties`) inhabits `Inverse`. I can make adjustments and replace `∈-map⁻` and `∈-map⁺` with those from `Data.List.Membership.Propositional.Properties`, but I think we should `∈-map` around.  (Otherwise, I'll need to figure out how to convert from `Inverse` to `Equivalence`.)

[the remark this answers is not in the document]
WhatisRT: We're already doing the conversion from `Inverse` to `Equivalence` in lots of places. It's done with `↔⇒` from `Function.Related.Propositional`, which is usually available as `R.↔⇒`.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +102,579 @@ instance
...
+        ... | ((rwa , dp , c) , a≡dp , rwa-dp-c∈) with (from ∈-concatMapˢ rwa-dp-c∈)
+        ... | ((gaid , gast) , gaid-gast-∈-removed , rwa-dp-c-∈-map) with (from ∈-map rwa-dp-c-∈-map)
+        ... | ((dp' , _) , q≡ , q∈) = to ∈-map ((gaid , gast) , a≡GA , gaid-gast-∈-removed)
+          where
+          a≡GA : a ≡ GovActionDeposit gaid
+          a≡GA = let open ≡-Reasoning in begin
+            a                       ≡⟨ a≡dp ⟩
+            dp                      ≡⟨ cong (proj₁ ∘ proj₂) q≡ ⟩
+            dp'                     ≡⟨ cong proj₁ $ proj₂ (res-singleton''{m = utxoDeps} $ q∈) ⟩
+            GovActionDeposit gaid   ∎
```
WhatisRT, line 620: Can you not do this proof without splitting up the two inclusions? I see a lot of things like `from ∈-map x`, which reads to me like what you should actually be doing would be to use `∈-map` to rewrite your expression. It would be something like this:
```agda
... = from ≡ᵉ⇔≡ᵉ' λ a → let open R.EquationalReasoning in
        ...
          ∼⟨ ∈-map ⟩
        ...
```
And even if you cannot fully complete the proof this way, you can still eliminate a bunch of the repetition.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +102,579 @@ instance
...
+        ... | (aid×st , aid×st∈ , refl) with ∈-filter⁻ (λ u → ¿ P u ¿) aid×st∈
+        ... | (aid×st∈govSt , Paid×st) =
+          to ∈-filter (refl , ∈-resᶜ-dom⁺ (a∉χ' , (to dom∈ $ proj₂ (from ∈-filter a∈))))
+           where
+           a∈ : a ∈ filterˢ isGADeposit (dom utxoDeps)
+           a∈ = snd $ to ∈-fromList $ to ∈ˡ-map (aid×st , refl , aid×st∈govSt)
+
+           a∉χ' : a ∉ χ'
+           a∉χ' a∈χ' with ∈-map⁻' a∈χ'
+           ... | q , refl , q∈rem = Paid×st (to ∈-map (q , refl , q∈rem))
```
WhatisRT, line 650: The same here, many of those rewriting steps are bidirectional.

src/Axiom/Set/Map/Dec.agda
```diff
@@ -60,3 +61,27 @@ module Lookupᵐᵈ (sp-∈ : spec-∈ A) where
 
     aggregate₊ : FinSet (A × V) → Map A V
     aggregate₊ (_ , l , _) = foldl (λ m x → m ∪⁺ ❴ x ❵ᵐ) ∅ᵐ l
+
+    module _ {m m' : Map A V} where
+      dom→× : Σ A (λ a → a ∈ dom (m ˢ) ∪ dom (m' ˢ)) → A × V
```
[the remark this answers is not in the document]
williamdemeo: Okay, I'll rename it `∪-dom-lookup`.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +102,579 @@ instance
...
+        fromList (dpMap govSt ++ dpMap (updateGovStates (map inj₂ txprop ++ map inj₁ txvote) 0 []))
+          ≈˘⟨ ≡ᵉ.reflexive (cong fromList (map-++ (λ (id , _) → GovActionDeposit id) govSt (updateGovStates (txgov txb) 0 []))) ⟩
+        fromList (dpMap (govSt ++ updateGovStates (txgov txb) 0 []))
+          ≈˘⟨ ≡ᵉ.reflexive (cong fromList (updateGovStates≡ (txgov txb) 0)) ⟩
+        fromList (dpMap (updateGovStates (txgov txb) 0 govSt ))
+          ≈˘⟨ ≡ᵉ.reflexive (cong (fromList ∘ dpMap ) (STS→GovSt≡ utxosts tx-valid)) ⟩
+        fromList (dpMap govSt')
+          ∎
+        where
+        utxo-govst-connex : (props : List GovProposal)
```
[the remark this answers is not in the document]
WhatisRT: Yeah, it's a bit difficult. It's more about how specific and big something is than whether it's used by just that thing. So if it's just used in one place, there's a higher probability that it's also specific to that place, but it might not be. And this lemma looks like it might be useful to prove other things.

src/Ledger/Ledger/Properties.agda
williamdemeo: This was one of the harder, more technical proofs in the PR.  It is quite delicate/brittle and already required a substantial amount of time/effort to get it right and then simplified it to where it is now.  I can try to make some minor improvements, but doing both inclusions simultaneously, while maybe possible, will be challenging for me and I'm not sure it's worth the additional time/effort.

src/Data/List/Ext.agda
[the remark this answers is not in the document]
williamdemeo: By "lots" I think you mean "two" 😉 ...but I'll give it a try.

The only occurrences of they symbol `↔⇒` I could find are shown below.

```
File: Axiom/Set/List.agda
39:61:    (∃[ a ] b ≡ f a × a ∈ˡ X) ∼⟨ ∃-cong′ (I.↔⇒⇔ (×-comm _ _)) ⟩
40:48:    (∃[ a ] a ∈ˡ X × b ≡ f a) ⤖⟨ I.↔⇒⤖ (map-∈↔ f) ⟩

File: Ledger/Gov.lagda
28:51:open import Function.Related.Propositional using (↔⇒)
163:46:∃?-connecting-subperm L = from (map′⇔ (↔⇒ Any↔)) (any?-connecting-subperm L)
4 matches
```

src/Axiom/Set/Map.agda
```diff
@@ -178,7 +178,7 @@ mapˡ-uniq {inj = inj} uniq = λ h h' → case ∈⇔P h ,′ ∈⇔P h' of λ w
   (((_ , b) , refl , Ha) , ((_ , b') , eqb , Hb)) → uniq Ha
     $ subst _ ( sym
               $ ×-≡,≡→≡
-              $ map₁ (inj (from dom∈ (b , Ha)) (from dom∈ (b' , Hb)))
+              $ map₁ (inj (to dom∈ (b , Ha)) (to dom∈ (b' , Hb)))
```
williamdemeo, line 181: Reverse order of equivalence in definition of `dom∈` to be more consistent with other definitions and so that it's more natural in the sense that `to dom∈` takes a decomposition *to* a dom inclusion.

src/Axiom/Set/Rel.agda
```diff
@@ -80,28 +83,69 @@ mapˡ f R = map (Data.Product.map₁ f) R
 mapʳ : (B → B') → Rel A B → Rel A B'
 mapʳ f R = map (Data.Product.map₂ f) R
 
-dom∈ : ∀ {a} → a ∈ dom R ⇔ (∃[ b ] (a , b) ∈ R)
+dom∈ : ∀ {a} → (∃[ b ] (a , b) ∈ R) ⇔ a ∈ dom R
```
williamdemeo, line 86: Reverse order of equivalence in definition of `dom∈` to be more consistent with other definitions and so that it's more natural in the sense that `to dom∈` takes a decomposition *to* a dom inclusion.

src/Ledger/Ledger/Properties.agda
williamdemeo: I did my best to make this a bidirectional proof and separated out the one step I couldn't make bidirectional into a separate lemma, called `map-filter-decomp` (which now appears just above the `main-invariance-lemma`).

src/Ledger/Ledger/Properties.agda
williamdemeo: I simplified the proof as much as I could... still can't see how to make it bidirectional.

src/Data/Product/Properties/Ext.agda
```diff
@@ -45,3 +45,15 @@ private variable
   [ Prod.map₂ inj₁ , Prod.map₂ inj₂ ]′
   [ (λ _ → P.refl) , (λ _ → P.refl) ]
   (uncurry λ _ → [ (λ _ → P.refl) , (λ _ → P.refl) ])
+
+×-⇔ : {A₁ : Set a₁} {A₂ : Set a₂} {B₁ : Set b₁} {B₂ : Set b₂}
+ → A₁ ⇔ A₂ → B₁ ⇔ B₂ → (A₁ × B₁) ⇔ (A₂ × B₂)
+×-⇔ {A₁} {A₂} {B₁} {B₂} A⇔A' B⇔B' =
+  mk⇔ (λ x → Equivalence.to A⇔A' (proj₁ x) , Equivalence.to B⇔B' (proj₂ x) )
+      (λ x → Equivalence.from A⇔A' (proj₁ x) , Equivalence.from B⇔B' (proj₂ x))
```
WhatisRT, line 53: This already exists as `_×-cong_`

src/Axiom/Set/Properties.agda
```diff
@@ -76,7 +75,7 @@ module _ {f : A → B} {l : List A} {b : B} where
   ∈-fromList-map⁺ = (∈-map⁻ f) ∘ ∈-fromList⁻
 
   ∈-fromList-map⁻ : ∃[ x ] x ∈ˡ l × b ≡ f x → b ∈ fromList (Data.List.map f l)
-  ∈-fromList-map⁻ (x , x∈ , b≡) = ∈-fromList⁺ (∈-map⁺ (x , b≡ , x∈))
+  ∈-fromList-map⁻ (x , <″-offset x∈) = ∈-fromList⁺ (∈-map⁺ f x∈)
```
WhatisRT, line 78: This `<″-offset` stuff should be replaced

src/Data/List/Ext.agda
```diff
@@ -35,3 +42,30 @@ insert x (y ∷ ys) = (x ∷ y ∷ ys) ∷ map (y ∷_) (insert x ys)
...
+  -- ∈ˡ-map-filter⁻ : b ∈ map f (filter P? l) → (∃[ a ] b ≡ f a × a ∈ l × P a)
+  -- ∈ˡ-map-filter⁻ h with ∈-map⁻ f h
+  -- ... | a , a∈X , _≡_.refl = a , _≡_.refl , ∈-filter⁻ P? a∈X
+
+  -- ∈ˡ-map-filter⁺ : (∃[ a ] b ≡ f a × a ∈ l × P a) → b ∈ map f (filter P? l)
+  -- ∈ˡ-map-filter⁺ (a , _≡_.refl , a∈l , Pa) = ∈-map⁺ f (∈-filter⁺ P? a∈l Pa)
+
+  -- ∈ˡ-map-filter : (∃[ a ] b ≡ f a × a ∈ l × P a) ⇔ b ∈ map f (filter P? l)
+  -- ∈ˡ-map-filter = mk⇔ ∈ˡ-map-filter⁺ ∈ˡ-map-filter⁻
+
```
WhatisRT, line 62: Please remove those comments

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +105,551 @@ instance
...
+        a ∈ (filterˢ isGADeposit (dom utxoDeps')) ∼⟨ R.SK-sym ∈-filter ⟩
+        ((isGADeposit a) × a ∈ (dom utxoDeps'))  ∼⟨ ×-⇔ R.K-refl ∈-resᶜ-dom ⟩
+        ((isGADeposit a) × a ∉ χ' × ∃[ q ] (a , q) ∈ utxoDeps) ∼⟨ ×-⇔-swap ⟩
+        (a ∉ χ' × (isGADeposit a) × ∃[ q ] (a , q) ∈ utxoDeps) ∼⟨ ×-⇔ R.K-refl (×-⇔ R.K-refl dom∈) ⟩
+        (a ∉ χ' × isGADeposit a × a ∈ dom utxoDeps) ∼⟨ ×-⇔ R.K-refl ∈-filter ⟩
+        (a ∉ χ' × a ∈ filterˢ isGADeposit (dom utxoDeps)) ∼⟨ ×-⇔ R.K-refl (HYP a) ⟩
+        (a ∉ χ' × a ∈ fromList (map (GovActionDeposit ∘ proj₁) govSt)) ∼⟨ ×-⇔ R.K-refl (R.SK-sym ∈-fromList) ⟩
+        (a ∉ χ' × a ∈ˡ map (GovActionDeposit ∘ proj₁) govSt) ∼⟨ map-filter-decomp a ⟩
+        (a ∈ˡ map (GovActionDeposit ∘ proj₁)(filterᵇ Pᵇ govSt)) ∼⟨ ∈-fromList ⟩
+        a ∈ (fromList (map (GovActionDeposit ∘ proj₁)(filterᵇ Pᵇ govSt))) ∎
```
WhatisRT, line 625: Please align the reasoning steps, or put them in new lines.

WhatisRT (commented):
There are a bunch of `<″-offset` everywhere. Please remove those, they are confusing & will be removed in stdlib 2.0.

src/Ledger/Ledger/Properties.agda
```diff
@@ -82,29 +102,579 @@ instance
...
+      fromList (dpMap (propUpdate (propUpdate govSt y k) x (suc k))) ∪ fromList (dpMap (updateGovStates (map inj₂ ys) (suc (suc k)) []))
+        ≈⟨ ∪-fromList-++ (dpMap (propUpdate (propUpdate govSt y k) x (suc k))) (dpMap (updateGovStates (map inj₂ ys) (suc (suc k)) [])) ⟩
+      fromList ((dpMap (propUpdate (propUpdate govSt y k) x (suc k))) ++ (dpMap (updateGovStates (map inj₂ ys) (suc (suc k)) [])))
+        ≈˘⟨ ≡ᵉ.reflexive (cong fromList (map-++ (λ (id , _) → GovActionDeposit id)
+            (propUpdate (propUpdate govSt y k) x (suc k)) (updateGovStates (map inj₂ ys) (suc (suc k)) []))) ⟩
+      fromList (dpMap (propUpdate (propUpdate govSt y k) x (suc k) ++ (updateGovStates (map inj₂ ys) (suc (suc k)) [])))
+        ≈˘⟨ ≡ᵉ.reflexive (cong ((fromList ∘ dpMap)) (updateProps-decomp ys)) ⟩
+      fromList (dpMap (updateGovStates (map inj₂ ys) _ (propUpdate (propUpdate govSt y k) x (suc k))))
+        ∎
+
```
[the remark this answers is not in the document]
williamdemeo: Okay, thanks.  I managed to prove it without function extensionality.  The proof now relies on the following lemma, which I couldn't find in the standard library.  Do you know where it is?
```agda
∷-injective⁻ : {a : Level} {A : Set a} {xs ys : List A} {x y : A} → x ≡ y × xs ≡ ys → (x ∷ xs) ≡ (y ∷ ys)
∷-injective⁻ (refl , refl) = refl
```

williamdemeo:
@WhatisRT I think we're finally good to go... although, I looked for the new cosmetic change requests you mentioned earlier, but couldn't find any new ones.  Have I missed anything?

src/Ledger/Ledger/Properties.agda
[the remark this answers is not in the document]
williamdemeo: advice from slack: use `_cong₂_`

WhatisRT (approved):
LGTM

jmchapman (approved):
LGTM

PR #385, opened 2024-03-26
Add initial version of script verification

Ali-Hill:
# Description

Adds a concrete Agda implementation of the ledger that supports simple scripts that are represented as boolean functions. Scripts can then be evaluated as true using the evalScripts function.

# Checklist

- [ ] Self-reviewed the diff


src/ScriptVerification/HelloWorld.agda
```diff
@@ -0,0 +1,128 @@
...
+
+  -- need to check that the state is non-empty otherwise evalScripts will always return true
+  _ : notEmpty succeedState ≡ ⊤
+  _ = refl
+
+  _ : succeedExample ≡ true
+  _ = refl
+
+  _ : notEmpty failState ≡ ⊤
+  _ = refl
```
WhatisRT, line 125:
```suggestion
  _ = refl
```
Unicode whitespace

WhatisRT (commented):
There's a bunch of stuff that's been commented out. I think there's no reason for keeping it around, so it'd be better to just remove it.

I also think the examples should go just a little bit further. For example, I put this in `HelloWorld.agda`:
```
  succeedExample' : Bool
  succeedExample' = case compute Computational-UTXOW initEnv ⟦ initState , 0 , ∅ , 0 ⟧ᵘ succeedTx of λ where
    (success x) → true
    (failure x) → false
```
If I normalize (with `C-u C-n` to not get stuck on `abstract/opaque`), I get `false`, as well as if I use `Computational-UTXO`, but it works for `Computational-UTXOS`. I think it'd be nice to get the example working until all of these work. The same applies to `SucceedIfNumber`.

You can actually get error messages from these, by normalizing something like `compute Computational-UTXO initEnv ⟦ initState , 0 , ∅ , 0 ⟧ᵘ succeedTx`, which gives me `failure "¬(minfee (UTxOEnv.pparams Γ) tx ≤ txfee (Tx.body tx))"`. It doesn't actually tell you how much is missing or anything, but at least you know where to look.

WhatisRT (approved):
LGTM, I removed a few leftover comments

PR #388, opened 2024-04-09
Added a helper script for bumping the executable spec repo

Soupstraw:
# Description

This PR adds a script that builds the executable script Haskell library and copies all the files to the `cardano-ledger-executable-spec` repo. This will partially automate the bumping the executable spec for conformance testing.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/scripts/bump-executable-spec.sh
+++ b/scripts/bump-executable-spec.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -e
+
+EXEC_SPEC_REPO_DIR=$1
+if [ ! $EXEC_SPEC_REPO_DIR ]; then
+  echo "No repository path provided."
+  echo "Usage: ./bump-executable-spec.sh EXEC_SPEC_REPO_DIR [LEDGER_REPO_DIR]"
+  exit 1
+fi
+
+AGDA_COMMIT=$(git rev-parse HEAD)
+echo "Building executable spec.."
+nix-build -A ledger.hsSrc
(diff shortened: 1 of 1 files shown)


WhatisRT:
Ah, I've already done this, see the `haskell-package` branch/bottom of `.github/workflows/build.yml`. But maybe your script is prettier than what I've done.

scripts/bump-executable-spec.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "Entering executable spec repo.."
+chmod 755 -R $EXEC_SPEC_REPO_DIR
+pushd $EXEC_SPEC_REPO_DIR
+
+echo "Removing old files.."
+FILES_TO_RM=$(git ls-tree -r main --name-only | grep -vE "flake.lock|flake.nix|CHANGELOG.md|.gitignore")
+for f in $FILES_TO_RM; do
+  rm -rf $f
+done
```
WhatisRT, line 24: Is the purpose of this to not have anything but the Haskell code on the branch? Running `rm -rf` with a variable argument always makes me nervous.

I think the better way might be to have an empty branch from the beginning, then check that out and only add the new files.

Soupstraw: It's the opposite actually, the grep command is inverted.

WhatisRT:
Ok, then we'll have some duplicate scripts for now, but maybe it's fine.

WhatisRT (approved):

PR #390, opened 2024-04-11
Add `genError` macro that automatically generates error messages

WhatisRT:
# Description

This can be used to ensure the error messages generated by `Computational` instances are always correct. However, these messages are rather ugly.

# Checklist

- [x] Self-reviewed the diff


Soupstraw (approved):
I don't macros well enough to give feedback about `GenError.agda`, but LGTM otherwise

PR #393, opened 2024-04-15
Require reference inputs to be disjoint from regular inputs

### LINKED ISSUE #354: Disallow shared inputs between regular & reference
It would be nice if it didn't matter how to compute things in case of duplication between the two. The sane way is to prevent that duplication, which makes sense as a check in UTXO.

Since we don't have reference inputs yet in the spec, we can't work on this.

WhatisRT:
# Description

Closes #354. This will also allow reference inputs in conformance tests.

# Checklist

- [x] Self-reviewed the diff


Soupstraw (approved):
LGTM

PR #401, opened 2024-04-25
Remove PPUP imports from Conway

### LINKED ISSUE #117: Retire PPUp
Since we're currently working on the most recent spec which shouldn't include PPUp anymore we shouldn't depend on it anymore. However, we should keep it around since we may eventually want to model older eras. Maybe we can put it in a folder like `src/Outdated` for now.

WhatisRT:
# Description

Closes #117.

# Checklist

- [x] Self-reviewed the diff


williamdemeo (approved):
looks good

PR #402, opened 2024-04-26
Replace `allEnactable` with `hasParent` in `GOV`

WhatisRT:
# Description

This reduces runtime from O(n!) to O(n). We should then prove that `allEnactable` is an invariant of `GOV`.

# Checklist

- [x] Self-reviewed the diff


Soupstraw (approved):
LGTM

PR #420, opened 2024-05-06
`CERT` Conformance

Soupstraw:
# Description

Makes some changes for conformance testing the `CERT` rule:
 * Added necessary `LedgerTypes` and `Convertible` instances
 * Added a newtype wrapper for `HSMap` to make it easier to work with maps in conformance testing
 * ~~Removed `CommutativeMonoid` requirement from `_∪⁺_` and `aggregate₊` (I ended up not actually needing this, but I think it still makes sense)~~
 * Replaced some hard-coded error messages with `genErrors`
 * Added `InjectError` to simplify error message types

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Interface/ComputationalRelation.agda
+++ b/src/Interface/ComputationalRelation.agda
@@ -168,20 +168,34 @@ Computational⇒Dec' ⦃ comp = comp ⦄ = Computational⇒Dec comp
 
 open Computational ⦃...⦄
 
+record InjectError Err₁ Err₂ : Set where
+  field injectError : Err₁ → Err₂
+
+open InjectError
+
 instance
+  InjectError-⊥ : InjectError ⊥ Err
+  InjectError-⊥ = λ where
+    .injectError ()
+
+  InjectError-Id : InjectError Err Err
+  InjectError-Id = λ where
+    .injectError → id
+
   Computational-Id : {C S : Set} → Computational (IdSTS {C} {S}) ⊥
   Computational-Id .computeProof _ s _ = success (s , Id-nop)
   Computational-Id .completeness _ _ _ _ Id-nop = refl
 
 module _ {BSTS : C → S → ⊤ → S → Set} ⦃ _ : Computational BSTS Err₁ ⦄ where
-  module _ {STS : C → S → Sig → S → Set} ⦃ _ : Computational STS Err₂ ⦄ where instance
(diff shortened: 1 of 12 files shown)


src/Axiom/Set.agda
```diff
@@ -164,6 +164,9 @@ record Theory {ℓ} : Type (sucˡ ℓ) where
   ∈-fromList : ∀ {a} → a ∈ˡ l ⇔ a ∈ fromList l
   ∈-fromList = proj₂ $ listing _
 
+  listToFinSet : List A → FinSet A
```
WhatisRT, line 167: This is unused, and I don't think we'll ever need it. Or do you have a use-case in mind?

src/Axiom/Set/Map/Dec.agda
```diff
@@ -51,12 +51,12 @@ module Lookupᵐᵈ (sp-∈ : spec-∈ A) where
...
-    infixr 6 _∪⁺_
-    open IsCommutativeMonoid' mon
+  open import Algebra
+  module OpUnion {V : Type} (_∙_ : V → V → V) where
+    infixr 6 _∪̇_
 
-    _∪⁺_ : Map A V → Map A V → Map A V
-    _∪⁺_ = unionWith (fold id id _◇_)
+    _∪̇_ : ⦃ DecEq A ⦄ → Map A V → Map A V → Map A V
+    _∪̇_ = unionWith (fold id id _∙_)
```
WhatisRT, line 59: I'm not so sure about this change. It mostly seems to add boilerplate code elsewhere, like for example:
```
_∪⁺_ : ∀ {A} → ⦃ DecEq A ⦄ → A ⇀ ℕ → A ⇀ ℕ → A ⇀ ℕ
_∪⁺_ = OpUnion._∪̇_ _+_
```
If you like `_∪̇_`, why not define `_∪⁺_` with its original type in terms of it here? Same for `aggregate`.

src/Axiom/Set.agda
Soupstraw: I found a different way of solving the problem, so I can just get rid of this.

src/Axiom/Set/Map/Dec.agda
Soupstraw: I removed it since I don't really need that change.

WhatisRT (approved):
LGTM

PR #432, opened 2024-05-27
Update to Agda 2.6.4.3 & stdlib 2.0

WhatisRT:
# Description

This bumps some version & fixes issues that came from that.

# Checklist

- [x] Self-reviewed the diff


abailly:
I can confirm I was able to load `Everything.lagda` in my Emacs using a local install of stdlib 2.0, eg. not using nix.

WhatisRT:
I've come across a GHC bug, so I've reverted the nixpkgs bump for now. Will investigate some other time.

williamdemeo (approved):
I don't fully understand everything that's involved in this PR, however, I have tested the build on my machine and it works fine.

PR #439, opened 2024-05-29
Rename `Ledger.Deleg` to `Ledger.Certs` and overhaul its documentation

### LINKED ISSUE #406: document: Missing details regarding removal of pointer address support
Please document the changes coming regarding the pointer addresses in Conway era. Missing in the ledger specification

WhatisRT:
# Description

Closes #406. Originally I just wanted to address this issue, but then I got bothered by lots of other things in there, so I fixed them as well.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Certs.lagda
```diff
@@ -37,49 +37,82 @@ cwitness (ccreghot c _)      = c
 \end{figure*}
 
 \begin{figure*}[h!]
+\begin{AgdaAlign}
+\begin{AgdaSuppressSpace}
 \begin{code}
 record CertEnv : Set where
+\end{code}
+\begin{code}[hide]
   constructor ⟦_,_,_,_⟧ᶜ
```
williamdemeo, line 46: I see the rationale for hiding the constructors here and below, since they will be replaced by vertical vectors soon, but then, for consistency, I suppose we should also hide the other such constructors; e.g., those for `UTxOState`, `GovEnv`.

williamdemeo (changes requested):
Looks good. Just one minor change request.

src/Ledger/Certs.lagda
WhatisRT: Well, that's not a minor request. I made a new ticket for it, #441.

WhatisRT:
If you are fine with a PR and only want to make minor comments, then please approve it! You should reserve blocking a PR for only when you want a PR not merged in its current state!

williamdemeo:
When I say "minor comment," I don't always mean comments that can be ignored.  I seem to recall that on a recent PR review I made minor comments and marked the PR as "approved" and it seems you ignored the comments.  As a result, some bad grammar got merged. So this time I thought it would be better to make the change request official.

williamdemeo:
The files involved in keeping the rest of the documentation consistent with the changes in this PR may have nothing to do with this PR... but since you are making changes to the style of the documentation, shouldn't we make sure the rest of the docs reflect that change?

Anyway, it's not a lot of work to make the changes I'm suggesting.  There actually aren't many files involved... but if you don't want to do it, or don't think it should be done for some reason, that's fine.  I'll mark the PR approved.

williamdemeo (approved):

PR #443, opened 2024-06-03
Add `PParamsUpdate`

Soupstraw:
# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/PParams.lagda
+++ b/src/Ledger/PParams.lagda
@@ -112,6 +112,171 @@ paramsWellFormed pp =
                    ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit ∷ [] )
   where open PParams pp
 \end{code}
+\begin{code}[hide]
+instance
+  unquoteDecl DecEq-DrepThresholds = derive-DecEq
+    ((quote DrepThresholds , DecEq-DrepThresholds) ∷ [])
+  unquoteDecl DecEq-PoolThresholds = derive-DecEq
+    ((quote PoolThresholds , DecEq-PoolThresholds) ∷ [])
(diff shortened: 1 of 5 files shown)


src/Ledger/PParams.lagda
```diff
@@ -112,6 +113,156 @@ paramsWellFormed pp =
                    ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit ∷ [] )
   where open PParams pp
 \end{code}
+\begin{code}[hide]
+record PParamsUpdate : Set where
+  field
+        maxBlockSize maxTxSize        : Maybe {zeroˡ} ℕ
```
WhatisRT, line 119: I don't think you need any of those `{zeroˡ}`

src/Ledger/PParams.lagda
```diff
@@ -112,6 +113,156 @@ paramsWellFormed pp =
                    ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit ∷ [] )
   where open PParams pp
 \end{code}
+\begin{code}[hide]
+record PParamsUpdate : Set where
```
WhatisRT, line 117: How about putting this type & everything that relates to it in a module? Then you can leave `UpdateT` renamed to `PParamsUpdate` below, and don't have to change all the other files.

This should also fix the additional `⦃ _ : DecEq UpdateT ⦄` arguments you needed to introduce, since there'a a `⦃ DecEq-UpdT ⦄ : DecEq PParamsUpdate` field, but removing the renaming makes this the wrong instance.

WhatisRT (changes requested):
Looking good overall, just a bit of cleanup that's necessary.

WhatisRT (approved):
LGTM

PR #445, opened 2024-06-04
Don't parameterise the HasEmptySet class on a Theory

UlfNorell:
Only the instances need a theory, so I split them out into `Interface.HasEmptySet.Instances`. With this change instance search doesn't have to consider the particular theory, only the set representation.

It does mean that we can't have different empty sets for different theories with the same underlying set representation, but I don't think that's something we are likely to need.


WhatisRT (approved):
LGTM

PR #448, opened 2024-06-04
Make Credential a proper data type

UlfNorell:
This simplifies the marshalling to Haskell since we want a data type on the Haskell side and not an `Either`. Also it makes the spec easier to follow.

### DIFF
--- a/src/Ledger/Transaction.lagda
+++ b/src/Ledger/Transaction.lagda
@@ -66,7 +66,7 @@ the transaction body are:
   field crypto : _
   open Crypto crypto public
   open Ledger.TokenAlgebra ScriptHash public
-  open Ledger.Address Network KeyHash ScriptHash public
+  open Ledger.Address Network KeyHash ScriptHash ⦃ it ⦄ ⦃ it ⦄ ⦃ it ⦄ public
 
   field epochStructure : _
   open EpochStructure epochStructure public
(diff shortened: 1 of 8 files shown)


src/Ledger/Transaction.lagda
```diff
@@ -66,7 +66,7 @@ the transaction body are:
   field crypto : _
   open Crypto crypto public
   open Ledger.TokenAlgebra ScriptHash public
-  open Ledger.Address Network KeyHash ScriptHash public
+  open Ledger.Address Network KeyHash ScriptHash ⦃ it ⦄ ⦃ it ⦄ ⦃ it ⦄ public
```
WhatisRT, line 69: Is there a reason you added these? It works for me without them. Does that break with the new instance resolution?

UlfNorell: It broke some stuff for Haskell type generation.

Module applications don't add implicit/instance argument at the end, so without this, the types you get all take instance arguments that are resolved at use site, so when you want to generate a Haskell type for `Credential` say, the tactic needs to deal with types parameterised by instance arguments.

WhatisRT: Ah, makes sense. Thank you!

WhatisRT (approved):
LGTM

PR #454, opened 2024-06-05
Cleanup `Epoch` a bit, and a few other random things

### LINKED ISSUE #378: Clean up `EPOCH`
`EPOCH` is very complicated, especially considering that we use a lot of `open` and record stuff. Some of the logic that is responsible for splitting up rewards into what we pay out and what goes into the treasury is shared with `POOLREAP`, so it might make sense to factor that out in some way.

WhatisRT:
# Description

Closes #378.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -47,11 +47,17 @@ record NewEpochState : Set where
 private variable
   nes nes' : NewEpochState
   e lastEpoch : Epoch
-  fut' : RatifyState
+  fut fut' : RatifyState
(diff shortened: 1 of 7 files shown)


williamdemeo (approved):
Looks good.  I noticed $fut$ and $fut'$ are used in the documentation to refer to the corresponding private variables in the  EPOCH transition system figure.  I recommend we use `\AgdaBound{fut}` and `\AgdaBound{fut'}` for these instead so the font matches that of the corresponding variables in the figure.

Issue #444, opened 2024-06-03
What to do about `Set` in the PDF?


WhatisRT:
`Set` is confusing to people not familiar to Agda/type theory. Here are some options:
- rename/display it as `Prop` (or even use the actual `Prop`?) or `Type` (probably as confusing as `Set`)
- use something like `Rel`
- others?

WhatisRT:
We agreed that we'd rename `Set` to `Type` everywhere, and add a very short explanation for 'propositions as types'. If this is still confusing to people, we can try `Proposition` as an alias for `Type`.

jmchapman:
I would avoid renaming Set to Type and also not then capturing the name Set. 

I think this is from Bengt's book.

I wise man once said "Set is Agda's name for Type" <- you can write this in the document.

WhatisRT:
There are two reasons why `Set` is bad in our case:
- it is confusing to people who don't know Agda (why does `actionWellFormed` return a `Set`?)
- we also have 'actual' sets, which are different from `Set`.

Anything that people will understand without explanation is better than something that does need an explanation, that why I personally prefer `Prop` (or `Proposition`), but Ulf and William convinced me to go with `Type`. That does need an explanation, but is at least less confusing than `Set`.

williamdemeo:
I'll go with the consensus we arrived at on Tuesday (change Agda's Set to Type).  We can revert if this makes @jmchapman very unhappy 😰 ...an alternative would be to find a new notation for @WhatisRT's Set.  Perhaps, `aSet` or `AkSet`?  The latter stands for "Andre Knispel Set" or "Also known as Set" (depending on the context).

jmchapman:
or `set`, `SET`,`ZSet`, `ZFSet`, `ZFCSet`. Although if it only appears for predicates and relations I'm coming around to Andre's notion of `Prop` :)

jmchapman:
I think the Agda terminology comes from here:

https://www.cse.chalmers.se/research/group/logic/book/book.pdf

WhatisRT:
Naming our axiomatic sets `Set` and types `Type` is clearly the least confusing option out of these for non-Agda people. This could make it a little more confusing to Agda people reading our specs, but most of our readers know little to no Agda. And I think most Agda people are very receptive to renaming `Set` to `Type` (since many of them do it as well).

williamdemeo:
> I think the Agda terminology comes from here:
> 
[...]

PR #461, opened 2024-06-07
Rename Agda's `Set` to `Type`

williamdemeo:
# Description

This closes issue #444.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Introduction.lagda
+++ b/src/Ledger/Introduction.lagda
@@ -230,11 +230,9 @@ module _ (ℙ_ : Type → Type) (_∈_ : ∀ {A : Type} → A → ℙ A → Type
 
 In type theory we represent propositions as types and proofs of a proposition as
 elements of the corresponding type.
-A \textit{predicate} \AgdaFunction{P} represents a property or a relation.
 A unary predicate is a function that takes each \AgdaBound{x} (of some type \AgdaBound{A}) and
 returns a proposition \AgdaFunction{P}(\AgdaBound{x}). Thus, a predicate is a function of type
 \AgdaBound{A}~\AgdaSymbol{→}~\Type.
-
 A \textit{binary relation} \AgdaFunction{R} between \AgdaBound{A} and \AgdaBound{B} is a
 function that takes a pair of values \AgdaBound{x} and \AgdaBound{y} and returns a proposition
 asserting that the relation \AgdaFunction{R} holds between \AgdaBound{x} and \AgdaBound{y}.
--- a/src/Ledger/Introduction.lagda
+++ b/src/Ledger/Introduction.lagda
@@ -183,6 +183,7 @@ code that generates this document also generates a Haskell library
 that lets anyone run this code.
 
 \subsection{Sets \& maps}
+\label{sec:sets-maps}
 
 The ledger heavily uses set theory. For various reasons it was
 necessary to implement our own set theory (there will be a paper on this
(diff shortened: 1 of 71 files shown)


src/latex/agda-latex-macros.sty
```diff
@@ -446,7 +446,9 @@
 \newcommand{\RTC}{\AgdaFunction{ReflexiveTransitiveClosure}\xspace}
 \newcommand{\RTCI}{\AgdaOperator{\AgdaBound{\AgdaUnderscore{}⊢\AgdaUnderscore{}⇀⟦\AgdaUnderscore{}⟧*\AgdaUnderscore{}}}}
 \newcommand{\RTCB}{\AgdaOperator{\AgdaBound{\AgdaUnderscore{}⊢\AgdaUnderscore{}⇀⟦\AgdaUnderscore{}⟧\AgdaUnderscore{}}}}
-\newcommand{\Set}{\AgdaPrimitive{Set}\xspace}
+% \newcommand{\Set}{\AgdaPrimitive{Set}\xspace}
```
WhatisRT, line 449: There's no need keeping this around then

williamdemeo: I suppose we should keep the following around instead:
```
\newcommand{\Type}{\AgdaPrimitive{Type}\xspace}
```

williamdemeo:
> One thing that's still missing in the explanation that propositions, relations, etc. are also `Type`s.

I'm not sure whether we should say anything more than what I just added to the notation section (expanding on what was already mentioned about types).  In particular, I'm not sure we should mention the Curry-Howard correspondence.  Either readers know some type theory, in which case they already know about Curry-Howard, or they don't, in which case a few sentences about it probably won't help and may even make the pdf seem more advanced or "theoretical" than necessary.

WhatisRT:
I generally agree, but we still have to answer the actual question why a `Type` is a proposition. E.g. we got this question in the audit:
> Why is the output of validHFAction (Fig 13) a set (and not true/false), given that it is used to check validity? Similarly, canVote in Fig 13 also outputs a set.

So yes, we shouldn't explain Curry-Howard, but we need to at least make it clear that this is intentional, that there are good reasons for this (even if we don't say what they are) and that we cannot use a boolean. Note that this is exactly why I think `Prop` is a good name, because then we can just say "Prop is the type of propositions, which the reader can think of being an alternative name for Bool".

src/Ledger/Introduction.lagda
```diff
@@ -199,32 +200,47 @@ we use here to form a type of sets with elements in a given type.
...
+A \textit{predicate} \AgdaFunction{P} represents a property or a relation.
+A unary predicate is a function that takes each \AgdaBound{x} (of some type \AgdaBound{A}) and
+returns a proposition \AgdaFunction{P}(\AgdaBound{x}). Thus, a predicate is a function of type
+\AgdaBound{A}~\AgdaSymbol{→}~\Type.
+
+A \textit{binary relation} \AgdaFunction{R} from \AgdaBound{A} to \AgdaBound{B} is a
+function that takes a pair of values \AgdaBound{x} and \AgdaBound{y} and returns a proposition
+asserting that the relation \AgdaFunction{R} holds between \AgdaBound{x} and \AgdaBound{y}.
+Thus, such a relation is a function of type \AgdaBound{A}~\AgdaFunction{×}~\AgdaBound{B}~\AgdaSymbol{→}~\Type
+or \AgdaBound{A}~\AgdaSymbol{→}~\AgdaBound{B}~\AgdaSymbol{→}~\Type.
```
williamdemeo, line 242: This is the new paragraph addressing the change request.

src/Ledger/Introduction.lagda
```diff
@@ -199,32 +200,47 @@ we use here to form a type of sets with elements in a given type.
...
+\subsection{Propositions as types, properties and relations}
+
+In type theory we represent propositions as types and proofs of a proposition as
+elements of the corresponding type.
+A \textit{predicate} \AgdaFunction{P} represents a property or a relation.
+A unary predicate is a function that takes each \AgdaBound{x} (of some type \AgdaBound{A}) and
+returns a proposition \AgdaFunction{P}(\AgdaBound{x}). Thus, a predicate is a function of type
+\AgdaBound{A}~\AgdaSymbol{→}~\Type.
+
+A \textit{binary relation} \AgdaFunction{R} from \AgdaBound{A} to \AgdaBound{B} is a
```
WhatisRT, line 238:
```suggestion
A \textit{binary relation} \AgdaFunction{R} between \AgdaBound{A} and \AgdaBound{B} is a
```

WhatisRT: Often relations aren't directional

src/Ledger/Introduction.lagda
```diff
@@ -199,32 +200,47 @@ we use here to form a type of sets with elements in a given type.
...
+  _⇀_ : Type → Type → Type
   A ⇀ B = r ∈ Rel A B ﹐ left-unique r
 \end{code}
 \end{figure*}
 
+\subsection{Propositions as types, properties and relations}
+
+In type theory we represent propositions as types and proofs of a proposition as
+elements of the corresponding type.
+A \textit{predicate} \AgdaFunction{P} represents a property or a relation.
```
WhatisRT, line 233: I'm not sure what this sentence adds. Also, a relation is usually binary, so that sentence doesn't really make sense. I think it can just be removed safely.

src/Ledger/Introduction.lagda
williamdemeo: You have a point.  I've been using that wording for years because I felt it would make it less jarring for students when I tell them a function from A to B is just a special kind of binary relation from A to B.  I would hate to think that actually *added* to their confusion. 😟  In any case, I'll go with your suggested change, which is probably more standard anyway.

WhatisRT (approved):
LGTM

PR #464, opened 2024-06-11
larger rightharpoonup; add note about the symbol to notation sec

### LINKED ISSUE #325: `⇀` may benefit from a rename
I like this notation, but the issue is that it can be accidentally mixed up with the regular function arrow. It would have been nice to force parentheses around the operator, but this appears to be impossible. We could rename it to something like `（_⇀_）`, but that may be too heavy.

williamdemeo:
# Description

Closes #325 

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Notation.lagda
+++ b/src/Ledger/Notation.lagda
@@ -44,14 +44,14 @@ a list \AgdaBound{l} to its set of elements, we write \fromList~\AgdaBound{l}.
 \item[Map addition] For maps \AgdaBound{m} and \AgdaBound{m'}, we write
   \AgdaBound{m}~\AgdaFunction{∪⁺}~\AgdaBound{m'} for their union, where keys that appear
   in both maps have their corresponding values added.
-\item[Mapping a partial function] A \textit{partial function} is a function on \AgdaBound{A} which may not be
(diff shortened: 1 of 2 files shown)


src/Ledger/Notation.lagda
```diff
@@ -36,7 +36,10 @@ In this section, we introduce some notations used in this document.
   \AgdaBound{m}~\AgdaFunction{∪⁺}~\AgdaBound{m'} for their union, where keys that appear
   in both maps have their corresponding values added.
 \item[Maping a partial function] A partial function is a function on \AgdaBound{A} which may not be
-  defined for all elements of \AgdaBound{A}. The \mapPartial operation takes such a function \AgdaBound{f}
+  defined for all elements of \AgdaBound{A}. We denote such a function by
+  \AgdaBound{f}~:~\AgdaBound{A}⇀\AgdaBound{B}.  (If we happen to know that the function is total---defined for all
```
WhatisRT, line 40: I think there should be some extra spacing around the arrow here and the next one.

WhatisRT (approved):

WhatisRT:
Note that this PR still needs to be rebased before it can be merged.

PR #474, opened 2024-06-19
bugfix: formatting for mkStakeDistrs and others in Epoch

### LINKED ISSUE #472: Postprocessing duplicates code
The following code produces the wrong output:

```agda
mkStakeDistrs : Snapshot → GovState → Deposits → (Credential ⇀ VDeleg) → StakeDistrs
mkStakeDistrs ⟦ stake , _ ⟧ˢ govSt ds delegations .StakeDistrs.stakeDistr =
  aggregateBy (proj₁ delegations) (stake ∪⁺ gaDepositStake govSt ds)
```

Screenshot 2024-06-18 at 15 07 27

There are at least two issues here:
- [ ] the pattern after the vector shouldn't be on a new line
- [ ] the pattern is repeated three times

Also, I don't know if this is caused by the same problem, but the spacing of that pattern is super weird.

williamdemeo:
# Description

This closes #472 by addressing the two problems revealed by the new code:

- [X] the pattern after the vector shouldn't be on a new line
- [X] the pattern is repeated three times

The example shown in the issue should now be typeset as in the screenshot below.

mkStakeDistr

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/agda2vec.py
+++ b/src/agda2vec.py
@@ -11,6 +11,9 @@ def read_file(filename):
         return file.readlines()
 
 ### General Helper functions #########################################################################
+def is_slash_gt_number(s: str) -> bool:
+    pattern = r'\\>\[(?:[1-9]?\d)\]'
+    return bool(re.fullmatch(pattern, s))
 
 def rm_weird_agda_patterns(lines):
     # The pattern to match the specified form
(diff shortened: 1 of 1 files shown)


WhatisRT (approved):
LGTM

PR #488, opened 2024-06-27
Use reflection to generate Haskell types

UlfNorell:
I think this is ready now. It requires bleeding edge Agda (I've tested with Agda/Agda@d311d7b10a).

I couldn't find an issue for this (I thought there was one?).

The only hand-rolled types are `Rational`, sets and maps, and `ComputationResult`. These are defined in `Ledger.Foreign.HSTypes`. Everything else is auto generated to match the Agda types, except `GovProposal` and `GovActionState` that has a dependent field. These are generated through non-dependent versions (see `HSLedger.Gov`).

The tests in `cardano-ledger-executable-spec` pass, but minor caveat that the Makefile didn't work for me so I ran with
```bash
agda -c --ghc-dont-call-ghc --no-main --compile-dir ../dist/haskell/Ledger Ledger/Foreign/HSLedger.agda
cp Ledger/hs-src/cardano-ledger-executable-spec.cabal ../dist/haskell/Ledger
cp Ledger/hs-src/Lib.hs ../dist/haskell/Ledger
cp Ledger/hs-src/test/*.hs ../dist/haskell/Ledger/test
(cd ../dist/haskell/Ledger; \
 find MAlonzo -name '*.hs' -print | sed 's/^/      /;s/\.hs//;s+/+.+g' >> cardano-ledger-executable-spec.cabal; \
 cabal test)
```

I haven't touched `Midnight.HSLedger`.


WhatisRT:
I've spent some time trying to get all the nix stuff to work, but no luck so far. I don't really know whats wrong, especially since it worked before...

UlfNorell:
I'm afraid I can be of little help with nix.

If you can't get it working there's always the option to keep this branch alive until Agda-2.7.0 is released, which (hopefully) should be sometime early autumn.

WhatisRT:
Yeah, my plan is to get help if possible, and if that doesn't work out we just wait I guess

UlfNorell:
It looks like the 2.7.0 release is moving faster than I thought. There should be a release candidate available by next week.

WhatisRT:
That's great, thanks for letting me know!

WhatisRT (approved):
LGTM

Issue #487, opened 2024-06-25
Readability of the figures 2


WhatisRT:
#478 was getting too big, so I'm migrating stuff here.

- [x] "Figure 38: EPOCH transition system"
  - [x] doesn't fit on page; adding horizontal vector support to
        `agda2vec` script and making the first two vertical vectors in
        the figure horizontal should help.
  - [ ] It might make sense to replace some `record` construction with
        vertical vectors
- [x] Alignment is broken in `certDeposit`, Fig. 12
- [x] Some unnecessary line breaks after vertical vectors.
  - [x] Figure 37: `stakeDistr`
  - [x] Figure 41: Invocation of `LEDGER`

PR #500, opened 2024-07-06
Improve readability of the figures (2)

williamdemeo:
# Description

This closes issue #487.

# Checklist

- [X] Self-reviewed the diff


WhatisRT (approved):
LGTM. I think there should be a way to control whether a vector should be rendered horizontally on a case-by-case basis, but that can be done some other time.

Issue #501, opened 2024-07-08
Readability of the Figures 3


WhatisRT:
Some more stuff I found while reviewing PRs:

- [ ] Figure 11: `Derived types` is empty
- [ ] Figure 16: this is broken in a bunch of ways:
  - the first function is cut off
  - indentation of `case` is wrong
  - there's no empty line between `allEnactable` and `hasParentE`

PR #502, opened 2024-07-08
Readability of Figures: Fig 11 and Fig 16 adjustments.

williamdemeo:
# Description

This closes issue #501.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -110,7 +110,7 @@ validHFAction _ _ _ = ⊤
 \footnotetext{\AgdaBound{l}~\AgdaFunction{∷ʳ}~\AgdaBound{x} appends element \AgdaBound{x} to list \AgdaBound{l}.}
 \begin{figure*}[h]
 \begin{AgdaMultiCode}
-\begin{code}
+\begin{code}[hide]
 -- Convert list of (GovActionID, GovActionState) pairs.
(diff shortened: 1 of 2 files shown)


src/Ledger/Gov.lagda
```diff
@@ -110,13 +110,13 @@ validHFAction _ _ _ = ⊤
 \footnotetext{\AgdaBound{l}~\AgdaFunction{∷ʳ}~\AgdaBound{x} appends element \AgdaBound{x} to list \AgdaBound{l}.}
 \begin{figure*}[h]
 \begin{AgdaMultiCode}
-\begin{code}[hide]
--- convert list of (GovActionID,GovActionState)-pairs to list GovActionID pairs.
+\begin{code}
```
WhatisRT, line 113: This isn't what we wanted, this stuff was intentionally hidden (you've even proposed that). So the figure should be fixed without showing that again.

williamdemeo: Yes, I even proposed that, but in response to that proposal you said we would have to add an explanation to the prose if we hide this code, which I thought was a worse option than simply showing the code with a comment.

williamdemeo: I've gone ahead and removed the functions and added a paragraph describing the figure and mentioning the hidden function and their purpose.

WhatisRT (approved):
LGTM

Issue #407, opened 2024-05-01
document: ref scripts for plutus v1


ashisherc:
document the details regarding the support of ref scripts for plutus v1 in the ledger specification

WhatisRT:
Doing this the right way is currently blocked by #416. If we're under time pressure to get the spec out we may put this in the PDF manually, but for now I consider this being blocked.

WhatisRT:
Closed by #504.

PR #504, opened 2024-07-10
Add support for Plutus versions

### LINKED ISSUE #416: Add the ability to talk about Plutus versions
We do have a `Language` type, but it's abstract and don't have access to inhabitants like `PlutusV1`. There are at least two options to resolve this:

- Change `Language` to be a regular sum type
- Add new fields to `PlutusStructure`, such as `PlutusV1 : Language`

The second option has the benefit of being more flexible. `Language` could contain things that we don't need to talk about or once we have different versions of the spec we wouldn't need to translate between different types here. If there are no downsides, that option is preferred.

WhatisRT:
# Description

Closes #416 and #407. Because of issues with instance resolution, I had to add `HasInlineDatum` and `UsesV3Features` as explicit data types.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Utxow.lagda
+++ b/src/Ledger/Utxow.lagda
@@ -25,7 +25,7 @@ only if it is present.
 Conway. If a transaction contains any votes, proposals, a treasury
 donation or asserts the treasury amount, it is only allowed to contain
 Plutus V3 scripts. Additionally, the presence of reference scripts or
-inline scripts does not prevent Plutus V1 scripts to be used in a
+inline scripts does not prevent Plutus V1 scripts from being used in a
 transaction anymore. Only inline datums are now disallowed from
 appearing together with a Plutus V1 script.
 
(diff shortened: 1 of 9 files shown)


src/Ledger/Utxow.lagda
```diff
@@ -21,12 +21,77 @@ projecting the same information out of \credsNeeded. Note that the
 last component of \credsNeeded adds the script in the proposal policy
 only if it is present.
 
+\allowedLanguages has additional conditions for new features in
+Conway. If a transaction contains any votes, proposals, a treasury
+donation or asserts the treasury amount, it is only allowed to contain
+Plutus V3 scripts. Additionally, the presence of reference scripts or
+inline scripts does not prevent Plutus V1 scripts to be used in a
```
williamdemeo, line 28:
```suggestion
inline scripts does not prevent Plutus V1 scripts from being used in a
```

williamdemeo (approved):
Looks good... just one minor grammar suggestion.

PR #509, opened 2024-07-15
Votes in the state must be by registered entities

### LINKED ISSUE #491: Votes in the state must be by registered entities
For DReps, this is needs to be an invariant, so we need to check this when voting, and also remove votes when deregistering.

For SPOs and CC members, we just need to check that they exist when they vote.

WhatisRT:
# Description

Closes #491.

# Checklist

- [x] Self-reviewed the diff


WhatisRT (commented):
This looks good so far, the only issue is that it needs to be rebased. The whole FFI part has been overhauled since then, but it should just make things easier. There's now also a whole copy of the ledger in `Ledger.Conway.Conformance`, which also needs the same changes, but it should just be a copy & paste.

javierdiaz72:
This PR should be closed since the final work is now included in the new PR #603.

PR #514, opened 2024-07-15
Fix `ccMinSize` check

### LINKED ISSUE #508: `activeCC` does not double-count
For the `ccMinSize` check, if multiple cold credentials delegate to the same hot credential, they are only counted once, but should be counted multiple times.

WhatisRT:
# Description

Closes #508.

# Checklist

- [x] Self-reviewed the diff


jmchapman:
Can this be reviewed even though it's waiting for something?

WhatisRT:
The issue is that there is another bug with that code, so this probably needs to be thrown away anyway.

Soupstraw (approved):

WhatisRT:
Merging this now, we need to fix the other issue later.

PR #516, opened 2024-07-18
DO NOT MERGE: Ratify governance actions in the correct order

### LINKED ISSUE #427: Ratify governance actions in the correct order
We are currently ratifying the governance actions in order in which they were submitted, but this is only the secondary order. First, we need to sort them by their type.

The order is:
1. Motion of no-confidence
2. New committee/threshold
3. Update to the Constitution or proposal policy
4. Hard Fork initiation
5. Protocol parameter changes
6. Treasury withdrawals
7. Info

williamdemeo:
# Description

Will close #427.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -110,6 +110,14 @@ insertGovAction ((gaID₀ , gaSt₀) ∷ gaPrs) (gaID₁ , gaSt₁)
   with (govActionPriority (action gaSt₀)) ≤? (govActionPriority (action gaSt₁))
 ... | yes _  = (gaID₀ , gaSt₀) ∷ insertGovAction gaPrs (gaID₁ , gaSt₁)
 ... | no _   = (gaID₁ , gaSt₁) ∷ (gaID₀ , gaSt₀) ∷ gaPrs
+
+insertGovAction-∈ : (gSt : GovState){x : GovActionID × GovActionState}
+  → x ∈ˡ insertGovAction gSt x
+insertGovAction-∈ [] = Any.here refl
+insertGovAction-∈ ((gaID₀ , gaSt₀) ∷ gaPrs) {(gaID₁ , gaSt₁)}
+  with (govActionPriority (action gaSt₀)) ≤? (govActionPriority (action gaSt₁))
+... | yes _  = Any.there (insertGovAction-∈ gaPrs)
+... | no _   = Any.here refl
 \end{code}
 \begin{code}
 addVote : GovState → GovActionID → Voter → Vote → GovState
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -98,21 +98,10 @@ govActionPriority Info                     = 6
 insertGovAction : GovState → GovActionID × GovActionState → GovState
 insertGovAction [] gaPr = [ gaPr ]
 insertGovAction ((gaID₀ , gaSt₀) ∷ gaPrs) (gaID₁ , gaSt₁)
-  = if (govActionPriority (action gaSt₀)) ≤? (govActionPriority (action gaSt₁)) then
-      (gaID₀ , gaSt₀) ∷ insertGovAction gaPrs (gaID₁ , gaSt₁)
-    else
-      (gaID₁ , gaSt₁) ∷ (gaID₀ , gaSt₀) ∷ gaPrs
-\end{code}
-\begin{code}[hide]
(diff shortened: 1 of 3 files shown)


williamdemeo:
@WhatisRT  Will a `GovState` contain `GovAction`s with different `GovActionID`s?  If so, should the ordering of elements in the `GovState` take into account `GovActionID`s?

williamdemeo:
TODO: fix checks/invariants in `Ledger.Ledger.Properties` to account for inserting instead of appending new elements to `GovState`.

WhatisRT:
The `GovActionID`s are irrelevant for the order here.

src/Ledger/Gov.lagda
```diff
@@ -81,6 +82,40 @@ private variable
...
+govActionPriority : GovAction → ℕ
+govActionPriority NoConfidence             = 0
+govActionPriority (UpdateCommittee _ _ _)  = 1
+govActionPriority (NewConstitution _ _)    = 2
+govActionPriority (TriggerHF _)            = 3
+govActionPriority (ChangePParams _)        = 4
+govActionPriority (TreasuryWdrl _)         = 5
+govActionPriority Info                     = 6
+
+groupGovActions : GovState → GovState
```
WhatisRT, line 95: This is unused

src/Ledger/Gov.lagda
```diff
@@ -81,6 +82,40 @@ private variable
 \end{code}
 \emph{Functions used in the GOV rules}
 \begin{AgdaMultiCode}
+\begin{code}[hide]
+govActionPriority : GovAction → ℕ
```
WhatisRT, line 86: This needs to be in the visible part of the PDF

src/Ledger/Gov.lagda
```diff
@@ -81,6 +82,40 @@ private variable
...
+groupGovActions : GovState → GovState
+groupGovActions gs  =   filter (λ x → govActionPriority (action (proj₂ x)) ≟ 0) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 1) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 2) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 3) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 4) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 5) gs
+                    ++  filter (λ x → govActionPriority (action (proj₂ x)) ≟ 6) gs
+
+insertGovAction : GovState → GovActionID × GovActionState → GovState
```
WhatisRT, line 104: This also needs to be visible. Also, the `with` isn't that great there, since we'd have to explain it. But I think you should be able to do
```
if govActionPriority (action gaSt₀) ≤ govActionPriority (action gaSt₁) then ... else ...
```
and I expect that this probably doesn't break any proofs.

src/Ledger/Ledger/Properties.agda
```diff
@@ -134,6 +134,12 @@ module _ where
 
 -- ** Proof that the set equality `govDepsMatch` (below) is a LEDGER invariant.
 
+-- Mapping a list of `GovActionID × GovActionState`s to a list of
```
WhatisRT, line 137: It's very interesting to me that you're proving things about a more complicated relation now, but the proof has gotten a lot shorter.

WhatisRT (changes requested):
Overall I like it a lot, just a few cosmetic changes before we can merge it

src/Ledger/Gov.lagda
williamdemeo: Yes, it's unused now, but wouldn't it be useful if/when we want to make sure all gov actions in a GovState are in the right order? Well, I'll just delete it for now. We can always add it back in later if needed.

src/Ledger/Ledger/Properties.agda
williamdemeo: New gov actions used to be appended to the end of the GovState.  Now we insert according to the priority.  So I don't think it's much more complicated, though all the lemmas about list concatenation became pretty useless.  Also, this change gave me another chance to analyze, rework, and simplify the proof.

williamdemeo:
I removed another unused lemma (`insertGovAction-∈`) and made the new definition of `insertGovAction` more compact, so now the `insertGovAction` and `govActionPriority` functions fit inside Fig 15.

WhatisRT (approved):

PR #517, opened 2024-07-18
Reorder `PParams` and `PParamsUpdate` fields

### LINKED ISSUE #507: Rearrange `PParams` and `PParamsUpdate` fields so they match the order of the corresponding Haskell data type
As discussed during the meetings, it's a bit bothersome (and potentially dangerous) when the field order of the `Agda` and `Haskell` types diverge. This is currently the case regarding `PParams`, so this should be rectified for both the `Foreign` interface and for the `convertibles` in `BaseTypes`.

Lucsanszky:
# Description
Resolves #507 

# Checklist

- [ ] Self-reviewed the diff


WhatisRT:
Note that the parameters should still be grouped together according to their groups, otherwise specifying the groups will be a mess.

Lucsanszky:
> Note that the parameters should still be grouped together according to their groups, otherwise specifying the groups will be a mess.

Ah, okay. I'll just revert that commit then. How about getting rid of two fields on one line (e.g. `maxBlockSize maxTxSize`, `a b` etc.) and moving them to separate lines? @lehins requested it but I don't know if this still desired if we can't reorder the fields in the PDF to match the Haskell data type while also preserving the groupings.

WhatisRT:
Why not reorder the fields on the Haskell side as well? The implementation probably doesn't have any real preference on the order of the fields, and since positional matching should be almost unused it should be a fairly small change.

I don't mind splitting the lines and changing the order within the groups as well.

Lucsanszky:
> Why not reorder the fields on the Haskell side as well? The implementation probably doesn't have any real preference on the order of the fields, and since positional matching should be almost unused it should be a fairly small change.
> 
> I don't mind splitting the lines and changing the order within the groups as well.

I guess the reason why we would prefer to change the order here is because it's a smaller change. It seems like that on the Haskell side, the code is organised everywhere in a way to remain consistent with the field order (defaults, instances, lens etc.). Although you are right, that positional matching is mainly unused (except perhaps for a few places), doing this change on the Haskell side would be a much bigger hassle if we want to preserve consistency everywhere. Also, I think we would need to pay particular attention to the `EncCBOR`, `DecCBOR` instances on the Haskell side if the ordering is changed. 

Anyways, I think it would be a good compromise to merge my changes pertaining to the `BaseTypes` and `LedgerTypes` and I'll just leave the PDF alone for now. What do you think?

src/Ledger/Foreign/HSLedger/BaseTypes.agda
```diff
@@ -159,59 +159,59 @@ instance
...
         ; maxHeaderSize = to maxHeaderSize
-        ; maxValSize = to maxValSize
-        ; minUTxOValue = to minUTxOValue
-        ; poolDeposit = to poolDeposit
         ; keyDeposit = to keyDeposit
+        ; poolDeposit = to poolDeposit
         ; Emax = to Emax
         ; nopt = to nopt
         ; pv = to pv
+        ; minUTxOValue = to minUTxOValue
```
Lucsanszky, line 167: Although this has been retired after the `Mary` era and the Haskell implementation doesn't have it anymore, I opted to move it here to preserve consistency with the `Shelley` era field order, rather than moving it to the end of the record.

WhatisRT (approved):
LGTM

PR #520, opened 2024-07-22
hldiff script for highlighting new Agda objects

### LINKED ISSUE #437: `\hldiff` for Agda code
In previous specs we've had the `\hldiff` LaTeX macro to highlight what has changed with respect to previous specifications. To properly target subexpressions in the Agda code, we'd need some marker that is translated into highlighting after the LaTeX export.

williamdemeo:
# Description

This will close #437.

# Checklist

- [X] Self-reviewed the diff


src/Ledger/Transaction.lagda
```diff
@@ -55,7 +55,9 @@ Ingredients of the transaction body introduced in the Conway era are the followi
   \item \AgdaField{txvote}, the list of votes for goverance actions;
   \item \AgdaField{txprop}, the list of governance proposals;
   \item \AgdaField{txdonation}, the treasury donation amount;
-  \item \AgdaField{curTreasury}, the current value of the treasury.
+  \item \AgdaField{curTreasury}, the current value of the treasury;
+  \item \AgdaField{txsize}, the size of the serialized form of the transaction;
+  \item \AgdaField{txId}, the hash of the serialized form of the transaction.
```
williamdemeo, line 60: I added the `txsize` and `txId` since they do not appear in Babbage or Alonzo.

src/Ledger/Utxo.lagda
```diff
@@ -138,10 +138,10 @@ epoch boundary and are not removed by these functions.
...
 deposits. One option would be to populate the \deposits field of
 \UTxOState with the correct keys and values that can be extracted from
-the state of the previous era at the transition into the Conway
-era. Alternatively, this logic can be implemented in older eras and
-replaying the chain with that implementation, effectively treating it
-as an erratum to the Shelley specification.
+the state of the previous era at the transition into the Conway era.
+Alternatively, we can effectively treat the old handling of deposits
+as an erratum in the Shelley specification, which we fix by implementing
+the new deposits logic in older eras and then replaying the chain.
```
williamdemeo, line 144: (This sentence was badly in need of improvement, but I don't think the change deserves its own issue/PR.)

src/latex/conway-ledger.tex
```diff
@@ -34,8 +34,8 @@
 
 \appendix
 
-\input{Ledger/EssentialAgda}
-\input{Ledger/PDF/ConwayBootstrapEnact}
-\input{Ledger/PDF/ConwayBootstrap}
+\input{Conway/Ledger/EssentialAgda}
+\input{Conway/Ledger/PDF/ConwayBootstrapEnact}
+\input{Conway/Ledger/PDF/ConwayBootstrap}
```
williamdemeo, line 40: Moving the tex files generated for Conway to a separate directory allows for simpler build rules.  Otherwise, we have to be more careful about duplication in the Makefile and redefinition of build rules.

williamdemeo:
TODO: include all the terms we want highlighted in the Conway pdf in the file `hldiff_list.txt`.

WhatisRT:
Looking at the result of the github action, I don't see any yellow being applied anywhere. I'm also not convinced that the approach of putting things in a file is the best, since it's quite brittle. If we make a small change, we might lose highlighting without noticing. But we can change that some time after the HF.

williamdemeo:
> Looking at the result of the github action, I don't see any yellow being applied anywhere.

I'm not sure where you're looking, but the yellow highlighting that the `hldiff.py` script currently produces is limited to the Conway pdf and the six fields listed in the `hldiff_list.txt` file.  The resulting highlighting in `conway-ledger.pdf` is shown in the screenshot below. 

hldiff-sampl

One nice thing about this approach is that we can easily highlight whatever we want by simply placing a string in a single file (`hldiff_list.txt`).

> I'm also not convinced that the approach of putting things in a file is the best, since it's quite brittle. If we make a small change, we might lose highlighting without noticing. But we can change that some time after the HF.

Quite brittle relative to what?  Is there another approach you have in mind?  If so, please describe how it would work.

WhatisRT:
Every PR builds the PDF and uploads it as an artefact. For this PR, it's here: https://github.com/IntersectMBO/formal-ledger-specifications/actions/runs/10066064822?pr=520. If you download the `PDF specs` archive, `conway-ledger.pdf` doesn't have the highlighting.

There are two problems with the current approach. First it highlights everything that matches, which isn't what we want. The point of the highlighting it so that when you have a figure that has a bunch of old stuff mixed with new stuff you can tell the two apart. So for example, you want to highlight `txvote` in the definition of a transaction, but there is no point in highlighting it in the definition of `txgov` since that entire function is new. There is nothing to distinguish.

The second issue is that we want to highlight large pieces of code in some places. For example, in `credsNeeded` we should highlight the last two sets, i.e.
```
  ∪  mapˢ (λ v        → (Vote   v , proj₂ v)) (fromList $ map voter txvote)
  ∪  mapPartial (λ p  → case  p .policy of
                              (just sh)  → just (Propose  p , ScriptObj sh)
                              nothing    → nothing) (fromList txprop)
```
Now you could go and copy & paste that into the file, and I assume it'll work, but what happens if we decide to change it to this:
```
  ∪  mapˢ (λ v        → (Vote   v , proj₂ v)) (fromList (map voter txvote))
  ∪  mapPartial (λ p  → case  p .policy of
                              (just sh)  → just (Propose  p , ScriptObj sh)
                              nothing    → nothing) (fromList txprop)
```
Now you need to remember to manually update that line in the highlighting. So now on any code change, you need to ask yourself whether you broke any highlighting.

The solution to both of these issues were in the original text of the issue:
> To properly target subexpressions in the Agda code, we'd need some marker that is translated into highlighting after the LaTeX export.
One option would be to define a function like `⟪_⟫ = id`, and then you wrap everything you want to highlight in those brackets. This only works whenever we want to highlight a term however, so there would also have to be a backup solution such as some kind of tag, or maybe something else.

However, we don't have the time to mess with this much more before we need to have a finished document, so for now please just make sure the highlighting shows up. We can then do a better approach another time.

williamdemeo:
This was the only way I could think of to implement the highlighting feature without getting into the weeds of the agda.sty file.  I agree with all of your points, except for the following one:

> The solution to both of these issues were in the original text of the issue:
> 
> > To properly target subexpressions in the Agda code, we'd need some marker that is translated into highlighting after the LaTeX export.

In my view, "some marker that is translated into highlighting" seems very vague and is not, by itself, a solution.  I'm not even sure what a "marker" is in this context.

Thanks for the pointers to the artifacts.  Now that I'm aware of them, I can make certain what I think is working is actually working.

WhatisRT (approved):
LGTM

PR #530, opened 2024-08-02
Improve `MAlonzo` workflow

Lucsanszky:
# Description
- Create an orphan branch (`MAlonzo-code`) for the generated code
- When opening a PR for a feature branch, create a separate branch
  based on the `MAlonzo-code` branch which will contain newly
  generated code from the feature branch
- Once the PR is successfully merged, the branch based on
  `MAlonzo-code` gets squash merged into `MAlonzo-code` and deleted

Esentially, the above setup attempts to replicate
the git feature branch workflow for the generated code,
where the `MAlonzo-code` branch corresponds to the `main` / `master`
branch.

There will probably be some conflicts from time to time but these will be taken care of automatically by rebasing.
I opted to always keep the most recent changes, so the changes from the `*-MAlonzo` "feature" branches, rather than the ones from `MAlonzo-code`. This should mostly be okay, since the feature branches that we create are based on `master`, the `*-MAlonzo` branches contain the generated code of these feature branches (and we keep this up-to-date with `master` as required) and these are merged into `MAlonzo-code` when the PR is merged so the `MAlonzo-code` branch will contain generated code that is based on `master` by proxy and kept up-to-date via the feature branches of the PRs. However, since we can directly push to `master`, that can interfere with this slightly and might cause some problems.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,11 +10,24 @@ on:
 permissions:
   contents: write
 
+env:
+  MAlonzo_branch: ${{ github.event.pull_request.head.ref }}-MAlonzo
+
 jobs:
   build:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: MAlonzo-code
+
+      - name: Create branch ${{ env.MAlonzo_branch }} for generated code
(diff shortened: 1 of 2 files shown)


Soupstraw (approved):
This looks much better, thanks

PR #535, opened 2024-08-08
Prevent invalid CC from blocking NoConfidence

Soupstraw:
# Description

This PR fixes a bug that makes it impossible to pass `NoConfidence` and `UpdateCommittee` when the CC size is below `ccMinSize`.

# Checklist

- [ ] Self-reviewed the diff


WhatisRT (approved):
Good catch!

Issue #541, opened 2024-08-14
Restrict withdrawals only for key hashes


WhatisRT:
Right now, we check `wdrlCreds ⊆ dom voteDelegs` in `CERTBASE`. However, we should take the subset of `wdrlCreds` that contains only the elements with a `KeyHashObj` constructor.

PR #543, opened 2024-08-14
DO NOT MERGE: Restrict withdrawals to KeyHashObj in subset relation

williamdemeo:
# Description

Closes issue #541.

# Checklist

- [X] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Certs.lagda
+++ b/src/Ledger/Certs.lagda
@@ -429,9 +429,9 @@ data _⊢_⇀⦇_,CERTBASE⦈_ where
     let open PParams pp
         refresh         = mapPartial getDRepVote (fromList vs)
         refreshedDReps  = mapValueRestricted (const (e + drepActivity)) dreps refresh
-        wdrlCreds       = filterˢ isKeyHash (mapˢ stake (dom wdrls))
+        wdrlCreds       = mapˢ stake (dom wdrls)
     in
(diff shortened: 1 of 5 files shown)


WhatisRT (approved):
LGTM

src/Ledger/Certs.lagda
```diff
@@ -429,7 +429,7 @@ data _⊢_⇀⦇_,CERTBASE⦈_ where
     let open PParams pp
         refresh         = mapPartial getDRepVote (fromList vs)
         refreshedDReps  = mapValueRestricted (const (e + drepActivity)) dreps refresh
-        wdrlCreds       = mapˢ stake (dom wdrls)
+        wdrlCreds       = filterˢ isKeyHash (mapˢ stake (dom wdrls))
```
WhatisRT, line 432: You shouldn't update the binding here, since `wdrlCreds` is used twice. But the only place where the filter should happen is in `wdrlCreds ⊆ dom voteDelegs`, not in the state update. Otherwise we'd break preservation of Ada.

WhatisRT (changes requested):
Sorry, I missed an issue

src/Ledger/Certs.lagda
williamdemeo: @WhatisRT This should be fixed now.

WhatisRT (approved):

Issue #540, opened 2024-08-12
Missing network ID checks


WhatisRT:
We need to check that the network IDs in the return address of proposals and in treasury withdrawal proposals is correct.

PR #544, opened 2024-08-14
Missing network id checks

williamdemeo:
# Description

Closes issue #540.

# Checklist

- [x] Self-reviewed the diff


WhatisRT (changes requested):
As discussed on Slack, you should include the check for treasury withdrawals, and I prefer using `addr` over `(GovProposal.returnAddr prop)`, which is the same thing anyway.

WhatisRT:
It turns out the check is already implemented in `actionWellFormed`, so only the return address needs to be checked.

WhatisRT (approved):
LGTM

Issue #512, opened 2024-07-15
Cannot register and deregister (or the other way) the same stake credential in the same transaction


WhatisRT:
This is because the deposits are updated in `UTXO`, so during `CERT` the deposits map is not being updated and only at most one of those certs will be valid.

Here are some ways of fixing this:
- [ ] ~Add a `temporaryDeposits` field to `CertState`, that is updated and properly but then discarded~
- [ ] Check the correctness of refunds in `UTXO`, maybe during `updateCertDeposits`

TimSheard:
Consider this work around.

newtype  DeltaDeposit = DeltaDeposit (Map DepositPurpose DeltaCoin)poist
emptyDD = DeltaDepost Map.empty
delta :: DeltaDeposit ->Cert -> DeltaDeposit
addDD, subDD :: DeltaDeposit -> CertState -> CertState
  To add and subtract the changes to the CertState


In a transaction with N certs, we  let deltaDD = foldl' delta emptyDD certs
In the conformance test, in the CERTS rule we subtract the delta deposits (since the Agda rule didn't add them)
In the UTXO rule we add the delta deposits, since the Agda rule added them but the STS rule did not.

TimSheard:
Something strange going on here.  There are 4 constructors in the DELEG signal.

`data ConwayDelegCert c
  = -- | Register staking credential. Deposit, when present, must match the expected deposit
    -- amount specified by `ppKeyDepositL` in the protocol parameters.
    ConwayRegCert !(StakeCredential c) !(StrictMaybe Coin)
  | -- | De-Register the staking credential. Deposit, if present, must match the amount
    -- that was left as a deposit upon stake credential registration.
    ConwayUnRegCert !(StakeCredential c) !(StrictMaybe Coin)
  | -- | Redelegate to another delegatee. Staking credential must already be registered.
    ConwayDelegCert !(StakeCredential c) !(Delegatee c)
  | -- | This is a new type of certificate, which allows to register staking credential
    -- and delegate within a single certificate. Deposit is required and must match the
    -- expected deposit amount specified by `ppKeyDepositL` in the protocol parameters.
    ConwayRegDelegCert !(StakeCredential c) !(Delegatee c) !Coin`

But the spec only has 2 rules for DELEG Figure 22 page 25.  why is that? Are we not showing some of the rules that already appear in earlier Eras?

jmchapman:
@WhatisRT - William's initial version uses the first suggestion above 
[...]

PR #547, opened 2024-08-16
Cannot register and deregister same stake credential

williamdemeo:
# Description

This addresses issue #512

# Checklist

- [x] Self-reviewed the diff


WhatisRT (changes requested):
I think all the changes in `Ledger.Certs` and `Ledger.Epoch` are just reverting formatting/cleanup that I've done over a month ago. Can you revert those two files?

Also, note that `Ledger.Utxo` is a literate file, so please make sure it looks good in the PDF. In particular, the decision procedure should be hidden, and `ValidCertDeposits` might be better explained in prose than shown in a figure.

Otherwise, I think this looks pretty good!

src/Ledger/Utxo/Properties.lagda
```diff
@@ -684,23 +663,25 @@ module _
...
     getCoin d + _                  ≡⟨ sym ∪⁺singleton≡ ⟩
-    getCoin (d ∪⁺ ❴ _ ❵)           ≡⟨ getCoin∣∅≡id  {d ∪⁺ ❴ _ ❵} ⟩
-    getCoin ((d ∪⁺ ❴ _ ❵) ∣ ∅ ᶜ )  ∎
+    getCoin (d ∪⁺ ❴ _ ❵)           ∎
     where open ≤-Reasoning
 
   ≤updateCertDeps : (cs : List DCert) {pp : PParams} {deposits :  DepositPurpose ⇀ Coin}
-    → noRefundCert cs → getCoin deposits ≤ getCoin (updateCertDeposits pp cs deposits)
+    → noRefundCert cs
+    -- → ValidCertDeposits (mapˢ proj₁ (deposits ˢ)) cs   -- Do we need this? Maybe not.
```
WhatisRT, line 676: Yeah, that's not required for this proof

WhatisRT (commented):
Ah, sorry, I spoke too soon. I wrote
> Check the correctness of refunds in `UTXO`

in the issue, but what I meant was to check it there instead of in `CERTS`. So it's also necessary to remove those checks from `CERTS`, because that's what actually prevents us from registering and deregistering in the same transaction. I think the best way to do that is to remove `deposits` from `CertEnv` and then follow all the errors.

Note that this also means that `ValidCertDeposits` needs to be slightly adjusted, so that all the removed checks are done there. In particular, it also needs to check deposit amounts, so it needs to have a `Deposits` argument rather than `ℙ DepositPurpose`.

williamdemeo:
> I think all the changes in `Ledger.Certs` and `Ledger.Epoch` are just reverting formatting/cleanup that I've done over a month ago. Can you revert those two files?
> 

I made (most of) those changes to try to ensure that the vertical vector formatting doesn't get confused by the line breaks.  (The insertion of the `module` with no parameters was a mistake.)  In any case, I'll revert the changes and check if the pdf looks okay.  But I may need to reformat the vectors as long single lines so they look okay in the pdf.

williamdemeo:
> Ah, sorry, I spoke too soon. I wrote
> 
> > Check the correctness of refunds in `UTXO`
> 
> in the issue, but what I meant was to check it there instead of in `CERTS`. So it's also necessary to remove those checks from `CERTS`, because that's what actually prevents us from registering and deregistering in the same transaction. I think the best way to do that is to remove `deposits` from `CertEnv` and all the other environments in the file and then follow all the errors.
> 
> Note that this also means that `ValidCertDeposits` needs to be slightly adjusted, so that all the removed checks are done there. In particular, it also needs to check deposit amounts, so it needs to have a `Deposits` argument rather than `ℙ DepositPurpose`. I don't think it should affect much though.

@UlfNorell do you have time to work on these changes?  I'd like to get back to finishing the `Ledger` pov PR and work on issue #578?

Btw, I merged your branch into the one on which this PR is based and fill the remaining holes and rewrote some of the proofs, so it now type-checks and passes the CI.  I suppose you will want to make a new branch off this one like you did last time (assuming you have time to work on this).  Thanks!!

UlfNorell:
@williamdemeo I'll try to make some progress today. If it's ok with you I'll just push to this branch. Last time was a major change to the way things were done, but now it's just continuing work on the PR.

williamdemeo:
Sure, just push to this branch. 👍

williamdemeo:
> > But I may need to reformat the vectors as long single lines so they look okay in the pdf.
> 
> Did something change that affects the vectors? I remember being able to strategically insert newlines in certain places without them affecting how the PDF renders. Also, if I look at `CERTBASE` and `certState'` in `EPOCH`, they look fine in the PDF right now, I don't see what needs to be fixed.

Ah, sorry, I didn't realize the script output could be manipulated by strategically inserting newlines.  I guess nothing needs to be fixed then. 👍🏼

williamdemeo:
@UlfNorell I just merged the latest master, so I suggest a `git pull` before doing more work on this.

src/Ledger/Certs.lagda
```diff
@@ -369,14 +366,13 @@ data _⊢_⇀⦇_,GOVCERT⦈_ where
   GOVCERT-regdrep : ∀ {pp} → let open PParams pp in
     ∙ (d ≡ drepDeposit × c ∉ dom dReps) ⊎ (d ≡ 0 × c ∈ dom dReps)
       ────────────────────────────────
-      ⟦ e , pp , vs , wdrls , deps ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
+      ⟦ e , pp , vs , wdrls ⟧ᶜ ⊢  ⟦ dReps , ccKeys ⟧ᵛ ⇀⦇ regdrep c d an ,GOVCERT⦈
                                   ⟦ ❴ c , e + drepActivity ❵ ∪ˡ dReps , ccKeys ⟧ᵛ
 
   GOVCERT-deregdrep :
     ∙ c ∈ dom dReps
-    ∙ (CredentialDeposit c , d) ∈ deps
```
UlfNorell, line 377: Should this have been `(DRepDeposit c , d) ∈ deps`? That's what I put in `ValidCertDeposits`.

UlfNorell:
I think I managed moving the checks from `CERT` to `ValidCertDeposits`. There were just one small question about whether the check in `GOVCERT-deregdrep` was wrong before.

williamdemeo:
@WhatisRT I finished fixing the pdf.  I think this PR is good to go now.

src/Ledger/Utxo.lagda
```diff
@@ -130,8 +130,12 @@ updating this map, which is split into \updateCertDeposits and
 \updateProposalDeposits, responsible for certificates and proposals
 respectively. Both of these functions iterate over the relevant fields
 of the transaction body and insert or remove deposits depending on the
-information seen. Note that some deposits can only be refunded at the
-epoch boundary and are not removed by these functions.
+information seen.  More precisely, \updateCertDeposits is responsible
```
WhatisRT, line 133: There are now two explanations of `updateCertDeposits` and `updateProposalDeposits` here. Please remove this duplication.

src/Ledger/Utxo.lagda
```diff
@@ -205,9 +209,9 @@ module _ (let open Tx; open TxBody; open TxWitnesses) where opaque
 \end{NoConway}
 \begin{code}
   refScriptsSize : UTxO → Tx → ℕ
-  refScriptsSize utxo tx = ∑[ x ← mapValues scriptSize (setToHashMap (refScripts tx utxo)) ] x
+  refScriptsSize ux tx = ∑[ x ← mapValues scriptSize (setToHashMap (refScripts tx ux)) ] x
```
WhatisRT, line 212: We typically call `UTxO` variables `utxo` everywhere, so this change makes it inconsistent. It's also not a long variable name, so I think it's better to revert this line.

src/Ledger/Utxo.lagda
```diff
@@ -230,11 +234,76 @@ certRefund (dereg c _)      = ❴ CredentialDeposit c ❵
...
+  ... | yes p = mapDec (deregdrep p)  (λ where (deregdrep _ v) → v) (validCertDeposits? _ _)
+  ... | no ¬p = no (λ where (deregdrep p _) → ¬p p)
+
+instance
+  Dec-ValidCertDeposits : ∀ {pp deps certs} → ValidCertDeposits pp deps certs ⁇
+  Dec-ValidCertDeposits = ⁇ (validCertDeposits? _ _)
+\end{code}
+\begin{code}
+
+-- Assumes ValidCertDeposits (mapˢ proj₁ (deposits ˢ)) certs.
```
WhatisRT, line 292: This comment should rather be explained in the prose.

src/Ledger/Utxo.lagda
```diff
@@ -376,6 +444,14 @@ to \consumed or \produced depending on its sign. This is done via
 \negPart and \posPart, which satisfy the key property that their
 difference is the identity function.
 
+Figures~\ref{fig:functions:utxo} also shows the signature of an inductive type
```
WhatisRT, line 447: I don't like the mention of 'inductive type' here. It's an inductive proof, so you won't be able to understand this unless you understand 'propositions as types'. We avoid this elsewhere, so I think it could be avoided here as well.

WhatisRT (approved):
Just some minor stuff, otherwise LGTM

src/Ledger/Utxo.lagda
williamdemeo: I changed it to `ux` because saving those two characters makes the line fit inside the figure without splitting it up into multiple lines... but I'll change it back.

src/Ledger/Utxo.lagda
williamdemeo: I've already added to the prose, "Suffice it to say that \ValidCertDeposits is used to check the validity of the deposits in a transaction so that the function \updateCertDeposits can correctly register and deregister deposits in the UTxO state based on the certificates in the transaction."  Perhaps that's enough.

Issue #546, opened 2024-08-15
Improve output of `genErrors`


WhatisRT:
- [ ] `genErrors` always generates fully qualified names. The output would be a lot more readable if we postprocess the string to remove those qualifiers.
- [ ] We could also give a number that indicates which failure happened, for easier identification.
- [ ] The output generates non-breaking spaces. We may want to replace them with regular spaces (but maybe that should be a downstream responsibility?)

WhatisRT:
Note that that PR didn't fix any of those issues

williamdemeo:
> Note that that PR didn't fix any of those issues

As for the first issue, the fully qualified names issue is not an issue if we avoid `genErrors` and use simpler error handling, like the simple function I implemented to find the bug causing the latest conformance test failure.  When I used `show` in that function, the names did not come out fully qualified.

Also, using a bespoke error generator function has the added benefit of allowing us to say which error occurred.

I know this isn't a scalable, automated approach, but it's useful when we know what we're looking for (as in the most recent case).

WhatisRT:
Yeah, this is strictly an issue about `genErrors`, not about producing error messages in general. Of course a manual approach is going to be prettier, but it has the downside of being brittle: if you change the order of some clauses without updating the error messages for example, the error message will be misleading. This actually happened in the past which is why we came up with `genErrors` in the first place.

PR #551, opened 2024-08-21
Improve output of generrors

williamdemeo:
# Description

This will address issue #546.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Address.lagda
+++ b/src/Ledger/Address.lagda
@@ -23,9 +23,8 @@ credential contains a hash, either of a verifying (public) key
   ScriptHash
 \end{code}
 \begin{code}[hide]
-  : Type)  ⦃ _ : DecEq Network ⦄
-           ⦃ _ : DecEq KeyHash ⦄     ⦃ _ : Show KeyHash ⦄
-           ⦃ _ : DecEq ScriptHash ⦄  ⦃ _ : Show ScriptHash ⦄ where
+  : Type)  ⦃ _ : DecEq Network ⦄ ⦃ _ : DecEq KeyHash ⦄ ⦃ _ : DecEq ScriptHash ⦄
+           ⦃ _ : Show Network  ⦄ ⦃ _ : Show KeyHash  ⦄ ⦃ _ : Show ScriptHash  ⦄ where
 \end{code}
 \emph{Derived types}
 \AgdaTarget{Credential, BaseAddr, BootstrapAddr, RwdAddr, net, pay, stake, Addr,
@@ -76,6 +75,9 @@ record RwdAddr : Type where
         stake  : Credential
 \end{code}
 \begin{code}[hide]
+instance
+  unquoteDecl Show-RwdAddr = derive-Show [ (quote RwdAddr , Show-RwdAddr) ]
+
 open BaseAddr; open BootstrapAddr; open BaseAddr; open BootstrapAddr
 \end{code}
 \begin{code}
--- a/src/Ledger/Address.lagda
+++ b/src/Ledger/Address.lagda
@@ -23,8 +23,7 @@ credential contains a hash, either of a verifying (public) key
   ScriptHash
 \end{code}
 \begin{code}[hide]
-  : Type)  ⦃ _ : DecEq Network ⦄ ⦃ _ : DecEq KeyHash ⦄ ⦃ _ : DecEq ScriptHash ⦄
-           ⦃ _ : Show Network  ⦄ ⦃ _ : Show KeyHash  ⦄ ⦃ _ : Show ScriptHash  ⦄ where
+  : Type)  ⦃ _ : DecEq Network ⦄ ⦃ _ : DecEq KeyHash ⦄ ⦃ _ : DecEq ScriptHash ⦄ where
 \end{code}
 \emph{Derived types}
 \AgdaTarget{Credential, BaseAddr, BootstrapAddr, RwdAddr, net, pay, stake, Addr,
@@ -35,9 +34,6 @@ data Credential : Type where
   ScriptObj  : ScriptHash → Credential
 \end{code}
 \begin{code}[hide]
-instance
-  unquoteDecl Show-Credential = derive-Show [ (quote Credential , Show-Credential) ]
-
 isKeyHashObj : Credential → Maybe KeyHash
 isKeyHashObj (KeyHashObj h) = just h
 isKeyHashObj (ScriptObj _)  = nothing
@@ -75,9 +71,6 @@ record RwdAddr : Type where
         stake  : Credential
 \end{code}
 \begin{code}[hide]
-instance
-  unquoteDecl Show-RwdAddr = derive-Show [ (quote RwdAddr , Show-RwdAddr) ]
-
 open BaseAddr; open BootstrapAddr; open BaseAddr; open BootstrapAddr
 \end{code}
 \begin{code}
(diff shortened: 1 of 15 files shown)


williamdemeo:
I'm stuck on the following type-check error:

```agda
HSTransactionStructure != ⊤ of type Type
when checking that the pattern KeyHashObj x has type
Ledger.Address.Credential
(Ledger.Types.Epoch.GlobalConstants.Network
 (TransactionStructure.globalConstants HSTransactionStructure))
(Ledger.Crypto.isHashableSet.THash
 (Ledger.Crypto.Crypto.khs
  (TransactionStructure.crypto HSTransactionStructure)))
(Ledger.Crypto.Crypto.ScriptHash
 (TransactionStructure.crypto HSTransactionStructure))
```

at the fourth line of the following code block from `Ledger.Foreign.HSLedger.BaseTypes.agda`:

```agda
  -- Since the foreign address is just a number, we do bad stuff here
  Convertible-Addr : Convertible Addr F.Addr
  Convertible-Addr = λ where
    .to → λ where (inj₁ record { pay = KeyHashObj x }) → x
                  (inj₁ record { pay = ScriptObj  x }) → x
                  (inj₂ record { pay = KeyHashObj x }) → x
                  (inj₂ record { pay = ScriptObj  x }) → x
    .from n → inj₁ record { net = _ ; pay = KeyHashObj n ; stake = KeyHashObj 0 }
```

src/Ledger/Certs/Haskell/Properties.agda
```diff
@@ -121,13 +121,29 @@ instance
...
+
+    genErr : ¬ ( filterˢ isKeyHash (mapˢ RwdAddr.stake (dom wdrls)) ⊆ dom voteDelegs
+                 × mapˢ (Bifunctor.map₁ Bifunctor-× (RwdAddr.stake)) (wdrls ˢ) ⊆ proj₁ rewards)
+                  → String
+    genErr ¬p = case dec-de-morgan ¬p of λ where
+      (inj₁ a) → sep + " ¬ ( filterˢ isKeyHash (mapˢ RwdAddr.stake (dom wdrls)) ⊆ dom voteDelegs ) "
+                 + sep + " filterˢ isKeyHash (mapˢ RwdAddr.stake (dom wdrls)): "
+                 -- + show (filterˢ isKeyHash (mapˢ RwdAddr.stake (dom wdrls)))
+                 + show (dom wdrls)
+      (inj₂ b) → "¬ ( mapˢ (Bifunctor.map₁ Bifunctor-× (RwdAddr.stake)) (wdrls ˢ) ⊆ proj₁ rewards )"
```
Soupstraw, line 139: This approach can go out of sync when somebody changes the order of predicates. I think it's fine to use `genErrors` and then just append all of the debug information regardless of which rule fails.

Soupstraw:
> Commit [4659a61](https://github.com/IntersectMBO/formal-ledger-specifications/pull/551/commits/4659a61644d9e0eb8052a6b984cbe3d4587204aa) fails with the following type-check error:
> 
> ```agda
> HSTransactionStructure != ⊤ of type Type
> when checking that the pattern KeyHashObj x has type
> Ledger.Address.Credential
> (Ledger.Types.Epoch.GlobalConstants.Network
>  (TransactionStructure.globalConstants HSTransactionStructure))
> (Ledger.Crypto.isHashableSet.THash
>  (Ledger.Crypto.Crypto.khs
>   (TransactionStructure.crypto HSTransactionStructure)))
> (Ledger.Crypto.Crypto.ScriptHash
>  (TransactionStructure.crypto HSTransactionStructure))
> ```
> 
> at the fourth line of the following code block from `Ledger.Foreign.HSLedger.BaseTypes.agda`:
> 
> ```agda
>   -- Since the foreign address is just a number, we do bad stuff here
>   Convertible-Addr : Convertible Addr F.Addr
>   Convertible-Addr = λ where
>     .to → λ where (inj₁ record { pay = KeyHashObj x }) → x
>                   (inj₁ record { pay = ScriptObj  x }) → x
>                   (inj₂ record { pay = KeyHashObj x }) → x
>                   (inj₂ record { pay = ScriptObj  x }) → x
>     .from n → inj₁ record { net = _ ; pay = KeyHashObj n ; stake = KeyHashObj 0 }
> ```

I think that when you try to use the `KeyHashObj` and `ScriptObj` constructors then Agda tries to resolve the instance parameters for `Ledger.Address` and then fails to find them. I think you're just missing some `Show` instances.

src/Ledger/Address.lagda
```diff
@@ -22,7 +23,8 @@ credential contains a hash, either of a verifying (public) key
   ScriptHash
 \end{code}
 \begin{code}[hide]
-  : Type) ⦃ _ : DecEq Network ⦄ ⦃ _ : DecEq KeyHash ⦄ ⦃ _ : DecEq ScriptHash ⦄ where
+  : Type)  ⦃ _ : DecEq Network ⦄ ⦃ _ : DecEq KeyHash ⦄ ⦃ _ : DecEq ScriptHash ⦄
+           ⦃ _ : Show Network  ⦄ ⦃ _ : Show KeyHash  ⦄ ⦃ _ : Show ScriptHash  ⦄ where
```
Soupstraw, line 27: I wouldn't add the `Show` instances as module parameters since they're only needed for defining other `Show` instances. It'd be better to move these parameters to the `Show` instance definitions that are actually using them.

WhatisRT: Yeah. Might be worth considering doing the same for `DecEq`, but there might be some stuff that requires it that isn't some other `DecEq` instance in which case I think this might be cleaner.

src/Ledger/Certs/Haskell/Properties.agda
WhatisRT: Yes, the error messages don't have to be all that readable, they're just for us. And not having to maintain them is a big plus. Just doing some super basic postprocessing of the generated strings to get rid of the fully qualified names also goes a long way in improving readability.

WhatisRT:
> I think that when you try to use the KeyHashObj and ScriptObj constructors then Agda tries to resolve the instance parameters for Ledger.Address and then fails to find them. I think you're just missing some Show instances.

No, it's something much weirder. This error message is in the pattern and the instances actually get fully resolved somewhere earlier, so `KeyHashObj` doesn't have instance arguments anymore.

Looks like it could be an Agda bug to me, especially because the first thing I tried when playing around with this triggered an `IMPOSSIBLE` error. I guess I'll go on the third code bisection this week...

WhatisRT:
Oh, no, it's actually what you said. If you add `⦃ it ⦄ ⦃ it ⦄ ⦃ it ⦄` at the end of `open Ledger.Address Network KeyHash ScriptHash ⦃ it ⦄ ⦃ it ⦄ ⦃ it ⦄ public` in `Ledger.Transaction` it works just fine.

I still don't understand why this happens though. `KeyHashObj` has the instances fully applied, but maybe this breaks when used in a pattern?

src/Ledger/Foreign/LedgerTypes.agda
```diff
@@ -131,14 +131,14 @@ ProtVer = Pair ℕ ℕ
...
+    = KeyHashObj Integer
+    | ScriptObj Integer
     deriving (Show, Eq, Generic)
 #-}
 data Credential : Type where
-  ScriptObj  : Hash → Credential
   KeyHashObj : Hash → Credential
-{-# COMPILE GHC Credential = data Credential (ScriptObj | KeyHashObj) #-}
+  ScriptObj  : Hash → Credential
+{-# COMPILE GHC Credential = data Credential (KeyHashObj | ScriptObj) #-}
```
TimSheard, line 141: Ah ... The culprit has been found!

TimSheard (approved):
So there is a lot of work involved in making Show instances in Agda.
These will probably come in handy in the future as well.

PR #553, opened 2024-08-27
hasParent needs to match GAs

### LINKED ISSUE #539: `hasParent` needs to match GAs
Currently we just throw away everything except the `GovActionID`. Instead, we need to ensure that the parent also has the same type of GA, otherwise `hasParent` should fail.

williamdemeo:
# Description

This closes #539.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Gov.lagda
+++ b/src/Ledger/Gov.lagda
@@ -95,17 +95,20 @@ govActionPriority (ChangePParams _)        = 4
 govActionPriority (TreasuryWdrl _)         = 5
 govActionPriority Info                     = 6
 
-_gaPrRel_ : ℕ → ℕ → Type
-n gaPrRel m = (n ≡ m) ⊎ (n ≡ 0 × m ≡ 1) ⊎ (n ≡ 1 × m ≡ 0)
+_∼_ : ℕ → ℕ → Type
+n ∼ m = (n ≡ m) ⊎ (n ≡ 0 × m ≡ 1) ⊎ (n ≡ 1 × m ≡ 0)
 
-_gaPrRel?_ : (n m : ℕ) → Dec (n gaPrRel m)
-n gaPrRel? m = (n ≟ m) ⊎-dec (n ≟ 0 ×-dec m ≟ 1) ⊎-dec (n ≟ 1 ×-dec m ≟ 0)
+_≈_ : GovAction → GovAction → Type
+a ≈ a' = (govActionPriority a) ∼ (govActionPriority a')
+\end{code}
+\begin{code}[hide]
+_∼?_ : (n m : ℕ) → Dec (n ∼ m)
+n ∼? m = n ≟ m ⊎-dec (n ≟ 0 ×-dec m ≟ 1) ⊎-dec (n ≟ 1 ×-dec m ≟ 0)
 
-_≡ᵍᵃ_ : GovAction → GovAction → Type
-a ≡ᵍᵃ a' = (govActionPriority a) gaPrRel (govActionPriority a')
-
-_≡ᵍᵃ?_ : (a a' : GovAction) → Dec (a ≡ᵍᵃ a')
-a ≡ᵍᵃ? a' = (govActionPriority a) gaPrRel? (govActionPriority a')
+_≈?_ : (a a' : GovAction) → Dec (a ≈ a')
+a ≈? a' = (govActionPriority a) ∼? (govActionPriority a')
+\end{code}
+\begin{code}
 
 insertGovAction : GovState → GovActionID × GovActionState → GovState
 insertGovAction [] gaPr = [ gaPr ]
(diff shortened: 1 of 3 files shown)


src/Ledger/Gov.lagda
```diff
@@ -95,6 +95,18 @@ govActionPriority (ChangePParams _)        = 4
 govActionPriority (TreasuryWdrl _)         = 5
 govActionPriority Info                     = 6
 
+_gaPrRel_ : ℕ → ℕ → Type
+n gaPrRel m = (n ≡ m) ⊎ (n ≡ 0 × m ≡ 1) ⊎ (n ≡ 1 × m ≡ 0)
+
+_gaPrRel?_ : (n m : ℕ) → Dec (n gaPrRel m)
+n gaPrRel? m = (n ≟ m) ⊎-dec (n ≟ 0 ×-dec m ≟ 1) ⊎-dec (n ≟ 1 ×-dec m ≟ 0)
+
+_≡ᵍᵃ_ : GovAction → GovAction → Type
```
WhatisRT, line 104: I don't like `≡` here because it's not actually supposed to be an equality. I think `≈` is more appropriate.

src/Ledger/Gov.lagda
```diff
@@ -95,6 +95,18 @@ govActionPriority (ChangePParams _)        = 4
 govActionPriority (TreasuryWdrl _)         = 5
 govActionPriority Info                     = 6
 
+_gaPrRel_ : ℕ → ℕ → Type
+n gaPrRel m = (n ≡ m) ⊎ (n ≡ 0 × m ≡ 1) ⊎ (n ≡ 1 × m ≡ 0)
+
+_gaPrRel?_ : (n m : ℕ) → Dec (n gaPrRel m)
```
WhatisRT, line 101: Decision procedures shouldn't go into the visible part of the PDF.

src/Ledger/Gov.lagda
```diff
@@ -95,6 +95,18 @@ govActionPriority (ChangePParams _)        = 4
 govActionPriority (TreasuryWdrl _)         = 5
 govActionPriority Info                     = 6
 
+_gaPrRel_ : ℕ → ℕ → Type
```
WhatisRT, line 98: I don't like this name. It's not descriptive and you already need to know what it is to have an idea what the name could mean. The name either has to be some long self-explanatory name or it could just be a variation on `≈` together with an explanation in the prose (which I prefer).

src/Ledger/Gov.lagda
```diff
@@ -169,8 +181,10 @@ hasParentE e aid a = case getHashES e a of
    (just id)  → id ≡ aid
 
 hasParent : EnactState → GovState → (a : GovAction) → NeedsHash a → Type
-hasParent e s a aid with getHash aid
-... | just aid' = hasParentE e aid' a ⊎ Any (λ x → proj₁ x ≡ aid') s
+hasParent e s a aid with getHash {a = a} aid
```
WhatisRT, line 184: Is the `{a = a}` necessary?

WhatisRT (changes requested):

src/Ledger/Gov.lagda
williamdemeo: Agreed.  Will rename this `_≈_` (with no decoration since there's no conflict/ambiguity in `Gov` module).  Will also rename `_≡ᵍᵃ?_` to `_≈?_`.

src/Ledger/Gov.lagda
williamdemeo: I will rename the relation `_∼_` and will add the following explanation to the prose:

"Additionally, \govActionPriority assigns a priority to the various governance action types.  This is useful for ordering lists of governance actions as well as grouping governance actions according to its constructor (\NoConfidence, \UpdateCommittee, \NewConstitution, etc.).  In particular, the relations \AgdaFunction{_∼_} and \AgdaFunction{_≈_} defined in Figure~\ref{defs:enactable} are used for determining whether two actions are of the same ``kind'' in the following sense: either the actions arise from the same constructor, or one action is a \NoConfidence and the other is an \UpdateCommittee action."

src/Ledger/Gov.lagda
williamdemeo: Probably not.  I'll remove it.

src/latex/agda-latex-macros.sty
```diff
@@ -273,22 +273,23 @@
...
 \newcommand{\gid}{\AgdaField{gid}\xspace}
 \newcommand{\GovAction}{\AgdaDatatype{GovAction}\xspace}
+\newcommand{\govActionDeposit}{\AgdaField{govActionDeposit}\xspace}
 \newcommand{\GovActionID}{\AgdaDatatype{GovActionID}\xspace}
 \newcommand{\GovActionIDFunction}{\AgdaFunction{GovActionID}\xspace}
-\newcommand{\GovActionState}{\AgdaRecord{GovActionState}\xspace}
-\newcommand{\govActionDeposit}{\AgdaField{govActionDeposit}\xspace}
-\newcommand{\GovernanceGroup}{\AgdaInductiveConstructor{GovernanceGroup}\xspace}
 \newcommand{\govActionLifetime}{\AgdaField{govActionLifetime}\xspace}
+\newcommand{\govActionPriority}{\AgdaFunction{govActionPriority}\xspace}
```
williamdemeo, line 280: Only this line is new, but this section was getting out of order.  It's useful to keep them in alphabetical order as this makes it much easier to spot duplicates and quickly check whether a macro for a given Agda keyword has been defined yet.

WhatisRT (approved):
LGTM

I won't wait with merging this since I think it's somewhat unlikely that conformance tests will spot this and it's probably not that useful to explicitly look for.

PR #555, opened 2024-08-27
Parallelise workflow

Lucsanszky:
# Description

As the title suggests. Broke up the workflow into multiple jobs as this is how parallelisation is achieved. This necessitated the upload of `docs/` as artifacts and merge downloading them, as different jobs have different runners with separate working directories. 

# Checklist

- [x] Self-reviewed the diff


WhatisRT (approved):
I don't really understand some of the things going on here, but the things I do look good

Issue #467, opened 2024-06-13
Prove preservation of value for `CHAIN`


WhatisRT:
We have it for `UTXO`, but we need it for `CHAIN`. Here are a bunch of intermediate steps:

- [x] Improve `UTXO` preservation. In particular, `withdrawals` need to be included in ~`produced`~ `consumed` which will make the proof and the statement more complicated (because now for `UTXO` value actually increases by `withdrawals`).
- [ ] Prove preservation for other `CHAIN` dependencies.
   - [x] `CERTS`. The value in `CertState` (i.e. `HasCoin` `CertState`) is the money contained in reward accounts. That value decreases by `withdrawals`.
   - [x] `LEDGER`. which is just the combination of `UTXO` and `CERTS` (`GOV` does not move any funds).
   - [ ] `EPOCH`. The `HasCoin` for `EpochState` should be the sum of whats in `acnt` and `ls`.
   - [ ] `NEWEPOCH`. `HasCoin` for `NewEpochState` is the contents of `ru` in addition to `epochState`. Note that this essentially just means that `ru` sums up to zero.
- [ ] Combine everything to prove preservation for `CHAIN`.

PR #567, opened 2024-09-12
Prove preservation of value for CHAIN

williamdemeo:
# Description

Closes issue #467 

# Checklist

- [ ] Self-reviewed the diff


williamdemeo:
[Repost of @WhatisRT's Slack message]

"[We] need to change the statement for preservation to not add the withdrawals if isValid = false for the transaction.

For example, in the Alonzo spec in Appendix C we define a function φ that handles the branching on that.

We use it for mint there, but it probably should have included the withdrawals as well (i.e. I think the proof there is wrong for not including that)."

williamdemeo:
This is split up into smaller PRs.

Issue #571, opened 2024-09-16
Replace usage of `minUTxOValue` with `coinsPerUTxOByte`


lehins:
It was brought up by the audit that current implementation of `paramsWellFormed` is incorrect https://github.com/IntersectMBO/formal-ledger-specifications/blob/2e29cba60e5bbe4bbfa835ba73b976dd993b6bf6/src/Ledger/PParams.lagda#L102-L107

Instead of `minUTxOValue` in it should be using `coinsPerUTxOByte`, because `minUTxOValue` has been removed back in Alonzo era.

The only ticket that I could find ab out this is: https://github.com/IntersectMBO/formal-ledger-specifications/issues/162, which I think should have been taken care of properly and `minUTxOValue` should have been removed. It would have prevented it from being used in the current Conway specification.

Quick repo search also revealed that it `minUTxOValue` is also still used in the `Utxo` rule https://github.com/IntersectMBO/formal-ledger-specifications/blob/76ad4e9ba6ff84d7181d74e5e2dfc797ea45af4e/src/Ledger/Utxo/Haskell.agda#L104 which is also quite wrong

WhatisRT:
Yeah, there's a certain tension between historical stuff and new things. At some point we need to give in and deal with this properly, but it's best to do that as late as possible so that we don't need to duplicate a whole lot of work. And we have arguably more important things to work on still.

In this specific case I think replacing `minUTxOValue` with `coinsPerUTxOByte` everywhere is fine, and I guess we could even remove it. We'll just have to add it back later when we do previous eras.

PR #577, opened 2024-09-23
Replace minUTxOValue with coinsPerUTxOByte

williamdemeo:
# Description

This closes issue #571.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Utxo.lagda
+++ b/src/Ledger/Utxo.lagda
@@ -454,7 +454,7 @@ data _⊢_⇀⦇_,UTXO⦈_ where
     ∙ coin mint ≡ 0                          ∙ txsize ≤ maxTxSize pp
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject (utxoEntrySize txout * minUTxOValue pp) ≤ᵗ getValueʰ txout
+        inject (utxoEntrySize txout * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
         serSize (getValueʰ txout) ≤ maxValSize pp
     ∙ ∀[ (a , _) ∈ range txoutsʰ ]
--- a/src/Ledger/Utxo/Haskell.agda
+++ b/src/Ledger/Utxo/Haskell.agda
@@ -101,7 +101,7 @@ data _⊢_⇀⦇_,UTXO⦈_ : UTxOEnv → UTxOState → Tx → UTxOState → Type
     ∙ coin mint ≡ 0                          ∙ txsize ≤ maxTxSize pp
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject (utxoEntrySize txout * minUTxOValue pp) ≤ᵗ getValueʰ txout
+        inject (utxoEntrySize txout * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
         serSize (getValueʰ txout) ≤ maxValSize pp
     ∙ ∀[ (a , _) ∈ range txoutsʰ ]
--- a/src/ScriptVerification/HelloWorld.agda
+++ b/src/ScriptVerification/HelloWorld.agda
@@ -150,7 +150,8 @@ opaque
   succeedExample : ComputationResult String UTxOState
   succeedExample = UTXO-step initEnv ⟦ initState , 0 , ∅ , 0 ⟧ᵘ  succeedTx
 
-  _ : isSuccess succeedExample ≡ true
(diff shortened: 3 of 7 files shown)


src/Ledger/Utxo/Haskell.agda
```diff
@@ -101,7 +101,7 @@ data _⊢_⇀⦇_,UTXO⦈_ : UTxOEnv → UTxOState → Tx → UTxOState → Type
     ∙ coin mint ≡ 0                          ∙ txsize ≤ maxTxSize pp
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject (utxoEntrySize txout * minUTxOValue pp) ≤ᵗ getValueʰ txout
+        inject (utxoEntrySize txout * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
```
lehins, line 104: Ditto

src/Ledger/Utxo.lagda
```diff
@@ -454,7 +454,7 @@ data _⊢_⇀⦇_,UTXO⦈_ where
     ∙ coin mint ≡ 0                          ∙ txsize ≤ maxTxSize pp
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject (utxoEntrySize txout * minUTxOValue pp) ≤ᵗ getValueʰ txout
+        inject (utxoEntrySize txout * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
```
lehins, line 457: As I pointed out in the ticket this is also wrong. @williamdemeo, I am not sure if you plan to fix it in a separate PR, but the correct functionality is described in Babbage spec: https://github.com/IntersectMBO/cardano-ledger/blob/8d16d698dd499fd171845da16611c7c7026c1754/eras/babbage/formal-spec/utxo.tex#L194

On the up side, this is already closer to truth than how it was prior to PR :grin: 

For reference here is ledger implementation: https://github.com/IntersectMBO/cardano-ledger/blob/8d16d698dd499fd171845da16611c7c7026c1754/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs#L595-L618

src/ScriptVerification/HelloWorld.agda
```diff
@@ -150,7 +150,8 @@ opaque
   succeedExample : ComputationResult String UTxOState
   succeedExample = UTXO-step initEnv ⟦ initState , 0 , ∅ , 0 ⟧ᵘ  succeedTx
 
-  _ : isSuccess succeedExample ≡ true
+  --  This is now false since `minUTxOValue` was replaced with `coinsPerUTxOByte` (= 4310 in `ScriptVerification.Lib`)
```
WhatisRT, line 153: I think it' a bad idea to break these tests. It'd be better to set `coinsPerUTxOByte = 0` with a note to adjust that there.

src/Ledger/Utxo.lagda
WhatisRT: Yeah, I think that needs to be another PR, now is not the time to worry about serialization issues.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Add `curTreasury` field to transactions
 - Compute the voting stake distribution
 - Add deposit amount to `dereg` certificate
+- Replace minUTxOValue with coinsPerUTxOByte
```
WhatisRT, line 23:
```suggestion
- Replace `minUTxOValue` with `coinsPerUTxOByte`
```

WhatisRT (approved):
Just some minor changes, you can merge it once they are done

src/Ledger/Utxo.lagda
lehins: It is not about serialization, but rather about the formula that is being used for computing the min value.
We can deal with size of TxOut in the same way we deal with hashes in conformance. We can just pass it when translating it to MAlonzo

WhatisRT: Yeah, that's what I mean. Implementing that is a bunch of work and there are already too many high-priority things.

lehins: Yeah, seems like we are short staffed on the formal methods side.

src/ScriptVerification/Lib.agda
```diff
@@ -34,7 +34,10 @@ createEnv s = record { slot = s ; treasury = 0 ;
                                ; minUTxOValue = 0
                                ; poolDeposit = 500000000 -- lovelace
                                ; keyDeposit = 500000000 -- lovelace
-                               ; coinsPerUTxOByte = 4310 --lovelace
+                               ; coinsPerUTxOByte = 0  --lovelace
+                               -- ^^^ was 4310, but that value resulted in failed tests once
```
lehins, line 38: Sounds like tests need to be fixed. @williamdemeo, could you please create a ticket about it so we don't forget to fix 'em.

lehins (approved):
LGTM :+1:

WhatisRT (approved):

PR #580, opened 2024-09-24
Reorganize conformance code

### LINKED ISSUE #579: Make a conformance version of every rule
So far we only have the conformance versions of the `UTXO` and `CERTS` rule. We need to make conformance versions of all the rules, because all the rules that call `UTXO` and `CERTS` are using the original types which are incompatible with the modified types that are used for conformance testing.

I think it's useful to think of the conformance version as a separate parallel era to Conway and use the same approach as we would for adding a new era to the spec, which means copying the code from the original Conway implementation and then modifying the files to make them compatible with the Haskell implementation.

Soupstraw:
# Description

This PR reorganizes the conformance code so that all of the code lives in the `src/Ledger/Conway/Conformance` directory. I also duplicated all of the rules so that the conformance code has all the types up to `CHAIN`. The types used for proofs are no longer exposed via MAlonzo exports.

closes #579 

# Checklist

- [ ] Self-reviewed the diff


WhatisRT (commented):
Awesome! There are a few changes to the structure that I think are worth making, but overall this looks great!

- I don't think we need any of the `Ledger.Conway.Conformance` modules to be literate Agda files, we probably won't compile those into a pdf. Removing all that stuff also should help restore some readability.
- There are some files that will likely be completely identical between the two forever, for example `Address` and probably `Transaction`. Having duplication here comes at an actual cost, which is that we'd have to write conversion functions for everything to be able to prove equivalence. So I think for those files it'd be better to just import them.
- The `Foreign` subtree probably makes more sense one level higher, i.e. in `Ledger.Conway`, especially given the previous point.

Soupstraw:
@WhatisRT I agree with your points, but I don't want to spend too much time on these tasks right now. Maybe we can make some issues about these so someone can fix them later?

WhatisRT (approved):
Yeah, makes sense

PR #587, opened 2024-10-02
Separate precondition for updateCertDeposits instead of making it partial

UlfNorell:


williamdemeo (approved):
This all looks great.  Thank you, Ulf!  There are a few mods I will want to make and I still need to fill in a (new) proof of the `gmsc` theorem, but I can do that after merging this into my old 512 branch.

PR #589, opened 2024-10-07
Removed deposit updating from Haskell version of UTXO

Soupstraw:
# Description

This PR modifies the alternative spec so that the deposit update happens in the LEDGER rule instead of the UTXO rule like in the implementation. It also makes the `isSigned` check always succeed in conformance testing.

# Checklist

- [ ] Self-reviewed the diff


WhatisRT (approved):
LGTM

PR #592, opened 2024-10-09
Factor out the set theory into a separate library

WhatisRT:
# Description

This introduces a dependency of the new [agda-sets](https://github.com/input-output-hk/agda-sets). I also did some refactoring as part of this, which is where most changes come from.

I've requested reviews from everyone who I think might be interested, but feel free to ignore it or just have a quick look.

# Checklist

- [x] Self-reviewed the diff


javierdiaz72:
I'm just curious regarding https://github.com/input-output-hk/agda-sets/blob/6dd8cebbd46090b5909c13a1e8d609a44f02a5d3/src/Axiom/Set/Predicates.agda#L1: Why is `--type-in-type` needed here? As I understand it (with my still limited experience in Agda), this flag renders the logic inconsistent.

javierdiaz72 (approved):

WhatisRT:
Yes, `--type-in-type` is indeed inconsistent. It's necessary there only because of stupid reasons: In `Theory`, the field `Set` has type `Type ℓ → Type ℓ`, but `λ A → A → Type` has type `Type ℓ → Type (ℓ ⊔ 1)`.

However, I've made this issue: https://github.com/input-output-hk/agda-sets/issues/2 Resolving that will allow dropping `--type-in-type`. So I'd say that in this case the inconsistency is something not to worry about.

default.nix
```diff
@@ -53,7 +53,23 @@ let
...
   };
 
-  deps = [ agdaStdlib agdaStdlibClasses agdaStdlibMeta ];
+  agdaSets = customAgda.agdaPackages.mkDerivation {
+    inherit (locales) LANG LC_ALL LOCALE_ARCHIVE;
+    pname = "agda-sets";
+    version = "2.1.1";
+    src = fetchFromGitHub {
+      repo = "agda-sets";
+      owner = "input-output-hk";
```
williamdemeo, line 62: Just curious why the `agda-sets` repo is owned by input-output-hk, while the formal spec repo is with intersect.

williamdemeo (approved):
It seems fine.  The only reservation I have is that I often add things, or just work inside, the set theory library while working on the formal spec; working inside modules under `Axiom.Set` means faster type-checking, less module overhead, etc, compared to working inside a ledger module.  Now it seems there's one extra hoop to jump through (which you kindly described in `CONTRIBUTING.md`---thanks!) in order to work that way, but I'm sure I'll get used to it.

WhatisRT:
Yeah, sorry for the extra friction, but it's required for being able to use it it other projects. You can of course always add stuff in a module like `Axiom.Set.Ext` and then upstream it when it's done.

default.nix
WhatisRT: I don't think I can make a repository on the Intersect repo, but I'm also not super sure where this repo should live, so I've put it there for now.

Issue #578, opened 2024-09-24
Change the way SPO votes are counted


williamdemeo:
This issue reflects the [corresponding cardano-ledger issue](https://github.com/IntersectMBO/cardano-ledger/issues/4645), and the description below is nearly identical to the description given there.

It has been brought up by a number of community members that the way we currently compute SPO votes is not what everyone expected.
Current implementation is such that for `HardForkInitiation` behavior is exactly the same as for DReps, however for the other three governance actions, that SPOs can vote on, behavior is different for SPOs that did not vote. In other words for these three governance action types:
* `NoConfidence`
* `UpdateCommittee`
* `ParameterUpdate` (when it includes updates to parameter belonging to the security group)

the default vote counted for SPOs that didn't vote is `Abstain`, instead of `No`. The default vote of `No` is the behavior for DReps, while for SPOs it is default only when `HardForkInitiation` is proposed.

Current behavior is specified in the following lines of the spec: https://github.com/IntersectMBO/formal-ledger-specifications/blob/762d3f42ef2876f039fafda1eb7f69d651129261/src/Ledger/Ratify.lagda#L378-L380

The corresponding implementation is [here](https://github.com/IntersectMBO/cardano-ledger/blob/8d16d698dd499fd171845da16611c7c7026c1754/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ratify.hs#L204-L207).

This led to some concerns about inclusion of SPOs in future parameter updates, which is why many have suggested we mimic the behavior of DReps exactly. However, defaulting of SPOs to vote `No` raises other concerns. In particular there is a danger that sensible proposals could be inadvertently blocked by SPOs that are not active enough. Inactivity could be potentially blamed either on indifference or issues with legality for some SPOs.

In order to solve this conundrum we made it the topic of [Ledger Working Group Meeting #4](https://lu.ma/foop6fuh) and had a very productive discussion with some community members and came up with the following ingenious solution:

[...]

PR #594, opened 2024-10-11
Change the way SPO votes are counted

williamdemeo:
# Description

This closes issue #578.

# Checklist

- [X] Self-reviewed the diff


CHANGELOG.md
```diff
@@ -44,7 +45,6 @@
 - Fix that `RATIFY` wasn't total
 - Properly implement treasury withdrawals
 - Remove unused hot credentials at the epoch boundary
-- Implement proper vote counting for SPOs
```
williamdemeo, line 47: @WhatisRT I'm not sure what is the purpose of the WIP section of the CHANGELOG. I moved the issue this PR addresses from the 0.9 section to the WIP section, but I'm not sure if that's what I was meant to do.

WhatisRT (dismissed):
Sorry, what actually needs to be done here is a bit more complicated. `DState.voteDelegs` needs to be available to `actualSPOVotes`. Then, if you have a `credVoter SPO c`, you need to lookup `c` in `voteDelegs` and the result of that lookup is the default vote.

williamdemeo:
@WhatisRT @Lucsanszky I've revised the SPO default vote function.  I *think* it's correct now, but let me know if you notice anything wrong with it.  Thanks!

Lucsanszky:
@WhatisRT @williamdemeo conformance is still failing for me. I had a better look at the ledger implementation (here's my PR which updates the SPO vote counting: https://github.com/IntersectMBO/cardano-ledger/pull/4659) and the spec. To me it seems like that they indeed don't really match. In the implementation, we check if the SPO's reward address is delegated to an `AlwaysAbstain` or an `AlwaysNoConfidence` DRep. See: https://github.com/IntersectMBO/cardano-ledger/pull/4659/files#diff-d479e8ce99ea392464360b8cea0bb7c42e04863929fd3f4ae4c92dd5742e535cR222-R223
That is, on top of `DState.voteDelegs`, `PState.pools` needs to be available to `actualSPOVotes` as well, so we can first lookup the pool parameters of the given pool in `pools`, grab the reward address and look that up in `voteDelegs`. Unless:
1. My implementation is wrong.
2. I misunderstand something here as I'm not too familiar with Agda and the spec and somehow the `c` in `credVoter SPO c` is the reward address of the given pool.

williamdemeo:
@Lucsanszky You description seems reasonable to me.  I'll go ahead and try to change the Agda code to match your understanding of how the policy should be implemented. The changes will appear in the commit that follows the "finally got TeX to compile..." commit.

williamdemeo:
@Lucsanszky I believe `Ratify.lagda` now agrees with the implementation as you describe it above.  Thanks for the feedback.  Let me know how the conformance tests go.

Lucsanszky:
At a first glance, this looks good! Unfortunately, conformance is still failing but that could be because of something else (and I have an idea why). Anyways, I'll wait for @WhatisRT's approval. Thanks!

Lucsanszky:
> At a first glance, this looks good! Unfortunately, conformance is still failing but that could be because of something else (and I have an idea why). Anyways, I'll wait for @WhatisRT's approval. Thanks!

Okay, conformance passes. :) I opened a [PR](https://github.com/IntersectMBO/formal-ledger-specifications/pull/596) which also adds your changes to the conformance files, feel free to cherry pick the commits. Also, I took the liberty to slightly change your specification and opted not to pass in the complete `CertState` to `RatifyEnv`. I think it's better to be more specific with what we need and also helps with the conformance tests on the `cardano-ledger` side.

williamdemeo:
Looks good.  Thanks @Lucsanszky!  I've merged your mods into this PR, so you can close your PR.

Lucsanszky (approved):
LGTM, thanks! :+1:

src/Ledger/Conway/Conformance/Ratify.lagda
```diff
@@ -272,18 +274,35 @@ actualVotes Γ pparams cc ga votes
...
-
-  actualCCVote : Credential → Epoch → Vote
-  actualCCVote c e = case getCCHotCred (c , e) of
+    (true , just (just c'))  → just c'
+    _                        → nothing -- expired, no hot key or resigned
 \end{code}
 \begin{code}[hide]
-    λ where
+  getPoolParams : Credential → Maybe PoolParams
+  getPoolParams c = case lookupᵐ? stakeDelegs c of λ where  nothing → nothing
```
WhatisRT, line 282: This lookup is incorrect, you just need to lookup `c` in `pools`.

src/Ledger/Conway/Conformance/Ratify.lagda
```diff
@@ -272,18 +274,35 @@ actualVotes Γ pparams cc ga votes
...
+      _                       → Vote.no
+  SPODefaultVote _ (credVoter SPO c) = case getPoolParams c of λ where
+    nothing → Vote.no
+    (just p) → case lookupᵐ? delegatees (PoolParams.rewardAddr p) of λ where
+      (just abstainRep)  → Vote.abstain
+      _                  → Vote.no
+  SPODefaultVote _ _ = Vote.no
+
+  actualCCVote : Credential → Epoch → Vote
+  actualCCVote c e = case getCCHotCred (c , e) of λ where
```
WhatisRT, line 303: Why did you un-hide the `λ where`? It just makes the PDF more confusing to people who don't know Agda.

WhatisRT:
Sorry, this is incorrect, and the implementation is probably too. Also, looking at this it would make much more sense to just use the credential of the stake pool itself instead of the reward account. Why should voting & setting the default for the vote require 2 different credentials?

src/Ledger/Conway/Conformance/Ratify.lagda
williamdemeo: For some reason, the only way I could get the TeX to compile without the "missing parameter l" error was to expose the `λ where` here and above.  I tried many permutations of exposing and hiding various parts of the code, and the one I ended up with was the closest to what we want while still getting the pdf to compile without error.

WhatisRT: It's probably just this bug again: https://github.com/agda/agda/issues/7290

You probably just need to indent the clauses of the `λ where` a bit more.

src/Ledger/Conway/Conformance/Ratify.lagda
williamdemeo: That's good to know, thanks.  If there is a particular place in a prior version of the spec or a CIP that would have helped me understand this better so that I wouldn't have made this mistake, please advise.

WhatisRT: Not really. It's just that `lookupᵐ? pools c` is the `PoolParams` associated to `c` and contains `c`'s `RewardAddr`,  which is what you want. As the code is written here, it gives you the `PoolParams` of the pool that `c` is delegating to which could be any random pool.

williamdemeo: I see.  Thanks for the explanation.  Very helpful. 👍🏼

src/Ledger/Conway/Conformance/Ratify.lagda
williamdemeo: Yes, it's exactly that bug.  I tried indenting all sorts of ways, but I'll try again to get it to compile while hiding all the `λ where`s and I'll also fix the bug mentioned above, of course.

lehins:
> looking at this it would make much more sense to just use the credential of the stake pool itself instead of the reward account.

It is too late to change implementation. `cardano-node-10.x` has already been released, which means at this point any discrepancies will have to be changed in the spec, instead of implementation.

> Why should voting & setting the default for the vote require 2 different credentials?

Why would Stake Pool private key be used for both managing a pool and registering a stake credential?

Issue #511, opened 2024-07-15
Should we only allow DRep delegations to existing DReps?


WhatisRT:
We do this for stake pools, so it would be consistent.

lehins:
Relevant ticket on the ledger implementation side: https://github.com/IntersectMBO/cardano-ledger/issues/4598

PR #597, opened 2024-10-22
Only allow delegations to existing DReps

williamdemeo:
# Description

Will close issue #511.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Certs.lagda
```diff
@@ -405,7 +404,7 @@ data _⊢_⇀⦇_,CERT⦈_ where
 \end{code}
 \begin{code}
   CERT-deleg :
-    ∙ ⟦ pp , PState.pools stᵖ ⟧ᵈᵉ ⊢ stᵈ ⇀⦇ dCert ,DELEG⦈ stᵈ'
+    ∙ ⟦ pp , PState.pools stᵖ , DState.voteDelegs stᵈ ⟧ᵈᵉ ⊢ stᵈ ⇀⦇ dCert ,DELEG⦈ stᵈ'
```
WhatisRT, line 407: This needs to be `GState.dreps`

src/Ledger/Certs.lagda
```diff
@@ -48,7 +48,7 @@ record PoolParams : Type where
 \end{NoConway}
 \begin{code}
 data DCert : Type where
-  delegate    : Credential → Maybe VDeleg → Maybe KeyHash → Coin → DCert
+  delegate    : Credential → Maybe Credential → Maybe KeyHash → Coin → DCert
```
WhatisRT, line 51: You can't change this constructor

src/Ledger/Certs.lagda
```diff
@@ -317,16 +315,17 @@ data _⊢_⇀⦇_,DELEG⦈_ where
   DELEG-delegate : let open PParams pp in
     ∙ (c ∉ dom rwds → d ≡ keyDeposit)
     ∙ (c ∈ dom rwds → d ≡ 0)
+    ∙ mc ∈ mapˢ just (dom dReps) ∪ ❴ nothing ❵
```
WhatisRT, line 318: This is pretty much correct, except that the type needs to be `Maybe VDeleg`. So you also need to map `credVoter DRep` over it.

src/Ledger/Certs.lagda
```diff
@@ -138,12 +138,13 @@ record CertState : Type where
...
-  constructor ⟦_,_⟧ᵈᵉ
+  constructor ⟦_,_,_⟧ᵈᵉ
   field
 \end{code}
 \begin{code}
-    pparams  : PParams
-    pools    : KeyHash ⇀ PoolParams
+    pparams       : PParams
+    pools         : KeyHash ⇀ PoolParams
+    delegatees    : Credential ⇀ Epoch
```
WhatisRT, line 147:
```suggestion
    delegatees    : ℙ Credential
```
Since only the domain is used anyway, this is a bit cleaner

WhatisRT (changes requested):
So what also needs to be added is that when a DRep deregisters, all the delegations to it need to be cleaned up.

The issue is that you need to have access to the delegation mapping, so maybe it's best to do this in `CERT-base`. The easiest option might be to filter out those that don't exist instead of looking at any individual certificates.

In the end, the property `range vDelegs ⊆ map (credVoter DRep) (dom DReps)` always needs to hold. It might be a good idea to make sure this property type checks and to put it into a comment so it can be proven soon.

src/Ledger/Conway/Conformance/Certs.agda
```diff
@@ -57,33 +57,36 @@ updateCertDeposit pp cert deposits
...
+  Γ : CertEnv
+  pp : PParams
+  dep ddep gdep : Deposits
+  vs : List GovVote
+  poolParams : PoolParams
+  wdrls  : RwdAddr ⇀ Coin
+  ccHotKeys : Credential ⇀ Maybe Credential
+  voteDelegs : Credential ⇀ VDeleg
+  stakeDelegs : Credential ⇀ KeyHash
+  rewards : Credential ⇀ Coin
```
WhatisRT, line 89: What happened to the nice alignment?

src/Ledger/Certs.lagda
```diff
@@ -426,20 +426,25 @@ data _⊢_⇀⦇_,CERT⦈_ where
...
-    let open PParams pp
-        refresh         = mapPartial getDRepVote (fromList vs)
-        refreshedDReps  = mapValueRestricted (const (e + drepActivity)) dreps refresh
-        wdrlCreds       = mapˢ stake (dom wdrls)
+  CERT-base : let
+    open PParams pp
+    refresh          = mapPartial getDRepVote (fromList vs)
+    refreshedDReps   = mapValueRestricted (const (e + drepActivity)) dReps refresh
+    wdrlCreds        = mapˢ stake (dom wdrls)
+    freshVoteDelegs  = voteDelegs ∣^ mapˢ (credVoter DRep) (dom refreshedDReps)
```
WhatisRT, line 434: If it fits on the page, I think it's better to inline this. You can also just use `dReps` instead of `refreshedDReps`, since they have the same domain

src/Ledger/Conway/Conformance/Certs.agda
williamdemeo: Was there nice alignment here before?  I didn't remove any in this PR.  Anyway, I'll align this section nicely in the next, hopefully final, commit.

src/Ledger/Certs.lagda
williamdemeo: Yes, the domain is the same, but do you think using `refreshedDReps` makes the point of the restriction to existing DReps clearer?  Well, I guess not, since the refresh doesn't change the set of existing DReps.  Okay, I'll change it back to `dReps`.

williamdemeo: Inlining makes the code run far outside the figure box, so I'll change back to an extra variable definition, `validVoteDelegs`, for the restricted `voteDelegs` (restricted to registered DReps).

src/Ledger/Conway/Conformance/Certs.agda
```diff
@@ -175,7 +179,7 @@ data _⊢_⇀⦇_,CERTBASE⦈_ : CertEnv → CertState → ⊤ → CertState →
       ⟦ e , pp , vs , wdrls ⟧ᶜ ⊢
       ⟦ ⟦ voteDelegs , stakeDelegs , rewards , ddep ⟧ᵈ
       , stᵖ
-      , ⟦ dreps , ccHotKeys , gdep ⟧ᵛ
+      , ⟦ dReps , ccHotKeys , gdep ⟧ᵛ
       ⟧ᶜˢ
       ⇀⦇ _ ,CERTBASE⦈
       ⟦ ⟦ voteDelegs , stakeDelegs , constMap wdrlCreds 0 ∪ˡ rewards , ddep ⟧ᵈ
```
WhatisRT, line 185: The removal of delegations is still missing from here

williamdemeo: I forgot about the duplicate code. 🤦🏼

Will fix this today.

WhatisRT (approved):
LGTM

PR #598, opened 2024-10-22
Rearranging some of the duplicated code (#582)

### LINKED ISSUE #582: Rearranging some of the duplicate code
- [ ] None of the `Ledger.Conway.Conformance` modules need to be literate Agda files, we probably won't compile those into a pdf. Removing all that stuff also should help restore some readability.
- [ ] There are some files that will likely be completely identical between the two forever, for example `Address` and probably `Transaction`. Having duplication here comes at an actual cost, which is that we'd have to write conversion functions for everything to be able to prove equivalence. So I think for those files it'd be better to just import them.
- [ ] The `Foreign` subtree probably makes more sense one level higher, i.e. in `Ledger.Conway`, especially given the previous point.

_Originally posted by @WhatisRT in https://github.com/IntersectMBO/formal-ledger-specifications/pull/580#pullrequestreview-2328357124_

UlfNorell:
Fixes #582


WhatisRT (approved):
Great, thank you! The removal of the duplicate `GlobalConstants` and `TransactionStructure` seems to already have caught a difference!

@Soupstraw can you check that this does not break any conformance tests?

src/Makefile
```diff
@@ -63,7 +63,7 @@ $(CONWAY_TEX_FILES): $(CONWAY_TEX_DIR)/%.tex: %.lagda
 		$(PYTHON) $(HL_SCRIPT) $@ $@ $(HL_LIST); \
 	fi
 
-define latexToPdf =
+define latexToPdf
```
WhatisRT, line 66: I'm surprised that this ever worked! Did this cause a warning or something?

UlfNorell: For me the make target just silently did nothing. It took some experimenting to figure out that this was the cause.

Lucsanszky:
> Great, thank you! The removal of the duplicate `GlobalConstants` and `TransactionStructure` seems to already have caught a difference!
> 
> @Soupstraw can you check that this does not break any conformance tests?

As Joosep is currently on holiday, I'll have a look soon.

Lucsanszky (approved):
Great, it seems like the changes didn't break anything. :+1:

src/Makefile
WhatisRT: Oh, are you on a mac? I remember having this issue, and my thought was that `define` is probably just not supported there.

UlfNorell: Yes, on mac, and also not running in nix, so it might be that my make is old? The docs are conflicted:

This one has the `=` sign:

https://www.gnu.org/software/make/manual/make.html#Canned-Recipes

and this one doesn't:
https://ftp.gnu.org/old-gnu/Manuals/make/html_node/make_71.html

Hopefully removing the `=` sign doesn't break it for other people!

WhatisRT: Yeah, that was the context in which I didn't work for me as well. My thought was that macos might come with some non-GNU `make` that just doesn't support a non-standard feature. I've definitely had that with other standard tools before. But my `make --help` says `Report bugs to <bug-make@gnu.org>` so it may indeed be a version difference. It always worked with `nix` because that just gets all its tools from `nix` anyway.

Anyway, it's nice that it now also works outside of `nix`. And since it didn't break CI it probably doesn't break anything in `nix` elsewhere.

Issue #545, opened 2024-08-15
`CERTS` needs to do the base case first


WhatisRT:
`ReflexiveTransitiveClosure` does the base case last, but it should be done first.

PR #604, opened 2024-11-04
CERTS should do the base case first

williamdemeo:
# Description

This closes issue #545.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Certs.lagda
```diff
@@ -318,15 +317,15 @@ data _⊢_⇀⦇_,DELEG⦈_ where
         fromList ( nothing ∷ just abstainRep ∷ just noConfidenceRep ∷ [] )
     ∙ mkh ∈ mapˢ just (dom pools) ∪ ❴ nothing ❵
       ────────────────────────────────
-      ⟦ pp , pools , delegatees ⟧ᵈᵉ ⊢
-        ⟦ vDelegs , sDelegs , rwds ⟧ᵈ ⇀⦇ delegate c mv mkh d ,DELEG⦈
+      ⟦ pp , pools , delegatees ⟧ᵈᵉ ⊢ ⟦ vDelegs , sDelegs , rwds ⟧ᵈ
+        ⇀⦇ delegate c mv mkh d ,DELEG⦈
```
williamdemeo, line 321: Fix pdf figure where this line gets stretched and ugly.

src/Ledger/Certs.lagda
```diff
@@ -430,7 +429,8 @@ data _⊢_⇀⦇_,CERTBASE⦈_ where
     refresh          = mapPartial getDRepVote (fromList vs)
     refreshedDReps   = mapValueRestricted (const (e + drepActivity)) dReps refresh
     wdrlCreds        = mapˢ stake (dom wdrls)
-    validVoteDelegs  = voteDelegs ∣^ (mapˢ (credVoter DRep) (dom dReps) ∪ fromList (noConfidenceRep ∷ abstainRep ∷ []))
+    validVoteDelegs  = voteDelegs ∣^ (  mapˢ (credVoter DRep) (dom dReps)
+                                        ∪ fromList (noConfidenceRep ∷ abstainRep ∷ []) )
```
williamdemeo, line 433: Fix pdf figure box overrun.

src/Ledger/Certs.lagda
```diff
@@ -264,42 +264,41 @@ constitutional committee.
...
 \begin{code}
-  _⊢_⇀⦇_,CERTBASE⦈_  : CertEnv → CertState → ⊤ → CertState → Type
-\end{code}
-\begin{code}
-_⊢_⇀⦇_,CERTS⦈_     : CertEnv → CertState → List DCert → CertState → Type
-_⊢_⇀⦇_,CERTS⦈_ = ReflexiveTransitiveClosureᵇ _⊢_⇀⦇_,CERTBASE⦈_ _⊢_⇀⦇_,CERT⦈_
+    _⊢_⇀⦇_,CERTBASE⦈_  : CertEnv → CertState → ⊤ → CertState → Type
+
+_⊢_⇀⦇_,CERTS⦈_       : CertEnv → CertState → List DCert → CertState → Type
+_⊢_⇀⦇_,CERTS⦈_ = ReflexiveTransitiveClosure _⊢_⇀⦇_,CERT⦈_
```
WhatisRT, line 299: It would be much nicer if you just make an abstraction here that replaces `ReflexiveTransitiveClosureᵇ`. Then you wouldn't have to modify `LEDGER` at all, which is better because it keeps the concerns of the different STSs separate.

WhatisRT: It also means that the proof that this new closure operator preserves `Computational` would be general, whereas right now it's intertwined with the `Computational` proof of `LEDGER`.

williamdemeo: @WhatisRT I've tried a number of ways to replace `ReflexiveTransitiveClosureᵇ` with an abstraction that, in this instance, does `CERTBASE` for the base case and then `CERTS` for the `sigs` case, while also not having the empty list appear in the `CERTBASE` case... but I'm not able to get the completeness proofs to work with the various attempts I've made.  Could you be more specific about what you had in mind by "just make an abstraction"?

WhatisRT: Just do exactly what you did, except in general:
```
module _ (_⊢_⇀⟦_⟧ᵇ_ : C → S → ⊤ → S → Type) (_⊢_⇀⟦_⟧_ : C → S → Sig → S → Type) where
  data ReflexiveTransitiveClosureᵇ' : C → S → List Sig → S → Type where
      RTC :
          ∙ Γ ⊢ s ⇀⟦ _ ⟧ᵇ s'
          ∙ ReflexiveTransitiveClosure _⊢_⇀⟦_⟧_ Γ s' sigs s''
          ───────────────────────────────────────
          ReflexiveTransitiveClosureᵇ' Γ s sigs s''
```
Then, whatever was necessary to prove `Computational` for `LEDGER` also applies to this relation, and all you need to do is keep the original definitions of `CERTS` and `LEDGER`, but just use this closure instead. If it turns out that the original `ReflexiveTransitiveClosureᵇ` is unused, it's best to just call this relation `ReflexiveTransitiveClosureᵇ`.

williamdemeo: I see.  That makes sense.  Thanks.

I've tried it and I'm still having trouble getting the computation proofs to go through... I'll keep trying...

src/Interface/STS.agda
```diff
@@ -39,7 +39,7 @@ private
 
 -- small-step to big-step transformer
 
-module _ (_⊢_⇀⟦_⟧ᵇ_ : C → S → ⊤ → S → Type) (_⊢_⇀⟦_⟧_ : C → S → Sig → S → Type) where
+module _ {_⊢_⇀⟦_⟧ᵇ_ : C → S → ⊤ → S → Type} {_⊢_⇀⟦_⟧_ : C → S → Sig → S → Type} where
```
williamdemeo, line 42: If these parameters are explicit, then the underscores are not filled in properly by Agda in the *Agda* status buffer and it makes statements of goals, errors, etc. hard to comprehend.  When the parameters are implicit, the underscores work as desired.  Specifying the parameters now requires curly braces, but I think that's a small price to pay for making the underscores work as expected.

src/Ledger/Certs.lagda
williamdemeo: Got em to proved!🍾🎆

WhatisRT (approved):
LGTM

PR #612, opened 2024-11-18
Add a proper `EnterpriseAddr` type

WhatisRT:
# Description

Replace the `Maybe` in the `stake` field of `BaseAddr` with a proper `EnterpriseAddr` type.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Address.lagda
+++ b/src/Ledger/Address.lagda
@@ -6,6 +6,7 @@ open import Ledger.Prelude
 
 open import Tactic.Derive.DecEq
 open import Tactic.Derive.Show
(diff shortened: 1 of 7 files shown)


lehins:
This approach does not match Haskell implementation. Less discrepancies we have the better.
Instead we just need to add a comment here saying that `Nothing` means an "Enterprise" address:

https://github.com/IntersectMBO/formal-ledger-specifications/blob/f93035b552d9778cc7c96dc24a40054f4073a6e1/src/Ledger/Address.lagda#L62

Issue #585, opened 2024-09-30
Prove an equivalence between the two ledgers


WhatisRT:
We have two models right now, one under `Ledger` and one under `Ledger.Conway.Conformance`. They differ by some logic being moved from `CERTS` into `UTXO` but are otherwise identical. This means that there should be a theorem of the form
```
Γ L.⊢ s ⇀⦇ tx ,LEDGER⦈ s' ⇔ Γ C.⊢ s ⇀⦇ tx ,LEDGER⦈ s'
```
except that we need to translate some of the types being involved (otherwise this won't type check). In the end, we need to prove this for `CHAIN`, but that should be trivial once we have it for `LEDGER`.

Our current strategy for maintenance is based on this being proven, otherwise these could drift out of sync. So the sooner this is done the better.

See also #512, which will actually make this provable once finished and #525.

Blocked by #512 and it would be a good idea to do #582 first.

UlfNorell:
Is there a precise(ish) description of what the difference should be? I'm expecting there will be some places where `Ledger` has moved since `Conformance` was forked where the correct thing is to just update `Conformance` to match.

WhatisRT:
If I'm not forgetting anything, the only actual difference should be in how the deposits are handled. Any other difference is probably unintentional drift between the two implementations.

I think you're right, it's probably more likely that in that case `Conformance` needs an update. I think in that case it'd be best to make a small separate PR just for that update so that the PR for this issue includes no logic changes after everything is rebased. That way the question 'which one is correct' becomes disentangled from this work.

UlfNorell:
Getting deeper into the proof, I'm getting confused about how the conformance rules are supposed to work (and consequently what the precise equivalence statement should be).

In particular, in the Conformance rules, there are three separate `Deposits` in `LState`: two in `CertState` (in `DState` and `GState`) and one in `UTxOState`. The ones in `CertState` are updated independently in `DELEG` and `GOVCERT` respectively:

https://github.com/IntersectMBO/formal-ledger-specifications/blob/d61bd383e54c7927798b6d2f7dcd946c0ac79384/src/Ledger/Conway/Conformance/Certs.agda#L103-L114

https://github.com/IntersectMBO/formal-ledger-specifications/blob/d61bd383e54c7927798b6d2f7dcd946c0ac79384/src/Ledger/Conway/Conformance/Certs.agda#L127-L135

and never synced up

[...]

PR #614, opened 2024-11-19
Prepare for conformance equivalence

UlfNorell:
Removes some duplicated rules and code from `Conway.Conformance` and fixes a couple of small issues.

Preparation for #585


WhatisRT (approved):
The sanity-checking script is failing CI because `Ledger.Conway.Conformance.Ratify.Properties` is unused, and `Ledger.Conway.Conformance.Ratify` is probably unused as well now, so you'll have to delete those. Otherwise LGTM

Issue #615, opened 2024-11-21
Add legacy registration certificates


Soupstraw:
The Haskell implementation supports legacy staking credential registration certificates that are different from the `delegate` cert in that they do not require the witnessing of the staking credential. 

We should add an additional certificate that registers a staking credential without requiring a witness to align the spec with the implementation.

PR #616, opened 2024-11-21
Added `reg` cert, fixed hashing of `VKey`s

Soupstraw:
# Description

Adds `reg` cert, which behaves similarly to `delegate c nothing nothing d`, except that it doesn't require witnessing. This is the legacy staking address registration certificate that we still want to support in Conway and hopefully we can get rid of in the future.

Adds `HSVKey`, which stores the verification key together with the hash that's provided during translation in conformance testing. This fixes the issue where the value of the VKey itself is used as the hash, which caused conformance tests to fail.

related: https://github.com/IntersectMBO/cardano-ledger/pull/4780
closes https://github.com/IntersectMBO/formal-ledger-specifications/issues/615

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -22,10 +22,6 @@ in {
     nativeBuildInputs = [
       specs.agda
       cabal-install
-      (haskellPackages.ghcWithPackages (pkgs: with pkgs; [
-        specs.ledger.hsExe
-        specs.midnight.hsExe
-      ]))
     ];
   };
 }
--- a/src/Ledger/Certs.lagda
+++ b/src/Ledger/Certs.lagda
@@ -56,16 +56,22 @@ data DCert : Type where
   deregdrep   : Credential → Coin → DCert
   ccreghot    : Credential → Maybe Credential → DCert
 \end{code}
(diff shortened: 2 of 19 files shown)


shell.nix
```diff
@@ -22,10 +22,6 @@ in {
     nativeBuildInputs = [
       specs.agda
       cabal-install
-      (haskellPackages.ghcWithPackages (pkgs: with pkgs; [
-        specs.ledger.hsExe
-        specs.midnight.hsExe
-      ]))
```
williamdemeo, line 28: We don't need this anymore?

williamdemeo (approved):
Looks great.  Nice work, Joosep!  LGTM! 🚀

williamdemeo:
@Soupstraw before merging, could you please add descriptions for each substantial change to the `CHANGELOG.md`?  (especially anything that changes the ledger semantics)

shell.nix
Soupstraw: I don't think it makes sense to build the project when entering `nix-shell`

PR #618, opened 2024-11-25
Reuse ledger code in conformance

UlfNorell:
Removing some more code duplication between Ledger and Conformance to simplify the equivalence proof in #585.


williamdemeo (approved):
This looks great.  LGTM!

PR #622, opened 2024-12-03
Equivalence proof between spec and conformance rules (#585)

### LINKED ISSUE #585: Prove an equivalence between the two ledgers

UlfNorell:
Fixes #585

There are still holes in the proofs about maps, but we are very confident that the lemmas are true. Since the main value of this PR is to force Ledger and Conformance rules to stay in sync, I think this can/should be merged even if the map lemmas are not proven in their entirety.

**ATTN**: @williamdemeo @WhatisRT @Soupstraw I had to add (in b511366) a `(DRepDeposit c , d) ∈ dep` check to `GOVCERT-deregdrep` in Conformance. This check is required by the spec and without it the equivalence doesn't hold.

Here is the final main result:
https://github.com/IntersectMBO/formal-ledger-specifications/blob/cf35c285ae7a1df346480fdf76aa6a233a81fe25/src/Ledger/Conway/Conformance/Equivalence.agda#L420-L431

There is a lot of cleanup that can be done (I focused on getting the proof through, sacrificing lemma naming and proof structure at times), but that can be addressed in a later PR.


UlfNorell:
@williamdemeo @WhatisRT 
I'd like some help with `Ledger.Conformance.Equivalence.Map`. One thing is just filling in the proofs, but what would be nicer would be to prove the properties for an arbitrary `Theory`. Right now they are talking about `Ledger.Prelude` things using the list model, and I couldn't figure out the right way to state them generically and still being able to use them for the list model.

williamdemeo:
I updated the branch by adding proofs of some lemmas in `Ledger/Conway/Conformance/Equivalence/Map` and then by merging master into this branch.  I don't believe the merge caused any issues or reverted some of your work, but if you notice anything wrong, please tell me and I'll fix it.

WhatisRT:
I can see if I find the time to have a look. The issue is that I may not be around much this month, since I still need to take a bunch of PTO.

UlfNorell:
**ATTN**: @williamdemeo @WhatisRT @Soupstraw I had to add (in b511366) a `(DRepDeposit c , d) ∈ dep` check to `GOVCERT-deregdrep` in Conformance. This check is required by the spec and without it the equivalence doesn't hold.

Here is the final main result:
https://github.com/IntersectMBO/formal-ledger-specifications/blob/cf35c285ae7a1df346480fdf76aa6a233a81fe25/src/Ledger/Conway/Conformance/Equivalence.agda#L420-L431

WhatisRT (approved):
Thank you, this looks fantastic! We decided to merge this now and move the cleanup into a followup PR so that we don't get into a state where it gets more difficult to merge after the holidays.

PR #628, opened 2024-12-17
Added witnessing for collaterals

Soupstraw:
# Description

This PR adds the credentials used for collaterals to the set of required witnesses.

The `reg` cert must now be witnessed when the deposit is non-zero. This makes the old registration certificate behave the same way as it does in the Haskell implementation. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Certs.lagda
+++ b/src/Ledger/Certs.lagda
@@ -57,6 +57,11 @@ data DCert : Type where
   ccreghot    : Credential → Maybe Credential → DCert
 \end{code}
 \begin{code}[hide]
+  -- The `reg` cert is deprecated in Conway, but it's still present in this era 
+  -- for backwards compatibility. This has been added to the spec to make 
+  -- conformance testing work properly. We don't talk about this certificate
+  -- in the pdf because it has been deprecated and we want to discourage people 
+  -- from using it.
   reg         : Credential → Coin → DCert
 \end{code}
 \begin{NoConway}
@@ -71,7 +76,12 @@ cwitness (deregdrep c _)     = just c
 cwitness (ccreghot c _)      = just c
 \end{code}
 \begin{code}[hide]
-cwitness (reg _ _)           = nothing
+-- The implementation requires the `reg` cert to be witnessed only if the 
+-- deposit is set. There didn't use to be a field for the deposit, but that was 
+-- added in the Conway era to make it easier to determine, just by looking at 
+-- the transaction, how much deposit was paid for that certificate. 
+cwitness (reg _ zero)        = nothing
+cwitness (reg c (suc _))     = just c
 \end{code}
 \end{NoConway}
 \end{AgdaMultiCode}
--- a/src/Ledger/Utxow.lagda
+++ b/src/Ledger/Utxow.lagda
@@ -99,7 +99,7 @@ getScripts = mapPartial isScriptObj
 
 credsNeeded : UTxO → TxBody → ℙ (ScriptPurpose × Credential)
 credsNeeded utxo txb
-  =  mapˢ (λ (i , o)  → (Spend  i , payCred (proj₁ o))) ((utxo ∣ txins) ˢ)
+  =  mapˢ (λ (i , o)  → (Spend  i , payCred (proj₁ o))) ((utxo ∣ (txins ∪ collateral)) ˢ)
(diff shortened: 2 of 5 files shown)


WhatisRT:
I'm surprised that this is required by the implementation. Has this always been that way? Because that's a deviation from the Shelley spec:
Screenshot 2024-12-17 at 14 40 37

Soupstraw:
This seems to be the PR in `cardano-ledger` that introduced this discrepancy
https://github.com/IntersectMBO/cardano-ledger/pull/3425

Soupstraw:
The witnessing requirement was added in the Conway era, before that we didn't have a field for deposits in the registration cert.

src/Ledger/Utxow.lagda
```diff
@@ -99,7 +99,7 @@ getScripts = mapPartial isScriptObj
 
 credsNeeded : UTxO → TxBody → ℙ (ScriptPurpose × Credential)
 credsNeeded utxo txb
-  =  mapˢ (λ (i , o)  → (Spend  i , payCred (proj₁ o))) ((utxo ∣ txins) ˢ)
+  =  mapˢ (λ (i , o)  → (Spend  i , payCred (proj₁ o))) ((utxo ∣ (txins ∪ collateral)) ˢ)
```
williamdemeo, line 102: Was this a bug? 😮

src/Ledger/Certs.lagda
```diff
@@ -71,7 +71,8 @@ cwitness (deregdrep c _)     = just c
 cwitness (ccreghot c _)      = just c
 \end{code}
 \begin{code}[hide]
-cwitness (reg _ _)           = nothing
+cwitness (reg _ zero)        = nothing
+cwitness (reg c (suc _))     = just c
```
williamdemeo, line 75: Is this reflected in the spec documentation (of a previous era, most likely)?  Or did you find this by conformance testing?  If the latter, is this another adjustment to make the spec fit the implementation?  If so, we should keep a log of such changes and document them somewhere.  I think this is important enough to justify marking my review "changes requested." ...but if you could add some text to the literate version of the module (i.e., in the `Certs.lagda` file), outside of a code block, then I'll approve.  Also, could we start putting an asterisk at the end of any line in the CHANGELOG that describes a change that moves the spec away from the original and/or closer to the implementation, so we can keep track of such changes?

williamdemeo (changes requested):

src/Ledger/Certs.lagda
williamdemeo: I recommend adding text in the `Certs.lagda` file---a sentence or two explaining what's in your remark above ("Actually it requires a witness only if the deposit fee is specified in the certificate. It has nothing to do with the type of the credential.").

Soupstraw: Andre suggested that we put anything related to the `reg` cert into a hidden code block, since it's sort of an implementation detail that would unnecessarily clutter up the spec. Do you think we should make all of this visible and explain it? I don't think it would make sense to explain the code in literate Agda if the code itself is not visible. I could instead add a regular comment which is only visible when reading the source.

Soupstraw: I guess the idea is to discourage people from using a certificate that is going to be deprecated in the next era.

williamdemeo: In general I favor transparency over obfuscation, but if there's a good reason `reg` should be hidden from the reader (e.g., it shouldn't be used in Conway), that's fine.  But, for maintainability, I agree that your comment above should at least appear in a code comment.

src/Ledger/Certs.lagda
```diff
@@ -57,6 +57,11 @@ data DCert : Type where
   ccreghot    : Credential → Maybe Credential → DCert
 \end{code}
 \begin{code}[hide]
+  -- The `reg` cert is deprecated in Conway, but it's still present in this era 
+  -- for backwards compatibility. This has been added to the spec to make 
+  -- conformance testing work properly. We don't talk about this certificate
+  -- in the pdf because it has been deprecated and we want to discourage people 
+  -- from using it.
```
williamdemeo, line 64: Nice. 👍🏼

williamdemeo (approved):
Nice work Joosep!  LGTM

Issue #626, opened 2024-12-11
Witnesses for native scripts in reference inputs


lehins:
It looks like there was deviation in the ledger implementation from specification in the Babbage era. Specification required witnesses for native (phase1) scripts present in reference inputs. 
![image](https://github.com/user-attachments/assets/2d57aac7-90ce-4c96-9c18-b759fd243e0c)
This was never implemented in this way and we only discovered this recently thanks to the conformance tests.

This is not a very important feature for two reasons:
1. Native scripts aren't really used that much in reference inputs
2. There is a [CIP-112](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0112) that is planned for implementation for the next era, which will provide a proper solution for enforcing arbitrary script execution.

So we need to:
1. Stop requiring witnesses for reference scripts in the conway formal specification defined in this repo
2. Fix Babbage specification to reflect the reality.

The last point could benefit from other fixes that probably lead to this discrepancy to begin with:

The predicate passes `neededHashes` as the argument to `txscripts`, which would imply that it filters out only the scripts that were used in the transaction:
![image](https://github.com/user-attachments/assets/2d57aac7-90ce-4c96-9c18-b759fd243e0c)

But the implementation of `txscripts` totally ignores the argument. In fact it doesn't even use the correct type, it should have been a set and it also accepts a `Tx` as its first argument, while its usage supplies tx witnesses: `txw`.
![image](https://github.com/user-attachments/assets/db1c40a2-90ba-4303-a455-7827f8915f21)

WhatisRT:
On the Agda side, this is pretty easy:
```
∙  ∀[ s ∈ mapPartial isInj₁ (txscripts tx utxo) ] validP1Script witsKeyHashes txvldt s
```
needs to be changed to
```
∙  ∀[ s ∈ mapPartial isInj₁ (txscripts tx utxo) ] (hash s ∈ neededHashes → validP1Script witsKeyHashes txvldt s)
```
and then whatever type errors appear need to be fixed.

In the Babbage spec this needs to be handled more carefully. This is a semantic difference that affects script validity, so it shouldn't just be silently adjusted. The Shelley spec has a bunch of errata, and I suggest to do the same here. So the logic should be changed and this situation should be documented in the erratum.

PR #631, opened 2024-12-20
Witnesses for native scripts in reference inputs

williamdemeo:
# Description

Addresses issue #626.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Script.lagda
+++ b/src/Ledger/Script.lagda
@@ -28,6 +28,22 @@ record P1ScriptStructure : Type₁ where
         ⦃ Hashable-P1Script ⦄ : Hashable P1Script ScriptHash
         ⦃ DecEq-P1Script    ⦄ : DecEq P1Script
 
+  -- open Hashable Hashable-P1Script public
+  -- I think `Hashable Hashable-P1Script` should be opened publicly here,
+  -- but (probably because that wasn't done to begin with), it now conflicts
+  -- with other instances, e.g., the `hash` in line 200 of `Ledger.Transaction`
+  -- would produce the following type-checking error:
+  -- ```
+  -- Cannot resolve overloaded projection hash because principal argument
+  -- has type (P1Script ⊎ P2Script) while it should be of record type
+  --   candidates in scope:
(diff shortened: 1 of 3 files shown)


src/Ledger/Utxow.lagda
```diff
@@ -161,9 +161,10 @@ data _⊢_⇀⦇_,UTXOW⦈_ where
         neededHashes      = scriptsNeeded utxo txb
         txdatsHashes      = dom txdats
         allOutHashes      = getDataHashes (range txouts)
+        nonRefScripts     = mapPartial isInj₁ (txscripts tx utxo)
```
Soupstraw, line 164: I think it should be called something like `nativeScripts`. `nonRefScripts` doesn't make much sense here because `txscripts` gives both the reference scripts and scripts in the `txins`.

src/Ledger/Script.lagda
```diff
@@ -159,7 +159,6 @@ record ScriptStructure : Type₁ where
 
   field hashRespectsUnion :
           {A B Hash : Type} → Hashable A Hash → Hashable B Hash → Hashable (A ⊎ B) Hash
-        ⦃ Hash-P1Script ⦄ : Hashable P1Script ScriptHash
```
Soupstraw, line 162: Why is this change necessary?

williamdemeo: I believe that line was a bug.  It used to be a timelock instance and, for some reason, was changed to a `Hash-P1Script` instance, but that clashes with the other `Hash-P1Script` that was introduced earlier in this module.  If both instances are present in the module, then Agda doesn't know how to choose an instance when the new call to `hash` is made in the following line:
```agda
    ∙  ∀[ s ∈ nonRefScripts ] (hash s ∈ neededHashes → validP1Script witsKeyHashes txvldt s)
```

WhatisRT (approved):
I think just update the variable name as suggested and then we can merge this

WhatisRT:
Since @williamdemeo is away for a bit I'll take care of it

Issue #621, opened 2024-12-03
Add diagrams to pdf


williamdemeo:
We need high-level diagrams that illustrate the overall layout of the ledger and how it works as a state machine (and as a ledger).

The new diagrams should go in cardano-ledger.pdf (probably not in conway-ledger.pdf which is mainly for describing aspects of the ledger that are specific to Conway).

Probably we want to start off with something simple, like the tikz latex package.  However, automating the generation of diagrams would enable us to make sure the diagrams are updated when changes to the ledger occur.  This will make the formal spec easier to maintain and help ensure the diagrams are correct and not misleading readers.

For automation, we might try to use the [diagrams](https://diagrams.github.io/) language, since it is Haskell based so could probably be plugged right into our stack without too much effort.  However, it may take a day or two to learn and figure out how to configure and integrate it.

williamdemeo:
**Existing Diagrams**

- Figure 78 (STS Rules, Sub-rules and Dependencies) in [shelley-ledger.pdf](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf)
- [Alexey's hand-drawn diagrams](https://input-output-rnd.slack.com/archives/C01LUSYT491/p1688131825767479) (<- links to internal Slack message)

PR #639, opened 2025-01-08
Add diagrams to pdf

williamdemeo:
# Description

Addresses issue #621.

# Checklist

- [ ] Self-reviewed the diff


WhatisRT:
I think it's best to just make `tex` files and put them in `src/latex`. That'll fix the build failures and there's no reason for those to be literate Agda anyway.

williamdemeo:
> I think it's best to just make `tex` files and put them in `src/latex`. That'll fix the build failures and there's no reason for those to be literate Agda anyway.

Okay, but I'm not sure how plain `tex` files should be integrated into the pdf?

I initially put the files under `src/latex/Diagrams`, as `tex` files, and then pulled them into the pdf via input lines in a `src/latex/Diagrams.tex` file, which I `\input` in `Ledger/PDF.lagda`, but that doesn't work.  I get an error.  (I'll paste the error here after I re-run `nix-build -A ledger.docs`.)

WhatisRT:
> I initially put the files under `src/latex/Diagrams`, as `tex` files, and then pulled them into the pdf via input lines in a `src/latex/Diagrams.tex` file, which I `\input` in `Ledger/PDF.lagda`, but that doesn't work.

Can you give me the error message for that? That's more or less what I'd expect to work.

> Here's the error I get when running `nix-build -A ledger.docs`:
> 
> ```
> Checking that all Agda files have been typechecked...
> stat: cannot statx '_build/2.7.0/agda/./latex/Diagrams/ApplyBlock.agdai': No such file or directory
>         FAIL: _build/2.7.0/agda/./latex/Diagrams/ApplyBlock.agdai is not up-to-date
> ```

Yeah, that's because those files are agda files which aren't imported from the top level. Maybe I'll improve the error message on that.

williamdemeo:
> > I initially put the files under `src/latex/Diagrams`, as `tex` files, and then pulled them into the pdf via input lines in a `src/latex/Diagrams.tex` file, which I `\input` in `Ledger/PDF.lagda`, but that doesn't work.
> 
> Can you give me the error message for that? That's more or less what I'd expect to work.
> 
> > Here's the error I get when running `nix-build -A ledger.docs`:
> > ```
> > Checking that all Agda files have been typechecked...
> > stat: cannot statx '_build/2.7.0/agda/./latex/Diagrams/ApplyBlock.agdai': No such file or directory
> >         FAIL: _build/2.7.0/agda/./latex/Diagrams/ApplyBlock.agdai is not up-to-date
> > ```
> 
> Yeah, that's because those files are agda files which aren't imported from the top level. Maybe I'll improve the error message on that.

No, that's the error message I get when I make them `.tex` files and set it up as you suggested (which, incidentally, is how I tried to integrate the diagrams initially, before switching to `.lagda` files).

williamdemeo:
I somehow forgot to rename the files from `.lagda` to `.tex` extension.  Integrating the diagrams as tex files works fine now.  (I'm not sure what about this is different from the very first time I tried using tex files for the diagrams.)

WhatisRT:
A few thoughts about the diagrams:
- Pretty much all arrows go back & forth, which is pretty noisy. It's also not clear what an arrow means, so I'd suggest defining the meaning of an arrow first (and explaining it) and then see what should and should't have arrows. Note that something like 'following the thread of the execution' is not a viable meaning here, since the spec doesn't have an execution order.
- What exactly do the colors on the STSs indicate? Is it supposed to mean when it was introduced or when it was last changed? I think last updated would be more useful.
- The types all refer to things as they are called in the implementation, e.g. `()` instead of `⊤`. There's also no `PulsingRewardUpdate` in the spec, that's an implementation detail.
- There are a bunch of references to implementation details, such as the entry point and which module they are in. These don't make sense here.
- The diagram for conformance testing is nice, but I'm not sure it should be in the spec. I think it would make more sense as a separate little document, possibly with some extra information about conformance testing more generally.
- The Shelley era diagram still looks quite useful to me. I wonder if the most sensible thing to do would be to just update that & include the color information. That means losing the types of the STSs (they wouldn't fit) but I'm not sure how useful they actually are. OTOH if we want to include the types the current split seems sensible but it'd be nice to still provide all the information. For that I think there should be a diagram starting from `CHAIN` that then references the subtrees that are defined in the other diagrams.

williamdemeo:
> A few thoughts about the diagrams:
> 

- [ ] Pretty much all arrows go back & forth, which is pretty noisy. It's also not clear what an arrow means, so I'd suggest defining the meaning of an arrow first (and explaining it) and then see what should and should't have arrows. Note that something like 'following the thread of the execution' is not a viable meaning here, since the spec doesn't have an execution order.
- [ ] What exactly do the colors on the STSs indicate? Is it supposed to mean when it was introduced or when it was last changed? I think last updated would be more useful.
- [ ] The types all refer to things as they are called in the implementation, e.g. `()` instead of `⊤`. There's also no `PulsingRewardUpdate` in the spec, that's an implementation detail.
- [ ] There are a bunch of references to implementation details, such as the entry point and which module they are in. These don't make sense here.
- [ ] The diagram for conformance testing is nice, but I'm not sure it should be in the spec. I think it would make more sense as a separate little document, possibly with some extra information about conformance testing more generally.
- [ ] The Shelley era diagram still looks quite useful to me. I wonder if the most sensible thing to do would be to just update that & include the color information. That means losing the types of the STSs (they wouldn't fit) but I'm not sure how useful they actually are. OTOH if we want to include the types the current split seems sensible but it'd be nice to still provide all the information. For that I think there should be a diagram starting from `CHAIN` that then references the subtrees that are defined in the other diagrams.

Thanks @WhatisRT.  (I've turned your comments into the list of todo items above.)

carlostome:
PDF compilation is failing. It seems you might have forgotten to push a file <https://github.com/IntersectMBO/formal-ledger-specifications/actions/runs/13622994816/job/38075519583?pr=639#step:4:6685>

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,177 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+present below.  Keep in mind that this section focuses on intuition, using
+terms (set in italics) which may be unfamiliar to some readers, but rest assured that
+later sections of the document will make the intuition and italicized terms precise.
+
+By \textit{ledger}, we mean a record of a data structure at a specific point in time.
+This data structure contains, for example, an accounting of how funds in the system
+are distributed among different accounts; the ADA currently held in the reserves; a
+list of \textit{stake pools} operating the network, and so on.
+
+This ledger can be updated in response to certain signals, such as receiving a new
```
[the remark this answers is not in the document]
williamdemeo: I agree, the sentence on line 74 isn't good and should probably be removed.  We should say a ledger is... what it says in the second sentence of that paragraph.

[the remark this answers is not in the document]
williamdemeo: I've revised the section.

In particular, I think it would be much clearer if we agreed to refer only to the whole ledger as a "transition _system_" and refer to each component STS (e.g., UTXO) as a "transition rule."  This will be less confusing than if we refer to each transition rule as a "state transition system."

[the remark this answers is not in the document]
williamdemeo: It's probably okay to point out initially that the rules are "state transition systems," but after that I think we should adhere to the convention of calling them "transition rules" or just "rules."

Here is what I propose:

"We will model this via a number of _state transition systems_ (STS) which from now on we refer to as 'transition rules.'  These rules describe the different behaviors that determine how the whole system evolves."

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,177 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+        purposes of the rule;
+  \item an \textit{initial state}, which consists of the subset of the full ledger
+        state that the transition can update;
+  \item a \textit{signal}, with associated data, that the transition system can
+        receive;
+  \item a set of \textit{preconditions} that must be met in order for the transition
+        to be valid;
+  \item a new state that results from said transition.
+\end{itemize}
+For example, the UTXOW transition system defined in \cref{fig:rules:utxow} of
```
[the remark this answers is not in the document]
williamdemeo: Actually, I don't think that's necessary after rewording the description.  I think it's clearer now.

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,177 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+  (cert)     edge  (deleg)
+  (cert)     edge  (pool)
+  (cert)     edge  (govcert)
+  (utxow)    edge  (utxo)
+  (utxo)     edge  (utxos);
+  \end{tikzpicture}
+  \caption{State transition systems of the ledger specification, presented as a
+  directed graph; each node represents a transition rule; an arrow from rule A to
+  rule B indicates that B appears among the premises of A.}
+  \label{fig:latest-sts-diagram}
```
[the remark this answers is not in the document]
williamdemeo: If we agree that we'll call the individual state transition systems "transition rules," I think the caption text is okay.

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,141 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+  \TrinaryInfC{CERT...CERT}
+  \UnaryInfC{CERTS}
+  \TrinaryInfC{LEDGER...LEDGER}
+  \UnaryInfC{LEDGERS}
+  \BinaryInfC{CHAIN}
+\end{prooftree}
+}
+\caption{State transition rules of the ledger specification, presented as a deduction tree.}
+\label{fig:new-chain-diagram}
+\end{figure}
```
carlostome, line 157: - Too big to fit in the PDF
- It seems to contain different nodes than the figure above
- Not clear what `LEDGER..LEDGER` means

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,141 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+  reached the thresholds it needs to be ratified (\cref{sec:ratification}).
+\item \LedgerModText{Epoch}{SNAP} computes new stake distribution snapshots (\cref{sec:epoch-boundary}).
+\item \LedgerModText{Epoch}{EPOCH} computes the new state as of the end of an epoch;
+  includes the ENACT, RATIFY, and SNAP transition rules (\cref{sec:epoch-boundary}).
+\item \LedgerModText{Epoch}{NEWEPOCH} computes the new state as of the start of a new
+  epoch; includes the previous EPOCH transition (\cref{sec:epoch-boundary}).
+\item \LedgerModText{Chain}{CHAIN} is the top level transition in response to a new
+  block that applies the NEWEPOCH transition when crossing an epoch boundary, and the
+  LEDGERS transition on the list of transactions in the body (\cref{sec:blockchain-layer}).
+\end{itemize}
```
carlostome, line 197: Order seems a bit random

src/Ledger/PDF.lagda
```diff
@@ -97,5 +97,6 @@ open import Ledger.PDF.ConwayBootstrap
 \input{EssentialAgda}
 \input{Ledger/PDF/ConwayBootstrapEnact}
 \input{Ledger/PDF/ConwayBootstrap}
+%% \input{Diagrams}
```
carlostome, line 100: Why is this commented out?

src/Makefile
```diff
@@ -1,13 +1,19 @@
...
+#
+# @file
+# @version 0.1
+
 # Constants
 AGDA?=agda
 AGDA_RUN=$(AGDA) --transliterate
 OUT_DIR?=../dist
-LATEX?=latexmk -xelatex -shell-escape -halt-on-error
+LATEX?=latexmk -xelatex -shell-escape -halt-on-error -gg
```
carlostome, line 11: I'm not sure we want this option here. According to the manual it forces recompilation of everything.
```
-gg "Super go mode" or "clean make": clean out generated files as if -C had been given, and then do a
regular make."
```

src/latex/macros.sty
```diff
@@ -3,7 +3,7 @@
 \RequirePackage{pifont}  %% needed for checkmark (\ding{51}).
 \RequirePackage{extarrows}
 \RequirePackage{slashed}
-\RequirePackage[tikz]{bclogo}
+
```
carlostome, line 6: Remove the empty line
```suggestion
```

carlostome (commented):
Good job, some nitpicks here and there.

src/Ledger/Introduction.lagda
```diff
@@ -60,6 +60,189 @@ Conway~\cite{cip1694} & Complete & Partial & Partial \\
...
+
+By \textit{ledger} we mean a structure that contains information about
+how funds in the system are distributed accross accounts---that is, account
+balances, how such balances should be adjusted when transactions and
+proposals are processed, the ADA currently held in the treasury reserve, a
+list of \textit{stake pools} operating the network, and so on.
+
+The ledger can be updated in response to certain events, such as receiving a new
+transaction, time passing and crossing an \textit{epoch boundary}, enacting a
+\textit{governance proposal}, to name a few.  Thus, the ledger also defines a set of
```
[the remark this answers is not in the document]
williamdemeo: Your suggestion reads as though "this document" is part of the behavior of the ledger, but what I think you mean to say is that which this document defines is part of the behavior of the ledger.  I'll rewrite the sentence as follows: "Part of the behavior of the ledger this document defines is a set of rules that determine which events..."

src/Ledger/Introduction.lagda
williamdemeo: I think I should just delete the deduction tree diagram.

src/Ledger/Introduction.lagda
williamdemeo: Okay, I'll fix that (by ordering the list according to the order in the diagram, from top to bottom and left to right.)

src/Makefile
williamdemeo: Yeah, I added that when I got frustrated by the "already up-to-date" error, when, in fact, the document was not up-to-date.  I'll remove it.

carlostome (commented):
LGTM

carlostome (approved):

PR #643, opened 2025-01-14
Improve the error message when a file hasn't been type checked

WhatisRT:
# Description

Just a minor wording improvement so that it becomes clear what to do when the script that checks if files have been checked fails.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/checkTypeChecked.sh
+++ b/src/checkTypeChecked.sh
@@ -4,8 +4,10 @@ echo "Checking that all Agda files have been typechecked..."
 for agdaFn in $(find . -name '*.*agda'); do
   agdaiFn="_build/2.7.0/agda/${agdaFn%.*agda}.agdai"
   if [ "$(( $(stat -c "%Y" $agdaiFn) - $(stat -c "%Y" $agdaFn) ))" -lt "0" ]; then
-    echo "        FAIL: $agdaiFn is not up-to-date"
+    if [ "$1" = "-m" ]; then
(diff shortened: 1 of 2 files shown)


src/checkTypeChecked.sh
```diff
@@ -4,8 +4,8 @@ echo "Checking that all Agda files have been typechecked..."
 for agdaFn in $(find . -name '*.*agda'); do
   agdaiFn="_build/2.7.0/agda/${agdaFn%.*agda}.agdai"
   if [ "$(( $(stat -c "%Y" $agdaiFn) - $(stat -c "%Y" $agdaFn) ))" -lt "0" ]; then
-    echo "        FAIL: $agdaiFn is not up-to-date"
+    echo "        FAIL: $agdaiFn does not exist. Please remove the corresponding agda file or import it somewhere."
```
williamdemeo, line 7: I had the impression that if the script doesn't find an agdai file then it exits and then the corresponding agda file will be type-checked, so the agdai file will be there on the next pass.  If that's true, then this error message isn't really what we want.  If that's wrong, then I'm fine with the change.

williamdemeo (approved):
Looks fine, assuming my understanding of the way this script is meant to work is wrong.

src/checkTypeChecked.sh
WhatisRT: I need to double check, but I think the first pass of this script is probably pointless and can be removed. If it does something we should add a flag to this script to give the old error message on the first run and the new one on the second run. IIRC the first run of this script only prevents a call to `agda` if all the interface files are there, but then that call would be very quick anyway (and this would only ever happen locally anyway).

PR #649, opened 2025-01-23
Eliminate Boolean properties

### LINKED ISSUE #342: Eliminate boolean properties
For convenience or other reasons, some properties/predicates produce booleans. Our properties should always go to `Set`, so we can use them without writing something like `≡ true`.

Here are some offenders:
- [x] `actionWellFormed`
- [ ] `isP2Script`
- [ ] `isAdaOnlyᵇ`
- [ ] `feesOK`

`delayingAction` is an exception, since we use this to update an actual `Bool` in a state.

carlostome:
# Description

Addresses #342.

Original:
- [x] `actionWellFormed`
- [x] `isP2Script`
- [x] `isAdaOnlyᵇ`
- [x] `feesOK`

Other:
- [x] `isTwoPhaseScriptAddress`

Left:
- [ ] `isValid`

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Transaction.lagda
+++ b/src/Ledger/Transaction.lagda
@@ -238,9 +238,9 @@ Ingredients of the transaction body introduced in the Conway era are the followi
   isP2Script : Script → Type
   isP2Script = From-inj₂
 
-  isP2Script? : ∀ {s} → Dec (isP2Script s)
-  isP2Script? {inj₁ x} = true because ofʸ (lift tt)
-  isP2Script? {inj₂ y} = true because ofʸ y
+  isP2Script? : ∀ {s} → isP2Script s ⁇
+  isP2Script? {inj₁ x} .dec = true because ofʸ (lift tt)
+  isP2Script? {inj₂ y} .dec = true because ofʸ y
 
   instance
     HasCoin-TxOut : HasCoin TxOut
--- a/src/Ledger/Transaction.lagda
+++ b/src/Ledger/Transaction.lagda
@@ -236,11 +236,11 @@ Ingredients of the transaction body introduced in the Conway era are the followi
 
 \begin{code}[hide]
   isP2Script : Script → Type
-  isP2Script = From-inj₂
+  isP2Script = T ∘ is-just ∘ isInj₂
 
   isP2Script? : ∀ {s} → isP2Script s ⁇
-  isP2Script? {inj₁ x} .dec = true because ofʸ (lift tt)
-  isP2Script? {inj₂ y} .dec = true because ofʸ y
+  isP2Script? {inj₁ x} .dec = false because ofⁿ (λ ())
+  isP2Script? {inj₂ y} .dec = true because ofʸ tt
 
   instance
     HasCoin-TxOut : HasCoin TxOut
--- a/src/Ledger/Transaction.lagda
+++ b/src/Ledger/Transaction.lagda
@@ -239,8 +239,8 @@ Ingredients of the transaction body introduced in the Conway era are the followi
   isP2Script = T ∘ is-just ∘ isInj₂
 
   isP2Script? : ∀ {s} → isP2Script s ⁇
-  isP2Script? {inj₁ x} .dec = false because ofⁿ (λ ())
-  isP2Script? {inj₂ y} .dec = true because ofʸ tt
+  isP2Script? {inj₁ x} .dec = no λ ()
+  isP2Script? {inj₂ y} .dec = yes tt
 
   instance
     HasCoin-TxOut : HasCoin TxOut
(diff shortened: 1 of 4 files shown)


src/Ledger/Transaction.lagda
```diff
@@ -235,8 +235,12 @@ Ingredients of the transaction body introduced in the Conway era are the followi
 \end{NoConway}
 
 \begin{code}[hide]
-  isP2Script : Script → Bool
-  isP2Script = is-just ∘ isInj₂
+  isP2Script : Script → Type
+  isP2Script = From-inj₂
```
WhatisRT, line 239: `From-inj₂` is incorrect here, since it's always inhabited whereas the original code gives a `false` on the `inj₁` constructor. It's a shame the stdlib doesn't have something like `IsInj₂`, defined like `From-inj₂` but with a `⊥` instead of a `⊤`.

src/Ledger/Transaction.lagda
```diff
@@ -235,8 +235,12 @@ Ingredients of the transaction body introduced in the Conway era are the followi
...
+
+  isP2Script? : ∀ {s} → isP2Script s ⁇
+  isP2Script? {inj₁ x} .dec = true because ofʸ (lift tt)
+  isP2Script? {inj₂ y} .dec = true because ofʸ y
```
WhatisRT, line 243: You can also see it here, it would be odd if we had a property that's always true. Also, we typically use `yes x` instead of `true because ofʸ x`.

src/Ledger/Utxo.lagda
```diff
@@ -61,8 +71,9 @@ opaque
 
   getInputHashes : Tx → UTxO → ℙ DataHash
   getInputHashes tx utxo = getDataHashes
-    (filterˢ (λ (a , _ ) → isTwoPhaseScriptAddress tx utxo a ≡ true)
-            (range (utxo ∣ txins)))
+    (filterˢ (λ (a , _ ) → isTwoPhaseScriptAddress tx utxo a)
+             ⦃ λ {(a , _)} → isTwoPhaseScriptAddress? {tx} {utxo} {a} ⦄
```
WhatisRT, line 75: Here we have an interesting problem: this code belongs to the visible section of the PDF and we shouldn't show stuff like that because we'd have to explain what it means (and the meaning isn't particularly interesting). However, marking `isTwoPhaseScriptAddress?` as an instance isn't a good idea and wouldn't even work anyway.

I think the only reasonable solution here is to turn `isTwoPhaseScriptAddress` into a `data` type and then provide a `⁇` instance for that type. As I see it, there are two options to do that:
- either completely replace the function with a type whose constructors contain the same logic, or
- provide a 'newtype' wrapper around `isTwoPhaseScriptAddress`, i.e. something of this form:
  ```
  data isTwoPhaseScriptAddress' : Tx → UTxO → Addr → Type where
    wrap : ∀ {tx utxo a} → isTwoPhaseScriptAddress tx utxo a → isTwoPhaseScriptAddress' tx utxo a
  ```
The second option has the upside that we can cheat a little and just hide this wrapper behind the curtain when rendering the PDF. We currently don't have the machinery to do that perfectly because we'd have to substitute `isTwoPhaseScriptAddress'` with `isTwoPhaseScriptAddress` in the line above when rendering the PDF, but we probably need a mechanism for that anyway. So I'm currently leaning more towards that option.

WhatisRT (changes requested):
`feesOK` looks great! The other two still need a bit of work.

src/Ledger/Utxo.lagda
carlostome: How would we go about substituting `isTwoPhaseScriptAddress'` by its unprimed version in the pdf?

williamdemeo: By "cheat a little" Andre means we use a (very ugly) python script that I wrote to tweak the Agda-generated latex source so that the resulting pdf is easier to digest but isn't really an authentic representation of the Agda code.  I'm generally opposed to this approach, and would prefer if the pdf were an authentic representation of the formalization, but I'm overruled by the need to keep things as simple as possible for the intended/expected audience, lest we end up having to explain too much and get into the Agda weeds, which we're assuming most readers will want to avoid.

So, in this case, removal of the prime could be added to the post-processing script.  Let me know if you want me to do that, because I'll be ashamed and embarrassed if you have to look at that script!

WhatisRT:
@carlostome I'd like to merge this, the only thing that's still open is the choice of `Dec` constructors. If you have a reason to prefer `_because_` we can leave it as is, but otherwise I'd prefer to go with `yes` and `no` for consistency.

WhatisRT (approved):
Thanks! I've renamed `isAdaOnlyᵇ` to `isAdaOnly` because the `ᵇ` was just to indicate that it's a boolean. I've also removed a `ˢ` that wasn't necessary anymore.

WhatisRT:
Whoa, apparently removing the `ˢ` breaks something in a completely different place. No idea what's going on there, I'll just revert that for now.

PR #650, opened 2025-01-23
Replace a call to `error` in our FFI with a safe alternative

WhatisRT:
# Description

There was no need to be unsafe here. Our code just happened to not match on that proof, but this could have broken at any time when doing refactoring. In the Consensus spec (which copies our code), this actually caused a crash.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Conway/Foreign/HSLedger/ExternalStructures.agda
```diff
@@ -24,7 +24,7 @@ instance
       ; sign             = λ _ _ → zero
         -- we can't prove these properties since the functions are provided by the Haskell implementation
```
Soupstraw, line 25:
```suggestion
        -- we can't prove correctness since the function is provided by the Haskell implementation
```

Soupstraw (approved):

PR #661, opened 2025-01-30
Disallow empty withdrawal proposals

### LINKED ISSUE #654: Conformance failure: Empty withdrawals are valid in the spec but not in the implementation
[test.txt](https://github.com/user-attachments/files/18572381/test.txt)

Soupstraw:
# Description

resolves #654

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/GovernanceActions.lagda
+++ b/src/Ledger/GovernanceActions.lagda
@@ -67,7 +67,9 @@ data GovAction : Type where
 
 actionWellFormed : GovAction → Type
 actionWellFormed (ChangePParams x)  = ppdWellFormed x
-actionWellFormed (TreasuryWdrl x)   = ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+actionWellFormed (TreasuryWdrl x)   = 
+  ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+  × ∃[ v ∈ range x ] ¬ (v ≡ 0)
 actionWellFormed _                  = ⊤
 \end{code}
 \begin{code}[hide]


src/Ledger/GovernanceActions.lagda
```diff
@@ -67,7 +68,9 @@ data GovAction : Type where
 
 actionWellFormed : GovAction → Type
 actionWellFormed (ChangePParams x)  = ppdWellFormed x
-actionWellFormed (TreasuryWdrl x)   = ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+actionWellFormed (TreasuryWdrl x)   = 
+  ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+  × ¬ (∑ˢ[ v ← range x ] v ≡ 0)
```
carlostome, line 73: Wouldn't this be more clear as a "positive" property: `\forall v \in range x. \neg (v = 0)?

Soupstraw: That would be a slightly different statement, this one is stating that the sum of withdrawals is non-zero, but yours is saying that there can't be any zero coin withdrawals. Your condition makes more sense in my opinion, but the implementation looks at the sum and at this point we can't really change that, so the spec has to behave like that.

WhatisRT: Yeah, it's a good point. If you're checking that the withdrawals are sensible then forbidding zeros in the map makes more sense. For now we don't have a choice here, but we could make a ticket for the next era to change it.

Soupstraw: well, I guess I could write it as `∃[ v ∈ range x ] ¬ (v ≡ 0)`

src/Ledger/GovernanceActions.lagda
```diff
@@ -20,6 +20,7 @@ different purpose.
 
 open import Data.Nat.Properties using (+-0-monoid)
 open import Data.Rational using (ℚ; 0ℚ; 1ℚ)
+open import Data.Product.Base
```
WhatisRT, line 23: Is that import required? I feel like everything in there should already be in scope from the prelude.

src/Ledger/GovernanceActions.lagda
WhatisRT: Ah, true, and then the proposed future change would be to replace `∃` by `∀`.

WhatisRT (approved):
LGTM

src/Ledger/GovernanceActions.lagda
```diff
@@ -67,7 +67,9 @@ data GovAction : Type where
 
 actionWellFormed : GovAction → Type
 actionWellFormed (ChangePParams x)  = ppdWellFormed x
-actionWellFormed (TreasuryWdrl x)   = ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+actionWellFormed (TreasuryWdrl x)   = 
+  ∀[ a ∈ dom x ] RwdAddr.net a ≡ NetworkId
+  × ∃[ v ∈ range x ] ¬ (v ≡ 0)
```
carlostome, line 72: This should be a forall not an exists

PR #670, opened 2025-02-03
Remove unused parameter in function stakeDistr

carlostome:
# Checklist

- [x] Self-reviewed the diff


WhatisRT (commented):
Ah, this function was only partially implemented. A correct implementation would need `PState`, so I don't think it's a good idea to remove it.

If you want to finish the implementation, look at Fig. 37 of the Shelley spec. Note that `addrPtr` has now been deprecated and removed, so that part is intentionally missing. This is also one of the few places where the spec really leans into set theory, so the inverse here turns a function into a relation and performs some operations on relations. I think `stakeRelation` is correct in this code.

Issue #31, opened 2022-11-23
Well-formedness properties


WhatisRT:
There are many properties that are currently part of the rules, which don't really fit into any particular rule thematically, and don't depend on anything but the signal. We could call these 'well-formedness properties', and bundle all of them into one place for them to be checked. This should improve consistency and readability of the STS rules.

williamdemeo:
@WhatisRT could you create a list of the well-formedness properties, or at least some examples to get me started?

WhatisRT:
I've added a bunch of examples to the top

PR #673, opened 2025-02-04
Moved action specific predicates to actionWellFormed and fix a conformance failure

Soupstraw:
# Description

This PR moves the action-specific predicates in `GOV-Propose` to `actionWellFormed` and adds a check that makes sure all the receiving staking credentials of a `TreasuryWithdrawal` are registered when the proposal is made.

This probably requires some changes to the prose. It seemed like the prose about `actionWellFormed` was already outdated before these changes, because it only mentioned we check the parameter updates, but we also had some conditions about treasury withdrawals as well.

I also added two other predicate checks to the GOV rule which are present in the implementation:
- In `GOV-Propose`: the return address of a new proposal must be a registered staking credential
- In `GOV-Vote`: the proposal being voted for must not be expired

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Conformance/Equivalence.agda
+++ b/src/Ledger/Conway/Conformance/Equivalence.agda
@@ -396,8 +396,8 @@ opaque
           → (record Γ { certState = certState' } , n)
             ⊢ s ⇀⦇ txgov ,GOVn⦈ s'
   castGOV deps (BS-base Id-nop) = BS-base Id-nop
-  castGOV {Γ} deps (BS-ind (C.GOV-Vote {voter = voter} (a , b , c)) rs) =
-    BS-ind (C.GOV-Vote (a , b , cast-isRegistered Γ deps voter c))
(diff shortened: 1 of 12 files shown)


WhatisRT (commented):
Actually, now that I look at it I think it's better to move it into a separate helper function. The idea is that `actionWellFormed` follows this pattern: https://github.com/IntersectMBO/formal-ledger-specifications/issues/31 And that's destroyed by adding these extra arguments.

williamdemeo:
Sorry @Soupstraw I didn't see this earlier.  I'll review it today!

williamdemeo:
I've moved `actionValid` and `actionWellFormed` out of the hidden code block and into a new figure.  I also added prose near that figure to explain the logic, and reduced the prose about this logic that appears below the GOV rule figure (at the bottom of `Gov.lagda`), and replaced it with a reference to the new figure.

williamdemeo (approved):
I just pushed my suggested changes directly to your branch since they merely reorganize and document the code a bit.  Otherwise, looks really nice.  Good work!

WhatisRT (approved):
LGTM

Issue #683, opened 2025-02-11
Review the Conway diff spec


WhatisRT:
Here are some things I've spotted in a quick first pass. I had hoped for this to be a short list, but it didn't really turn out that way.

To make progress on these items, it's probably best to tackle the items starting with the easiest and least controversial ones. What's left can then become its own issue, which may or may not be fixed before we tag a release version.

- Protocol Parameters:
  - [ ] Would be good to cite CIP-1694 for the groups more clearly
  - [x] `PParams` should have its own figure
  - [x] `positivePParams` has too long lines
- Fee calculation
  - [x] `with` shouldn't show up in figures, can it be replaced by `case`?
  - [x] Something isn't syntactically correct. Is there a `where` keyword missing?
  - [x] A whole bunch of proof bits of the well-founded recursion are visible
- Transactions
  - [x] The declaration of `TransactionStructure` is visible
- UTxO
  - [ ] Something needs to be done about `ValidCertDeposits`. It either needs to be properly explained or its definition visible. The latter is probably a bad idea because it uses indices IIRC.
  - [x] The explanation for `_≡?_` would probably be better in the Notation section.
  - Figure 16:
    - [x] I think we can remove the universally quantified names at the beginning of the rules again, since https://github.com/agda/agda/issues/7380 was fixed as part of Agda 2.7. It should be confirmed that this does not break conformance testing.
    - [x] Lot's of `open` going on. `Γ` and `s` can probably be 'pattern-mached' at the bottom of the rule instead of using `open` on them. The `open` statements for `tx` and `txb` can probably just be hidden (we do that elsewhere too).
    - [x] The only thing that actually changed here is the handling of the `donations` field, it would be good to highlight or mention that.
  - [x] Figure 17 also has the same issue with `open`, but here we can't 'pattern-match'. At least we can hide the ones for `tx` and `txb`, the other two could be done by actual pattern-matching in the let-binding (see Fig. 20).
  - [x] Figure 20 has broken alignment and is missing the actual `let` syntax
- Governance
  - [x] Figure 21 has a broken caption
  - [x] Relation `_≈ᵍ_` should be explained better. Two actions relate with it if they modify the same piece of `EnactState`, which also means they can depend on each other (see `hasParent`). `_≈ᵍ_` is called `_≈_` in the prose.
  - [x] Figure 22 doesn't need the 'Functions used in the GOV rules' label at the top, we never do that elsewhere
[...]

PR #695, opened 2025-02-21
Review the conway diff spec

carlostome:
# Description

This addresses #683 

# Checklist

- [ ] Self-reviewed the diff


src/Ledger/Fees.lagda
```diff
@@ -7,62 +7,74 @@ open import Ledger.Transaction
...
 
 \section{Fee Calculation}
 \begin{figure*}[h]
 \begin{AgdaMultiCode}
-\begin{code}[hide]
-_↑ℚ : ℕ → ℚ
-n ↑ℚ = mkℚ+ n 1 (1-coprimeTo n ∘ swap)
-\end{code}
 \begin{code}
 scriptsCost : (pp : PParams) → ℕ → Coin
```
carlostome, line 27: Why is the name in plural? If I understand correctly this is used to calculate the cost of one script

src/Ledger/Fees.lagda
```diff
@@ -71,16 +83,20 @@ scriptsCost pp scSz with (PParams.refScriptCostStride pp)
...
 the fee for reference scripts in the transaction using a function that
 is piece-wise linear in the size, where the linear constant multiple
-grows with each \refScriptCostStride{} bytes. Thus, the \scriptsCost{}
-function depends on the \AgdaFunction{scriptsTotalSize} function,
-which returns an integer that is the total size of the reference
-script in bytes, as well as the following protocol parameters:
+grows with each \AgdaFunction{refScriptCostStride} bytes.
+%
+This function takes as input the total size of the reference script in
+bytes. The function \AgdaFunction{scriptsTotalSize} is used to
```
carlostome, line 89: Where is `scriptsTotalSize` defined?

src/Ledger/Fees.lagda
WhatisRT: No, the argument it gets is the sum of the sizes of reference scripts, so this name is correct. Is there some inconsistency elsewhere?

src/Ledger/Fees.lagda
WhatisRT: It doesn't seem to exist. I think it may refer to `refScriptsSize` going by the explanation, so maybe there was some renaming going on?

This explanation is also quite confusing, I'm not sure what a function depending on another function is supposed to mean. I'd just write something like '\scriptsCost{} expects the sum of sizes of reference scripts as computed by \refScriptsSize{} as its input'.

WhatisRT: Now that I look at it, it's also confusing to call these arguments protocol parameters (below). They're currently hardcoded and may become protocol parameters in the future, if we don't manage to get rid of that function.

src/Ledger/Fees.lagda
carlostome: Somewhere in the explanation it was suggested:
"... which returns an integer that is the total size of the reference script ..."

src/Ledger/Fees.lagda
carlostome: I see. Maybe it would be best to redefine `scriptsCost` such that all those are given as arguments. WDYT?

src/Prelude.agda
```diff
@@ -76,3 +78,11 @@ syntax ∃₂-syntax (λ x y → C) = ∃₂[ x , y ] C
 import Data.Nat.Literals as ℕ
 
 instance Number-ℕ = ℕ.number
+
+-- (pseudo)equality (for Maybe)
+_≡?_ : {A : Type} → Maybe A → Maybe A → Type
```
carlostome, line 83: I find this symbol a bit confusing since `?` is everywhere else? used for decidability purposes.

I'd suggest to rename it to one of: ≊ ≋ ≅

src/Ledger/Fees.lagda
WhatisRT: That just shifts the problem to the call site, but they're still going to be hard coded somewhere. Having them at least be defined right next to where they are used makes it more local.

I guess we could have some type that holds these things, but I think this is the only instance we currently have of these magic numbers. Another option would be to add them to the `GlobalConstants`, but these also have a very specific meaning (these are set by the 'genesis file' in the implementation). So all in all I don't think this can really be improved.

src/Prelude.agda
williamdemeo: I like the first two equality symbols you suggest, but the last is overused.

carlostome: I've seen the symbol ≃ used for Kleene equality (https://en.wikipedia.org/wiki/Kleene_equality).

Since the notion of equality  we use is even weaker, I'm leaning towards ≊. We can think of every ~ as weakening the notion of equality.

carlostome: I've seen ≛ and ≜ being used to denote 'equal by definition'.

williamdemeo: I've see `≜` and `:=` (and sometimes an equal with a little `def` on top of it) used for "equal by definition," but I've never see `≛` used for that purpose.  I think we should avoid using `≜` for anything other than definitional equality... and even for that purpose I don't like it.  I agree with Andre that there so are many equality symbols to choose from, so we can probably find one that avoids using tildes.  Having said that, I have no problem with using tildes for this.

Tildes are used all over the place and for so many different notions of equality that one can never assume one knows in advance what it means.  I think that's actually what we want here---the reader needs to look at the definition to know what we mean by whatever symbol we choose for this "equality."

carlostome: > I've seen ≛ 

The symbol displays differently in a code block:
≛ vs `≛`

I definitely have not seen ≛ used for eq. by def.

src/Ledger/Ledger.lagda
```diff
@@ -73,47 +81,55 @@ allColdCreds govSt es =
...
-       Γ ⊢ s ⇀⦇ tx ,LEDGER⦈ ⟦ utxoSt' , govSt , certState ⟧
+    ∙ isValid tx ≡ true
+    ∙ ⟦ slot , pp , Γ .treasury ⟧  ⊢ utxoSt ⇀⦇ tx ,UTXOW⦈ utxoSt'
+    ∙ ⟦ epoch slot , pp , txvote , txwdrls , allColdCreds govSt enactState ⟧ ⊢ certState ⇀⦇ txcerts ,CERTS⦈ certState'
+    ∙ ⟦ txid , epoch slot , pp , ppolicy , enactState , certState' , dom rewards ⟧ ⊢ rmOrphanDRepVotes certState' govSt ⇀⦇ txgov txb ,GOVS⦈ govSt'
+      ────────────────────────────────
+      Γ ⊢ ⟦ utxoSt , govSt , certState ⟧ ⇀⦇ tx ,LEDGER⦈ ⟦ utxoSt' , govSt' , certState' ⟧
+
+  LEDGER-I :
+    let txb = tx .body
```
WhatisRT, line 120: Is `txb` unused here, or am I blind?

src/Ledger/Ledger.lagda
```diff
@@ -73,47 +81,55 @@ allColdCreds govSt es =
...
+    let  txb         = tx .body
+\end{code}
+\begin{code}[hide]
+         open TxBody txb
+\end{code}
+\begin{code}
+         pp          = Γ .pparams
+         slot        = Γ .slot
+         ppolicy     = Γ .ppolicy
+         enactState  = Γ .enactState
```
WhatisRT, line 109: I haven't tried it, but I feel this would look better as a fake pattern-match below, now that `Γ` is only used via its fields.

src/Ledger/Ledger/Properties.agda
```diff
@@ -264,40 +264,16 @@ module LEDGER-PROPS (tx : Tx) (Γ : LEnv) (s : LState) where
...
-        ≡⟨ map-++ _ ([ g ] |ᵒ certState) (govSt |ᵒ certState) ⟩
-      dpMap ([ g ] |ᵒ certState) ++ dpMap (govSt |ᵒ certState)
-        ≡⟨ cong (dpMap ([ g ] |ᵒ certState) ++_) (dpMap-|ᵒ govSt certState) ⟩
-      dpMap ([ g ] |ᵒ certState) ++ dpMap govSt
-        ≡⟨ cong (_++ dpMap govSt) (dpMap-|ᵒ-singleton g certState) ⟩
-      dpMap [ g ] ++ dpMap govSt
-        ≡˘⟨ map-++ _ [ g ] govSt ⟩
-      dpMap (g ∷ govSt) ∎
+    dpMap-rmOrphanDRepVotes : ∀ certState govSt → dpMap (rmOrphanDRepVotes certState govSt) ≡ dpMap govSt
+    dpMap-rmOrphanDRepVotes certState govSt = sym (fmap-∘ govSt) -- map proj₁ ∘ map (map₂ _) ≡ map (proj₁ ∘ map₂ _) ≡ map proj₁
```
WhatisRT, line 268: Nice!

src/Ledger/Ratify.lagda
```diff
@@ -546,55 +546,62 @@ an argument.
...
   d : Bool
 
+open RatifyEnv
+open GovActionState
 \end{code}
 \begin{figure*}[ht]
 \begin{AgdaSuppressSpace}
 \begin{code}
-data _⊢_⇀⦇_,RATIFY'⦈_ :
+data _⊢_⇀⦇_,RATIFIES'⦈_ :
```
WhatisRT, line 561: Shouldn't this be RATIFY?

src/Ledger/Utxo.lagda
```diff
@@ -485,32 +495,45 @@ data
...
-    ∀ {Γ} {s} {tx}
-    → let open Tx tx renaming (body to txb); open TxBody txb
-          open UTxOEnv Γ renaming (pparams to pp)
-          open UTxOState s
-          sLst = collectPhaseTwoScriptInputs pp tx utxo
+    let  pp         = Γ .pparams
+         utxo       = s .utxo
+         fees       = s .fees
+         deposits   = s .deposits
+         donations  = s .donations
```
WhatisRT, line 505: I think this would be nicer as a fake pattern-match below. It has the added benefit that the fields of the state would line up nicely.

src/Ledger/Utxo.lagda
```diff
@@ -485,32 +495,45 @@ data
...
           ────────────────────────────────
           Γ ⊢ s ⇀⦇ tx ,UTXOS⦈ ⟦ (utxo ∣ txins ᶜ) ∪ˡ (outs txb) , fees + txfee , updateDeposits pp txb deposits , donations + txdonation ⟧
-
   Scripts-No :
-    ∀ {Γ} {s} {tx}
-    → let open Tx tx renaming (body to txb); open TxBody txb
-          open UTxOEnv Γ renaming (pparams to pp)
-          open UTxOState s
-          sLst = collectPhaseTwoScriptInputs pp tx utxo
+    let  pp         = Γ .pparams
```
WhatisRT, line 519: Same as above

WhatisRT (approved):
Thank you, this looks fantastic! Sorry for taking so long, I wanted to be thorough and got distracted a few times.

I made some comments here and there, they should hopefully all be relatively easy changes if you do want to make them. Feel free to not do any of them, but please leave a comment or make an issue if you think it's worth doing some other time.

src/Ledger/Certs.lagda
```diff
@@ -309,34 +310,33 @@ _⊢_⇀⦇_,CERTS⦈_ = ReflexiveTransitiveClosureᵇ' {_⊢_⇀⟦_⟧ᵇ_ = _
...
 \begin{figure*}[h]
 \begin{AgdaSuppressSpace}
 \begin{code}[hide]
 data _⊢_⇀⦇_,DELEG⦈_ where
 \end{code}
 \begin{code}
-  DELEG-delegate : let open PParams pp in
-    ∙ (c ∉ dom rwds → d ≡ keyDeposit)
+  DELEG-delegate :
+    let Γ = ⟦ pp , pools , delegatees ⟧
```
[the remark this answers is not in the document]
carlostome: Yes, otherwise there is a very strange stretching

src/Ledger/Fees.lagda
```diff
@@ -7,80 +7,100 @@ open import Ledger.Transaction
...
-_↑ℚ : ℕ → ℚ
-n ↑ℚ = mkℚ+ n 1 (1-coprimeTo n ∘ swap)
-\end{code}
 \begin{code}
 scriptsCost : (pp : PParams) → ℕ → Coin
-scriptsCost pp scSz with (PParams.refScriptCostStride pp)
+scriptsCost pp scriptSize
+\end{code}
+\begin{code}[hide]
+  with (PParams.refScriptCostStride pp)
```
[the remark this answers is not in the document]
carlostome: Using with let's use have the equality symbol underneath
```
  = scriptsCostAux 0ℚ minFeeRefScriptCoinsPerByte scriptSize
```

src/Ledger/Ratify.lagda
carlostome: Good catch!

src/Ledger/Utxo.lagda
carlostome: What do you mean by fake pattern-match?

src/Ledger/Ledger.lagda
carlostome: Indeed

src/Prelude.agda
carlostome: I agree that a simple `~` will not lead to possible confusions with equality and will force people to look it up.

src/Ledger/Utxo.lagda
WhatisRT: I mean to construct `s` as an index in the conclusion, like with the final state, i.e.
```
Γ ⊢ ⟦ utxo , fees , deposits , donations ⟧ ⇀⦇ tx ,UTXOS⦈ ⟦ (utxo ∣ txins ᶜ) ∪ˡ (outs txb) , fees + txfee , updateDeposits pp txb deposits , donations + txdonation ⟧
```
I call it fake pattern match, because it looks and feels like one (as if pattern-matching on `s`) but it isn't.

src/Ledger/Fees.lagda
williamdemeo: That may be my fault.  I suppose it should say something like, "the sum total of the sizes of the reference scripts."  @WhatisRT could you confirm?

src/Ledger/Utxo.lagda
williamdemeo: That looks reasonable to me.  Actually, both options look good to me, but I think what Andre suggests is more consistent with the other step rules.

williamdemeo:
This all looks good to me.  I would submit an approval, but Andre has already done so.  LGTM! 🚀

WhatisRT:
Seems LaTeX is unhappy with the tilde: https://github.com/IntersectMBO/formal-ledger-specifications/actions/runs/13585219494/job/37978592462?pr=695#step:4:8378

src/Ledger/Fees.lagda
WhatisRT: Yeah, the explanation should be adjusted then

Issue #701, opened 2025-02-24
Define _~_ (in Gov.lagda) primitively


carlostome:
Currently the definition depends on `govActionPriority` but the concept does not.

carlostome:
After #695 this is named `Overlap`

carlostome:
Addressed by #703

PR #703, opened 2025-02-27
Refactor GovernanceActions using dep. types

carlostome:
# Description

This PR is a proof-of-concept refactor of GovernanceActions to use dependent types.

The motivation is the myriad of functions which are defined by pattern matching on arguments of type `GovAction` but only care about the name of the constructor (see `NeedsHash` or `govActionPriority`, e.g.)

Dependent types in this case let us express in a natural way:
1. The type of governance action
2. The data associated with a given type of governance action

In addition, addresses #701 

# Conformance testing

- [x] Tests pass on commit 81637a1c2

# Checklist

- [x] Self-reviewed the diff


src/Ledger/GovernanceActions.lagda
```diff
@@ -54,14 +54,27 @@ record Anchor : Type where
...
+  NoConfidence     : GovActionType
+  UpdateCommittee  : GovActionType
+  NewConstitution  : GovActionType
+  TriggerHF        : GovActionType
+  ChangePParams    : GovActionType
+  TreasuryWdrl     : GovActionType
+  Info             : GovActionType
+
+GovActionData : GovActionType → Type
+GovActionData = λ where
```
WhatisRT, line 67: I think this lambda is a bit too unusual in this place. People are probably going to understand it, but they might be confused why it wasn't written in the "normal" way like everything else. I don't think it's too repetitive to match on the LHS here.

WhatisRT (commented):
Looks pretty good overall, my only complaint is that there's a lot of `proj₁` going on. If you're a bit out of context, that could be anything (and I think we already have too many `proj₁`s, though maybe not in the parts that are visible in the PDF).

I think in general just some other name would be quite helpful, maybe with a pattern alias? Maybe `gaType`? But I also wonder if some functions that only depend on the type shouldn't still get the entire action. The thing that specifically worries me a bit is `hasParent : EnactState → GovState → (a : GovAction) → NeedsHash (a .proj₁) → Type`. Maybe it's fine if it isn't a `proj₁`, but it might be too much computation going on in the type.

Another option for this type in particular might be to match on `a`: `hasParent : EnactState → GovState → ((t , c) : GovAction) → NeedsHash t → Type`, I don't know if that's better or worse though.

carlostome:
> Looks pretty good overall, my only complaint is that there's a lot of `proj₁` going on. If you're a bit out of context, that could be anything (and I think we already have too many `proj₁`s, though maybe not in the parts that are visible in the PDF).

I agree with you. The proper thing to do I believe is not use a sigma type but a record type with self explanatory field names.

> I also wonder if some functions that only depend on the type shouldn't still get the entire action

I'll do another pass to check that functions really get what they need and no more.

src/Ledger/Gov/Properties.agda
```diff
@@ -150,10 +150,10 @@ instance
         computeProof = case H of λ where
           (yes (wf , av , dep , vHFA , HasParent' en , goodAddr , regReturn) , yes (new , rem , q , refl)) →
             case ¿ ∀[ e ∈ range new ] epoch < e × dom new ∩ rem ≡ᵉ ∅ ¿ of λ where
-              (yes newOk) → success (_ , GOV-Propose (wf , av , dep , vHFA , en , goodAddr , regReturn))
+              (yes newOk) → success (-, GOV-Propose {_} {_} {_} {_} {_} {_} {_} {_} {_} {(new , rem , q)} (wf , av , dep , vHFA , en , goodAddr , regReturn))
```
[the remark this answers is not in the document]
carlostome: It is not obvious what name is supposed to have:
```agda
 (-, GOV-Propose {Γ = _} {addr = _} {a = _} {achr = _} {p = _} {d = _} {prev = _} {s = _} {k = _} {(new , rem , q)} (wf , av , dep , vHFA , en , goodAddr , returnReg))
```
with the definition of `Gov-Propose` being:
```agda
  GOV-Propose :
    let pp           = Γ .pparams
        e            = Γ .epoch
        enactState   = Γ .enactState
        rewardCreds  = Γ .rewardCreds
        prop         = record { returnAddr = addr ; action = a ; anchor = achr
                              ; policy = p ; deposit = d ; prevAction = prev }
    in
    ∙ actionWellFormed a
    ∙ actionValid rewardCreds p (Γ .ppolicy) e a
    ∙ d ≡ pp .govActionDeposit
    ∙ validHFAction prop s enactState
    ∙ hasParent enactState s (a .gaType) prev
    ∙ addr .RwdAddr.net ≡ NetworkId
    ∙ addr .RwdAddr.stake ∈ rewardCreds
      ───────────────────────────────────────
      (Γ , k) ⊢ s ⇀⦇ inj₂ prop ,GOV⦈ addAction s (pp .govActionLifetime +ᵉ e)
                                                 (Γ .txid , k) addr a prev
```
It seems something Agda-funky is going on.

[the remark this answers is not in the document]
WhatisRT: I tried to give it a name, and this is the error message I got:
```
Function does not accept argument {x = _}
(possible arguments: prev.a.1)
when checking that {x = new , rem , q}
(wf , av , dep , vHFA , en , goodAddr , regReturn) are valid
arguments to a function of type
{prev.a.1 : GovActionData (_a_621 .proj₁)} →
...
```
Never seen that kind of name before, but literally writing `prev.a.1` was unsurprisingly a parse error. I think this comes from Agda's feature that a `variable` can introduce extra variables, in this case `prev : NeedsHash (proj₁ a)` needs an `a : GovAction`. And somehow that needs a `1`? Not sure what's going on there. Maybe it's just something to disambiguate these generated names. Or maybe there's some kind of unification going on, which fixes `proj₁ a`, but then the other part isn't fixed and needs to be quantified over, and that's where the name comes from. Very mysterious.

src/Ledger/Gov.lagda
```diff
@@ -99,42 +99,89 @@ open PState
...
+Overlap NoConfidence    UpdateCommittee  = ⊤
+Overlap UpdateCommittee NoConfidence     = ⊤
+Overlap a               a'               = a ≡ a'
 \end{code}
 \begin{code}[hide]
-Overlap? : (a a' : GovAction) → Dec (Overlap a a')
-Overlap? a a' = let n = govActionPriority a
-                    m = govActionPriority a'
-                in n ≟ m ⊎-dec (n ≟ 0 ×-dec m ≟ 1) ⊎-dec (n ≟ 1 ×-dec m ≟ 0)
+-- TODO: cleanup this
```
WhatisRT, line 117: This has come up before in a similar scenario on another project, and I think the answer is to make the `match` tactic (from `Tactic.Match` in stdlib-meta) work properly. Then you could just write something like `Overlap? = by $ match $ match $ anyOf (quote (Dec-⊤ .dec) ∷ quote (yes refl) ∷ quote (no (λ ())) ∷ [])`.

WhatisRT (approved):
LGTM

src/Ledger/Gov.lagda
carlostome: Opened an [issue](https://github.com/agda/agda-stdlib-meta/issues/27) in agda-stdlib-meta regarding this.

Issue #706, opened 2025-03-05
Add `createRUpd`


HeinrichApfelmus:
This issue is about progressing the specification of the reward calculation by adding the `createRUpd` function from the Shelley specification (Figure 51 "Reward Update Creation").

### Comments

* In the Shelley specification, the `RUPD` rule is used to apply the result of `createRUpd` to an epoch state. However, the `RUPD` rule is requested only by the `TICK` rule, which may now be within the purview of the new [Consensus specification](https://ouroboros-consensus.cardano.intersectmbo.org/assets/files/consensus-spec-agda-fdee8d65f730471bd62e2177650a579d.pdf) instead of this repository. I will sidestep this problem by focusing only on `createRUpd` here.
* See also #675

javierdiaz72:
@HeinrichApfelmus: Just as a follow-up to the discussion in today's Cardano FM Office Hours meeting: 

1. The scope of the current Consensus specification is block header validation only, i.e. basically Section 15.1 from the Shelley Ledger specification. That's why TICKF is used instead of TICK (and thus no RUPD is called).

2. The consensus implementation *does* call TICK and then calls BBODY (both exposed by the ledger implementation). Then, in some sense, the consensus implementation implements CHAIN.

HeinrichApfelmus:
Closing, as `createRUpd` has been added in #752.

PR #707, opened 2025-03-08
Add monetary and treasury expansion to `PParams` (old)

HeinrichApfelmus:
# Description

This pull request adds the monetary expansion `rho` and treasury expansion `tau` protocol parameters to the `PParams` type.

In order to be able to prove that the treasury and reserves are not negative, it is necessary to constrain these parameters need to the unit interval `[0, 1]`. These conditions need to be stored somewhere — I have chosen to make them part of the `PParams` type by refining the parameters to a new type `UnitInterval`. (This is similar to how the previous specification of the Shelley ledger encodes these parameters.)

Context: #706 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Class/DecEq/Instances/Extra.agda
+++ b/src/Class/DecEq/Instances/Extra.agda
@@ -1,0 +1,26 @@
+{-# OPTIONS --safe #-}
+
+-- Extra instances for decidable equality.
+module Class.DecEq.Instances.Extra where
+
+open import Data.Refinement
+  using (Refinement; value; _,_)
+open import Function
+  using (_∘_)
+open import Relation.Binary.PropositionalEquality
+  using (cong; refl)
+open import Relation.Nullary.Decidable
+  using (no; yes)
+
+open import Class.DecEq
+  using (DecEq; _≟_)
+
+-- Equality for a Refinement type is decide if the equality
+-- for the type to be refined is decidable.
+DecEq-Refinement
+    : ∀ {a p} (A : Set a) (P : A → Set p)
+    → ⦃ DecEq A ⦄ → DecEq (Refinement A P)
+DecEq-Refinement A P ._≟_ (x , px) (y , py)
+  with x ≟ y
+... | no neq = no (neq ∘ cong value)
+... | yes refl = yes refl
--- a/src/Ledger/BaseTypes.lagda
+++ b/src/Ledger/BaseTypes.lagda
@@ -7,7 +7,9 @@
 
 module Ledger.BaseTypes where
 
-open import Prelude using (ℕ)
+open import Prelude using (ℕ; _×_)
+open import Data.Rational using (ℚ; 0ℚ; 1ℚ; _≤_)
+open import Data.Refinement using (Refinement-syntax)
 \end{code}
 
 \begin{code}[hide]
@@ -20,4 +22,11 @@ private
   Epoch  = ℕ
 \end{code}
 \caption{Some basic types used in many places in the ledger}
+\end{figure*}
+
+\begin{figure*}[h]
+\begin{code}
+  UnitInterval = [ x ∈ ℚ ∣ (0ℚ ≤ x) × (x ≤ 1ℚ) ]
+\end{code}
+\caption{Refinement types used in some places in the ledger}
 \end{figure*}
--- a/src/Ledger/PParams.lagda
+++ b/src/Ledger/PParams.lagda
@@ -23,6 +23,7 @@ open import Ledger.Prelude
 open import Ledger.Crypto
 open import Ledger.Script
 open import Ledger.Types.Epoch
+open import Ledger.Types.Numeric using (UnitInterval)
 
 module Ledger.PParams
   (crypto : Crypto )
@@ -125,6 +126,8 @@ record PParams : Type where
         b                             : ℕ
         keyDeposit                    : Coin
         poolDeposit                   : Coin
+        monetaryExpansion             : UnitInterval -- formerly: rho
+        treasuryCut                   : UnitInterval -- formerly: tau
         coinsPerUTxOByte              : Coin
         prices                        : Prices
         minFeeRefScriptCoinsPerByte   : ℚ
@@ -229,6 +232,8 @@ module PParamsUpdate where
           a b                           : Maybe ℕ
           keyDeposit                    : Maybe Coin
           poolDeposit                   : Maybe Coin
+          monetaryExpansion             : Maybe UnitInterval
+          treasuryCut                   : Maybe UnitInterval
           coinsPerUTxOByte              : Maybe Coin
           prices                        : Maybe Prices
           minFeeRefScriptCoinsPerByte   : Maybe ℚ
@@ -279,6 +284,8 @@ module PParamsUpdate where
       ∷ is-just b
       ∷ is-just keyDeposit
       ∷ is-just poolDeposit
+      ∷ is-just monetaryExpansion
+      ∷ is-just treasuryCut
       ∷ is-just coinsPerUTxOByte
       ∷ is-just minFeeRefScriptCoinsPerByte
       ∷ is-just maxRefScriptSizePerTx
@@ -370,6 +377,8 @@ module PParamsUpdate where
       ; b                           = U.b ?↗ P.b
       ; keyDeposit                  = U.keyDeposit ?↗ P.keyDeposit
       ; poolDeposit                 = U.poolDeposit ?↗ P.poolDeposit
+      ; monetaryExpansion           = U.monetaryExpansion ?↗ P.monetaryExpansion
+      ; treasuryCut                 = U.treasuryCut ?↗ P.treasuryCut
       ; coinsPerUTxOByte            = U.coinsPerUTxOByte ?↗ P.coinsPerUTxOByte
       ; minFeeRefScriptCoinsPerByte = U.minFeeRefScriptCoinsPerByte ?↗ P.minFeeRefScriptCoinsPerByte
       ; maxRefScriptSizePerTx       = U.maxRefScriptSizePerTx ?↗ P.maxRefScriptSizePerTx
--- a/src/Ledger/Types/Numeric.agda
+++ b/src/Ledger/Types/Numeric.agda
@@ -1,0 +1,6 @@
+{-# OPTIONS --safe #-}
+
+-- Subsets of number types useful for the ledger specification.
+module Ledger.Types.Numeric where
+
+open import Ledger.Types.Numeric.UnitInterval public
--- a/src/ScriptVerification/Lib.agda
+++ b/src/ScriptVerification/Lib.agda
@@ -12,7 +12,9 @@ open import Ledger.Conway.Conformance.Utxo SVTransactionStructure SVAbstractFunc
 open import Ledger.Transaction
 open TransactionStructure SVTransactionStructure
 open import Ledger.Types.Epoch
+open import Ledger.Types.Numeric using (mkUnitInterval)
 open EpochStructure SVEpochStructure
+open import Data.Integer using (ℤ; +_)
 open import Data.Rational using (½; 1ℚ ; mkℚ+ ; _/_)
 open import Data.Nat.Coprimality using (Coprime; gcd≡1⇒coprime)
 open Implementation
@@ -33,6 +35,8 @@ createEnv s = record { slot = s ; treasury = 0 ;
                                ; minUTxOValue = 0
                                ; poolDeposit = 500000000 -- lovelace
                                ; keyDeposit = 500000000 -- lovelace
+                               ; monetaryExpansion = mkUnitInterval (+ 3 / 1000) refl
+                               ; treasuryCut = mkUnitInterval (+ 2 / 10) refl
                                ; coinsPerUTxOByte = 0  --lovelace
                                -- ^^^ was 4310, but that value resulted in failed tests once
                                -- we replaced `minUTxOValue` with `coinsPerUTxOByte`; see `HelloWorld.agda`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### WIP
 
+- Add `treasuryCut` and `monetaryExpansion` to `PParams`
 - Change the `DELEG-dereg` transition so that the deposit field can be empty
 - Require witnessing of `reg` credential if the deposit is non-zero
 - Add witnessing of collaterals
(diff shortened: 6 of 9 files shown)


HeinrichApfelmus:
One problem introduced by the definition `UnitInterval = Refinement ℚ …` is that I can no longer use the `autoHsType` and `autoConvert` macros, which are presumably needed by the conformance test suite.

However, this is a symptom of an underlying problem: The `Refinement` type mixes computational and logical aspects, the invariant that the number is constraint to the unit interval is now mixed into the code. There is no direct Haskell equivalent for this type!

(While working with agda2hs, I found that such invariants typically require irrelevance `.` and erasure `@0` to work well with Haskell code. If you ask me, the merging of code and logic in Agda has swung a bit too far into the "everything is code" direction.)

@WhatisRT — thoughts?

carlostome:
> One problem introduced by the definition `UnitInterval = Refinement ℚ …` is that I can no longer use the `autoHsType` and `autoConvert` macros, which are presumably needed by the conformance test suite.

They are not _needed_, these macros are convenient to derive the Haskell datatypes but it could be done "by hand".

> However, this is a symptom of an underlying problem: The `Refinement` type mixes computational and logical aspects: The invariant that the number is constraint to the unit interval is now mixed into the code. There is no direct Haskell equivalent for this type!

This problem already occurs in some places in the codebase. For example see `GovProposal` in the Ledger and in conformance:
- https://github.com/IntersectMBO/formal-ledger-specifications/blob/2cd9668057e09dcecb06b6c5adeeeda9f3301868/src/Ledger/GovernanceActions.lagda#L174 
- https://github.com/IntersectMBO/formal-ledger-specifications/blob/2cd9668057e09dcecb06b6c5adeeeda9f3301868/src/Ledger/Conway/Foreign/HSLedger/Gov.agda#L49

In a nutshell, the current approach is to define a nondependent version of the type, which can be compiled to Haskell, and define conversion functions from and to the dependent version. 
 - https://github.com/IntersectMBO/formal-ledger-specifications/blob/2cd9668057e09dcecb06b6c5adeeeda9f3301868/src/Ledger/Conway/Foreign/HSLedger/Gov.agda#L73

WhatisRT:
In these cases, we've usually just embraced that the FFI part is unsafe anyway, see e.g. [here](https://github.com/IntersectMBO/formal-ledger-specifications/blob/master/src/Ledger/Conway/Foreign/HSLedger/NewEpoch.agda#L43). This would work just fine in that case as well. That way, the code is neatly split into a safe part that has the required invariants and an unsafe part that will just crash if you actually do anything bad. If we want the caller to be able to handle that error we'd have to restructure `Convertible`, but I don't think we care right now (and maybe will never).

We even have some occurances of `coerce ⦃ TrustMe ⦄` (`unsafeCoerce`) in our FFI, so this is relatively harmless.

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+inUnitInterval x = (0ℚ ≤ x) × (x ≤ 1ℚ)
+
+-- Decide whether a rational number is in the unit interval.
+isInUnitInterval : (x : ℚ) → Dec (inUnitInterval x)
+isInUnitInterval x
+  with 0ℚ ≤? x
+  with x  ≤? 1ℚ
+... | no  ¬p  | q       = no λ prf → ¬p (proj₁ prf)
+... | yes 0≤x | no ¬q   = no λ prf → ¬q (proj₂ prf)
+... | yes 0≤x | yes x≤1 = yes (0≤x , x≤1)
```
carlostome, line 37:
```suggestion
isInUnitInterval x = (0ℚ ≤? x) ×-dec (x ≤? 1ℚ)
```

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+prop-inUnitInterval-*
+  : ∀ (x y : ℚ)
+  → inUnitInterval x
+  → inUnitInterval y
+  → inUnitInterval (x ℚ.* y)
+--
+prop-inUnitInterval-* x y ux (0≤y , y≤1) =
+  ( prop-inUnitInterval-*-0≤ x y ux 0≤y , x*y≤1)
+  where
+  x*y≤1 = ℚ.≤-trans (prop-inUnitInterval-*-≤y x y ux 0≤y) y≤1
```
carlostome, line 74: I'm surprised this property exists without defining multiplication of `UnitInterval` numbers. 
If its not going to be used I suggest not to include it (and the lemmas it depends on)

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+  with lemma ← ℚ.*-monoʳ-≤-nonNeg y ⦃ ℚ.nonNegative 0≤y ⦄ x≤1
+  rewrite ℚ.*-identityˡ y
+  = lemma
+
+-- Multiplying with a number from the unit interval stays positive.
+prop-inUnitInterval-*-0≤
+  : ∀ (x y : ℚ)
+  → inUnitInterval x
+  → 0ℚ ≤ y
+  → 0ℚ ≤ x ℚ.* y
```
carlostome, line 56: I feel this prop is just plumbing and thus should be inlined or placed in a `where` clause within `prop-inUnitInterval-*`

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+
+-- Equality for a Refinement type is decide if the equality
+-- for the type to be refined is decidable.
+DeqEq-Refinement
+    : ∀ {a p} (A : Set a) (P : A → Set p)
+    → ⦃ DecEq A ⦄ → DecEq (Refinement A P)
+DeqEq-Refinement A P ._≟_ (x , px) (y , py)
+  with x ≟ y
+... | no neq = no (neq ∘ cong value)
+... | yes refl = yes refl
```
carlostome, line 92: This should not be here. Maybe move it to a module `Class.DecEq.Instances.Extra` while it is not added to agda-stdlib-classes.

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+
+-- Helper function to make an element of the Refinement type
+-- while having the Adga compiler compute the evidence from Dec.
+-- Usage example:  mkUnitInterval (+ 2 / 10) refl
+mkUnitInterval
+  : ∀ (x : ℚ) → does (isInUnitInterval x) ≡ true → UnitInterval
+mkUnitInterval x evidence
+  with isInUnitInterval x
+... | yes p = x , [ p ]
+
```
carlostome, line 123: I get an Agda warning in this line:
```
Incomplete pattern matching for Ledger.Types.Numeric.with-.
Missing cases:
  mkUnitInterval x evidence | .false because ofⁿ ¬a
when checking the definition of Ledger.Types.Numeric.with-
```

src/Ledger/Types/Numeric.agda
```diff
@@ -0,0 +1,136 @@
...
+    Properties
+----------------------------------------------------------------------}
+-- to/from is the identity
+prop-toUnitInterval-fromUnitInterval
+  : ∀ (x : UnitInterval)
+  → toUnitInterval (fromUnitInterval x) ≡ just x
+--
+prop-toUnitInterval-fromUnitInterval (x , [ p0 ])
+  with isInUnitInterval x
+... | no ¬p = case [ ¬p p0 ] of λ ()
```
carlostome, line 135:
```suggestion
... | no ¬p = ⊥-elim-irr (¬p p0)
```

src/ScriptVerification/Lib.agda
```diff
@@ -33,6 +35,8 @@ createEnv s = record { slot = s ; treasury = 0 ;
                                ; minUTxOValue = 0
                                ; poolDeposit = 500000000 -- lovelace
                                ; keyDeposit = 500000000 -- lovelace
+                               ; tau = mkUnitInterval (+ 2 / 10) refl
+                               ; rho = mkUnitInterval (+ 3 / 1000) refl
```
carlostome, line 39: Would be nice if we can make mkUnitInterval deduce (or use a tactic) for the refl part

carlostome (commented):
Good work, here are some comments:

- I think the name `UnitInterval` is a bit misleading since it is using rationals and not reals. The best alternative I can come up atm is to use the symbol ℚ withing the name: `UnitIntervalℚ` or `ℚUnitInterval`.

- Regarding the module structure, I suggest to add another level in the hierarchy, `Ledger.Types.Numeric.UnitInterval`, since in the future we might want to use a similar approach for positive naturals etc.

src/Ledger/Types/Numeric.agda
HeinrichApfelmus: I don't remember seeing this warning before — it's probably due to the change of definition of `isInUnitInterval`? 🤔 But I can discharge the proof obligation.

src/Ledger/Types/Numeric.agda
HeinrichApfelmus: I will probably need this lemma when proving that the rewards update is nonnegative and bounded by the available rewards, that's why I have added it in advance.

HeinrichApfelmus:
Thanks for the review, @carlostome ! 😊 

> Regarding the module structure, I suggest to add another level in the hierarchy, `Ledger.Types.Numeric.UnitInterval`, since in the future we might want to use a similar approach for positive naturals etc.

Ok, done.

> I think the name `UnitInterval` is a bit misleading since it is using rationals and not reals. The best alternative I can come up atm is to use the symbol ℚ withing the name: `UnitIntervalℚ` or `ℚUnitInterval`.

I can do that if you really want me to, but I'm not sure if that's a good idea:

* As far as I can tell, `agda-stdlib` currently does not try to define constructive real numbers (but see https://github.com/agda/agda-stdlib/pull/2487). I think it's fair to tacitly assume that the real numbers are not used in software, and that any type with this name will have an "obvious" finite representation, here rational numbers.
* The `cardano-ledger` Haskell code uses a type called [UnitInterval]( https://github.com/IntersectMBO/cardano-ledger/blob/a3ab599f66da96b714e3c8a5aa9596ea9275af8b/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs#L469-L471) for the same purpose — though that type contains an additional size restriction.

src/Ledger/Types/Numeric.agda
carlostome: My bad, I should have double checked I didn't introduce the warning

src/Ledger/BaseTypes.lagda
```diff
@@ -21,3 +23,10 @@ private
 \end{code}
 \caption{Some basic types used in many places in the ledger}
 \end{figure*}
+
+\begin{figure*}[h]
+\begin{code}
+  UnitInterval = [ x ∈ ℚ ∣ (0ℚ ≤ x) × (x ≤ 1ℚ) ]
```
WhatisRT, line 29: It's probably a good idea to use literal overloading, so you can just write `0` and `1`. To do that, you have to add the `Number` instance from `Data.Rational.Literals`, ideally to the `Prelude` module next to the instance for `ℕ`. Then, just add `open import Agda.Builtin.FromNat` to any module that should use those overloaded literals.

src/Class/DecEq/Instances/Extra.agda
```diff
@@ -0,0 +1,26 @@
...
+  using (cong; refl)
+open import Relation.Nullary.Decidable
+  using (no; yes)
+
+open import Class.DecEq
+  using (DecEq; _≟_)
+
+-- Equality for a Refinement type is decide if the equality
+-- for the type to be refined is decidable.
+DeqEq-Refinement
```
WhatisRT, line 20: It'd be fantastic to upstream this to https://github.com/agda/agda-stdlib-classes

src/ScriptVerification/Lib.agda
WhatisRT: You can give `mkUnitInterval` a tactic argument:
```
open import Tactic.Defaults
open import Tactic.Constrs

mkUnitInterval : ∀ (x : ℚ) → {@(tactic tryConstrs 10) : does (isInUnitInterval x) ≡ true} → UnitInterval
```
Then you can write `mkUnitInterval (+ 3 / 1000)` and it'll find `refl` automatically. And in case `refl` doesn't work, you can still supply the proof manually: `mkUnitInterval x {myProof}`.

src/Ledger/PParams.lagda
```diff
@@ -125,6 +126,8 @@ record PParams : Type where
         b                             : ℕ
         keyDeposit                    : Coin
         poolDeposit                   : Coin
+        tau                           : UnitInterval -- treasury expansion
+        rho                           : UnitInterval -- monetary expansion
```
WhatisRT, line 130: I think it might be best to change those to the more approachable names. People are already complaining about short meaningless names.
```suggestion
        treasuryCut                           : UnitInterval
        monetaryExpansion                           : UnitInterval
```

HeinrichApfelmus: Ok, I can do that — the only trouble is that this would be incompatible with the existing nomenclature in the Shelley genesis file, though, so people will have to be instructed to use the new names.

src/ScriptVerification/Lib.agda
HeinrichApfelmus: Do you want me to add this implicit tactic argument?

src/Ledger/BaseTypes.lagda
HeinrichApfelmus: I prefer `0` and `1` as well, I will give it a try.

HeinrichApfelmus:
Thanks @williamdemeo and @WhatisRT !

> I'm just going to push some style mods to your branch. I hope you don't mind. (Just some things to make the code more consistent with the existing code.) If you don't agree with any of the changes, feel free to comment and/or revert them.

Thanks! I am only going to change

* overly long indentation in the body of `inUnitInterval-*`

As for the `prop-` prefix — I have converged on prefixing all logical statements with `prop-` in order to make them more easily distinguishable from computations, but that was in the context of adga2hs where I had a clear separation between target language (Haskell) and metalanguage (Agda). I'll use what you prefer here.

src/Class/DecEq/Instances/Extra.agda
HeinrichApfelmus: Ok, but not in the present pull request in order to keep the scope limited.

I have created an issue https://github.com/agda/agda-stdlib-classes/issues/17

src/ScriptVerification/Lib.agda
HeinrichApfelmus: The `tryConst` tactic does not quite work — the argument to the `tactic` annotation must be of type `Term → TC ⊤`. I'll stick to `refl` for now.

HeinrichApfelmus:
For technical reasons (CI step "MAlonzo"), I'm closing this pull request and I'm opening a new one with the code after review at #732.

PR #710, opened 2025-03-11
Made `CERTBASE` get applied at the end of `CERTS`

### LINKED ISSUE #635: Conformance failure: DRep delegations are not removed after the DRep is deregistered
I'm getting a conformance failure in `/Conway/Conformance/ImpTests/Conway Imp conformance/LEDGER/Withdraw from a key delegated to an unregistered DRep/` where there's a mismatch in the `dsVoteDelegs` field after submitting a transaction that deregisters a DRep which has delegations. 

The spec does not remove the delegations, but the implementation does and so there's a mismatch in the state after the transaction because of that.

Soupstraw:
# Description

Previously the `CERTBASE` rule was applied to the state before the `CERT` transitions happened, but this caused the DRep delegations not get removed when the DRep was removed with the `deregdrep` certificate.

I've modified `CERTS` so that `CERTBASE` now gets applied last, after all of the certificates have been applied. I also moved reward withdrawals from `CERTS` to `LEDGER`.

closes #635
 
# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Certs.lagda
+++ b/src/Ledger/Certs.lagda
@@ -466,14 +466,12 @@ data _⊢_⇀⦇_,CERTBASE⦈_ where
         validVoteDelegs  = voteDelegs ∣^ (  mapˢ (credVoter DRep) (dom dReps)
                                         ∪ fromList (noConfidenceRep ∷ abstainRep ∷ []) )
     in
-    ∙ mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ
-      ────────────────────────────────
       ⟦ e , pp , vs , wdrls , cc ⟧ ⊢
         ⟦ ⟦ voteDelegs , stakeDelegs , rewards ⟧
         , stᵖ
         , ⟦ dReps , ccHotKeys ⟧
         ⟧ ⇀⦇ _ ,CERTBASE⦈
-        ⟦ ⟦ validVoteDelegs , stakeDelegs , constMap wdrlCreds 0 ∪ˡ rewards ⟧
+        ⟦ ⟦ validVoteDelegs , stakeDelegs , rewards ⟧
         , stᵖ
         , ⟦ refreshedDReps , ccHotKeys ⟧
         ⟧
--- a/src/Ledger/Certs.lagda
+++ b/src/Ledger/Certs.lagda
@@ -320,7 +320,7 @@ module _ where  -- achieves uniform (but perhaps misleading) alignment of type s
   _⊢_⇀⦇_,CERTS⦈_     : CertEnv     → CertState  → List DCert  → CertState  → Type
 \end{code}
 \begin{code}[hide]
-  _⊢_⇀⦇_,CERTS⦈_ = ReflexiveTransitiveClosureᵇ' {_⊢_⇀⟦_⟧ᵇ_ = _⊢_⇀⦇_,CERTBASE⦈_} {_⊢_⇀⦇_,CERT⦈_}
+  _⊢_⇀⦇_,CERTS⦈_ = ReflexiveTransitiveClosureᵇ {_⊢_⇀⟦_⟧ᵇ_ = _⊢_⇀⦇_,CERTBASE⦈_} {_⊢_⇀⦇_,CERT⦈_}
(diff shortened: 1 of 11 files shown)


Soupstraw:
`ReflexiveTransitiveClosureᵇ'` just runs the base case as the first step and the one without tick runs the base case as the last step. I realized that this solution creates some new problems, because some conditions in `CERTBASE` actually need to be checked on the initial state.

Soupstraw:
@carlostome I added the issue number above

williamdemeo:
@Soupstraw I'm going to see if I can modify your PR so that it both addresses the conformance test issue (#635) while also preserving the changes I made a while back to resolve issue #545 ([PR 604](https://github.com/IntersectMBO/formal-ledger-specifications/pull/604)).

src/Ledger/Certs.lagda
```diff
@@ -466,15 +466,12 @@ data _⊢_⇀⦇_,CERTBASE⦈_ where
         validVoteDelegs  = voteDelegs ∣^ (  mapˢ (credVoter DRep) (dom dReps)
                                         ∪ fromList (noConfidenceRep ∷ abstainRep ∷ []) )
     in
-    ∙ filter isKeyHash wdrlCreds ⊆ dom voteDelegs
-    ∙ mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ
```
williamdemeo, line 470: I see you moved the check `filter isKeyHash wdrlCreds ⊆ dom voteDelegs` into a premise of the `LEDGER` rule, but where did you put the `mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ` check.  I couldn't find it.  If you think there's a reason to not have it, please add a comment to the change.

williamdemeo, line 470: @Soupstraw I see you moved these two premises to the `LEDGER` rule.  Could you explain why they can't remain here?  (I need them here for the proof of the `CERTBASE-pov` lemma, which is used in the proof of `CERTS-pov`.)

williamdemeo (changes requested):
The PR doesn't type check right now.  (Please see other comments.)

src/Ledger/Certs.lagda
Soupstraw: The problem with the first premise is that in the implementation we check it in the `LEDGER` rule. I guess we could keep it in the same place in the spec and change the implementation instead. The important part is that it's still checked before the certs are applied, like it was before this PR. 

The thing that was wrong was that `validVoteDelegs` was applied *before* the certificates had been applied. This caused DRep delegations to not get removed when the transaction had a `unregdrep` certificate, because the filter function used the old `dReps` map, where the DRep was still registered.

I think the second one could actually stay here as well if we call `CERTBASE` at the start, but I think it makes way more sense to do anything related to withdrawals in the `LEDGER` rule. The fact that we do the withrawals logic in `CERTS` is because that's how it was done in Shelley, before we had a dedicated `CERTS` rule. The withdrawals don't really have anything to do with certificates, so I don't think it makes sense to keep it here, and I've talked with @WhatisRT and @lehins about this and they both agreed. I've already moved that check to `LEDGER` in the implementation, so I think it'd be good to do it here also.

I think the best solution would be to have two base cases in the `CERTS` rule, one that gets applied at the start and then another that gets applied at the end. That way we can do the checks before we apply the certificates, and then update the `voteDelegs` only once we have applied all the certificates.

Soupstraw:
@williamdemeo I thought that if `nix-build -A ledger.hsSrc` succeeds, then all of the files typecheck, but I guess it only needs the conformance spec to typecheck. I'll see if I can figure out the PoV proofs.

williamdemeo:
> @williamdemeo I thought that if `nix-build -A ledger.hsSrc` succeeds, then all of the files typecheck, but I guess it only needs the conformance spec to typecheck. I'll see if I can figure out the PoV proofs.

Sure, but I can also help with the pov proofs.  As I mentioned above, it's possible we will need to adjust the statements of the properties we're proving.

williamdemeo:
This is superseded by #726.

PR #712, opened 2025-03-12
Refactor default.nix

carlostome:
# Description

Refactor default.nix:
- Make derivations depend on ./src and not {formalLedger} to reduce compilation time (when the `Makefile` is also fixed)
- Inline derivations to build html and hsSrc (~blocked on~ #704)
- Remove unused derivations
- Remove dependency on `src/typeCheck.sh`  (Agda has a flag to record building time on a per module basis)

Since `typeCheck.sh` is not used anymore it is also removed.

# Checklist

- [x] Self-reviewed the diff


default.nix
```diff
@@ -102,112 +99,77 @@ rec {
     meta = { };
     buildInputs = deps;
     buildPhase = ''
-      bash typeCheck.sh
+      agda --profile=modules Everything.agda \
+      | tee typecheck.log \
+      | sed '/Checking/!d'
```
carlostome, line 104: I put the piping through `tee` and `sed` so we have some feedback while agda is typechecking but not the time it took per module at the end.

williamdemeo: nice

williamdemeo (approved):
This looks much better to me.  Nice work!

Issue #713, opened 2025-03-12
Add definitions to full ledger spec


williamdemeo:
In cardano-ledger.pdf there are still some terms that are not precisely defined in the prose and some that were not even defined in any previous era of the spec.  To give just two examples, `slot` and `epoch` are used without definition.

These terms are defined in other Cardano documentation (e.g., [here](https://docs.cardano.org/about-cardano/explore-more/time)), but if we want the spec to be fairly self-contained, I propose adding a section to the appendix where we could collect miscellaneous definitions of technical blockchain terminology and cardano-specific nomenclature that are used but not defined elsewhere in the document.

PR #714, opened 2025-03-12
Add missing definitions to appendix of full ledger spec

williamdemeo:
# Description

Addresses issue #713.

# Checklist

- [x] Self-reviewed the diff


williamdemeo:
The idea here is to have a sort of "reference" or "glossary" with definitions of all, or most, technical terms we use to which readers can refer to quickly find out what we mean when we use those terms, without having to search, e.g., the online Cardano docs and then hope we're using the same definition as what they find there.  So, for that purpose, I think using some kind of "Definition" environment, but I'm open to suggestions about how it should look.  If "Definition" seems to formal, perhaps we should just have the term itself in bold, or just use bullet points.

src/latex/references.bib
```diff
@@ -684,3 +684,50 @@ @misc{small-step-semantics
...
+  urldate = {2025-03-12}
+}
+@online{www-ledger-academy-stake-ada,
+  title = {{C}ardano {S}taking: {H}ow {T}o {S}take {ADA}},
+  url = {https://www.ledger.com/academy/cardano-staking-how-to-stake-ada},
+  year = {2024},
+  month = {January},
+  day = {5},
+  urldate = {2025-03-12}
+}
```
williamdemeo, line 732: In retrospect, perhaps we don't want online resources in the References section and should just insert href (links) in the document instead.

williamdemeo: @carlostome wdyt?

latex/definitions.tex
```diff
@@ -0,0 +1,66 @@
...
+\begin{definition}[epoch]
+  An \emph{epoch} is a period of time, containing some number of slots, used to select
+  block-producing nodes.
+\end{definition}
+For example, in Shelley and later eras, an epoch consists of roughly 432,000 slots (or five
+days if we assume a slot length of 1 second).
+
+\begin{definition}[slot number]
+  \emph{Slot number} may refer to a slot's position within the current epoch or it
+  may mean the absolute slot count since the genesis block.  The context should make
```
carlostome, line 40: "genesis block" is used here as a unit of time even though its name doesn't suggest it is. Maybe explain briefly?

latex/definitions.tex
```diff
@@ -0,0 +1,66 @@
...
+  \item slot length = 1 second,
+  \item block coefficient = 0.05,
+  \item slots/epoch = 432,000,
+\end{itemize}
+% \begin{align*}
+%   \text{block time}        &= 20~\text{seconds},\\
+%   \text{slot length}       &= 1~\text{second},\\
+%   \text{block coefficient} &= 0.05,\\
+%   \text{slots/epoch}       &= 432,000,
+% \end{align*}
```
carlostome, line 64:
```suggestion
```
Remove this?

carlostome (approved):
LGTM!

latex/definitions.tex
williamdemeo: Actually, "since the genesis block" seems clear enough to me; I believe it simply means the time elapsed since the genesis block was created.

williamdemeo: ...but you're right, we might as well add a definition for "genesis block."

PR #732, opened 2025-03-26
Add monetary and treasury expansion to `PParams`

HeinrichApfelmus:
# Description

This pull request adds the `monetaryExpansion` and `treasuryCut` protocol parameters to the `PParams` type.

In order to be able to prove that the treasury and reserves are not negative, it is necessary to constrain these parameters need to the unit interval `[0, 1]`. These conditions need to be stored somewhere — I have chosen to make them part of the `PParams` type by refining the parameters to a new type `UnitInterval`. (This is similar to how the previous specification of the Shelley ledger encodes these parameters.)

Context: #706 

# Checklist

- [x] Self-reviewed the diff


WhatisRT:
1. I don't mind the name `UnitInterval`. In this context it'd be highly surprising if it included irrationals, and it's very easy to check if you're unsure.
2. Sorry, I slightly messed up things in my suggestion. I never realized that you're supposed to give a value of type `Term → TC ⊤` in that position rather than a macro. So even though we define `tryConstrs = tryConstrsᵗ` (in a `macro` block), you actually have to write it like this:
```
open import Tactic.Defaults
open import Tactic.Constrs

mkUnitInterval : ∀ (x : ℚ) → {@(tactic (tryConstrsᵗ 10)) _ : does (isInUnitInterval x) ≡ true} → UnitInterval
mkUnitInterval x {evidence} with isInUnitInterval x in eq
```
But all of this doesn't really matter, because I just remembered that you can simply do this instead:
```
mkUnitInterval : ∀ (x : ℚ) → {T (does (isInUnitInterval x))} → UnitInterval
mkUnitInterval x {evidence} with isInUnitInterval x in eq
```
Same effect, no tactics.
3. I think a table in the PParams section would be nice for this. There are also a few more parameters that should be renamed (e.g. `a` and `b`) and would need this treatment. But you don't need to do it in this PR, we can just open an issue for later if you don't feel like it.

I'll wait for the others to have another look if they want, but LGTM

HeinrichApfelmus:
> {T (does (isInUnitInterval x))}

Ooh, very nice, I did not know about this pattern. I have changed the code to use it. Apparently, the trick is that Agda type inference algorithm can infer an implicit argument of type `{T somebool}` by computation of `T somebool`, whereas it cannot infer the very similar argument `{somebool ≡ true}`.

WhatisRT:
Yeah, I'm not entirely sure why that's the case. It could be that the unit type (which is what `T true` reduces to) is just special for the purpose of unification.

WhatisRT:
You should now be able to use revision `4f3ab8562b17a9ccd3fa664072ed9da0451bf029` from stdlib-classes to get `DecEq-Refinement`, just update to this in `default.nix`:
```
    src = fetchFromGitHub {
      repo = "agda-stdlib-classes";
      owner = "omelkonian";
      rev = "4f3ab8562b17a9ccd3fa664072ed9da0451bf029";
      sha256 = "sha256-y1NWXhUY1qcStY7I3OofEOBHbmPwQE13kr13bbhPZBI=";
    };
```
You can then just remove the manually declared instance entirely.

carlostome (approved):
LGTM!

WhatisRT:
> > owner = "omelkonian";
> 
> Why not the one in the agda organization?

Ah, I didn't even notice. That's just historical, it would be good to change that to point to the agda org.

HeinrichApfelmus:
> Ah, I didn't even notice. That's just historical, it would be good to change that to point to the agda org.

Apparently, the repository was transferred at one point, as https://github.com/omelkonian/agda-stdlib-classes redirects to `agda` as owner. I'll changed it for this particular dependency.

PR #739, opened 2025-03-31
Add section "Rewards"

HeinrichApfelmus:
# Description

This pull request adds a section on the calculation of Rewards.

We begin with

* a subsection that briefly motivates the reward system, and
* the definition of the function `maxPool`.

Troubles:

* Unfortunately, I will not be able to add the definition of `createRUpd` in this section, as `EpochState` is not available until later. — This is one instance where more advanced literate programming tools would be beneficial, which allow us to decouple the order of presentation for printing from the order of presentation for compilation.

Comments:

* I managed to hide the proof obligations that we do not divide by `0` from the PDF. However, as type class inference does not work very well with expressions, I had to introduce the seemingly superfluous quantity `1+a0`.
* In order to satisfy the proof obligations, I had to locally change `a0` and `nopt` such that `a0 ≥ 0` and `nopt ≥ 1`. I intend to move these preconditions to `PParams` at a later time.

Context:

#706

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Rewards.lagda
+++ b/src/Ledger/Rewards.lagda
@@ -21,6 +21,55 @@ are calculated and paid out.
 
 \subsection{Rewards Motivation}
 \label{sec:rewards-motivation}
+In order to operate, any blockchain needs to attract parties that are
+willing to spend computational and network resources
+on receiving transactions and producing new blocks.
+These parties, called \defn{block producers},
+are incentivized by monetary \defn{rewards}.
+
+Cardano is a proof of stake (PoS) blockchain:
+Through a random lottery,
+one block producer is selected to produce one particular block.
+The probability for being select depends on their \defn{stake} of Ada,
+that is their attributed amount of Ada relative to the total amount of Ada.
+After successful block production,
+the block producer is eligible for a share of the rewards.
+
+The rewards for block producers come from two sources:
+During an initial period, rewards are paid out from the \defn{reserve},
+which is an initial allocation of Ada created for this very purpose.
+Over time, the reserve is depleted,
+and rewards are sourced from transaction fees.
+
+Rewards are paid out epoch by epoch.
+
+Rewards are collective, but depend on performance:
+After every epoch, a fraction of the available reserve
+and the transaction fees accumulated during that epoch
+are added together. This sum is paid out to the block producers
+proportionally to how many blocks they have created each.
+In order to avoid perverse incentives, block producers
+do not receive individual rewards that depend on the content
+of their blocks.
+
+Not all people can or want to set up and administier a dedicated computer
+that produces blocks. However, these people still own Ada,
+and their stake is relevant for block production.
+Specifically, these people have the option to \defn{delegate} their stake
+to a \defn{stake pool}, which belongs to a block producer.
+The rewards of the block producer are shared proportionally with the
+people that delegate to its stake pool, after fees.
+By design, delegation and ownership are separate
+--- delegation counts towards the stake of the block producer,
+but delegators remain in full control of their Ada,
+stake pools cannot spend delegated Ada.
+
+Stake pools compete for delegators based on fees and performance.
+In order to achieve stable blockchain operation,
+the rewards are chosen such that they incentivize the system to evolve into a
+large, but fixed number of stake pools that attract most of the stake.
+For more details about the design and rationale of the rewards and delegation
+system, see Ref.~\parencite{shelley-delegation-design}.
 
 \subsection{Rewards Distribution Calculation}
 \label{sec:rewards-distribution-calculation}
--- a/src/Ledger/Rewards.lagda
+++ b/src/Ledger/Rewards.lagda
@@ -33,20 +33,21 @@ are calculated and paid out.
 \label{sec:rewards-motivation}
 In order to operate, any blockchain needs to attract parties that are
 willing to spend computational and network resources
-on receiving transactions and producing new blocks.
+on processing transactions and producing new blocks.
 These parties, called \defn{block producers},
 are incentivized by monetary \defn{rewards}.
 
-Cardano is a proof of stake (PoS) blockchain:
-Through a random lottery,
+Cardano is a proof-of-stake (PoS) blockchain:
+through a random lottery,
 one block producer is selected to produce one particular block.
 The probability for being select depends on their \defn{stake} of Ada,
-that is their attributed amount of Ada relative to the total amount of Ada.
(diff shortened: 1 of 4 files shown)


src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+open import Ledger.Certs gs
+\end{code}
+This section defines how rewards for stake pools and their delegators
+are calculated and paid out.
+
+\subsection{Rewards Motivation}
+\label{sec:rewards-motivation}
+In order to operate, any blockchain needs to attract parties that are
+willing to spend computational and network resources
+on receiving transactions and producing new blocks.
```
carlostome, line 36:
```suggestion
on processing transactions and producing new blocks.
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+These parties, called \defn{block producers},
+are incentivized by monetary \defn{rewards}.
+
+Cardano is a proof of stake (PoS) blockchain:
```
carlostome, line 40:
```suggestion
Cardano is a proof-of-stake (PoS) blockchain:
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+Through a random lottery,
```
carlostome, line 41:
```suggestion
through a random lottery,
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+one block producer is selected to produce one particular block.
+The probability for being select depends on their \defn{stake} of Ada,
+that is their attributed amount of Ada relative to the total amount of Ada.
```
carlostome, line 44: I feel the word "attributed" doesn't really explain what is "their stake"

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+After successful block production,
+the block producer is eligible for a share of the rewards.
+
+The rewards for block producers come from two sources:
+During an initial period, rewards are paid out from the \defn{reserve},
```
carlostome, line 49:
```suggestion
during an initial period, rewards are paid out from the \defn{reserve},
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+which is an initial allocation of Ada created for this very purpose.
+Over time, the reserve is depleted,
+and rewards are sourced from transaction fees.
+
+Rewards are paid out epoch by epoch.
+
+Rewards are collective, but depend on performance:
+After every epoch, a fraction of the available reserve
```
carlostome, line 57:
```suggestion
after every epoch, a fraction of the available reserve
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+and the transaction fees accumulated during that epoch
+are added together. This sum is paid out to the block producers
+proportionally to how many blocks they have created each.
+In order to avoid perverse incentives, block producers
+do not receive individual rewards that depend on the content
+of their blocks.
+
+Not all people can or want to set up and administier a dedicated computer
```
carlostome, line 65:
```suggestion
Not all people can or want to set up and administer a dedicated computer
```

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+that produces blocks. However, these people still own Ada,
+and their stake is relevant for block production.
+Specifically, these people have the option to \defn{delegate} their stake
+to a \defn{stake pool}, which belongs to a block producer.
+The rewards of the block producer are shared proportionally with the
+people that delegate to its stake pool, after fees.
```
carlostome, line 71:
```suggestion
The block producer shares the rewards proportionally with the people that delegated their Ada to its stake pool, after fees.
```
I find the sentence hard to understand. I rewrote it in active voice, maybe it helps.

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,172 @@
...
+--- delegation counts towards the stake of the block producer,
+but delegators remain in full control of their Ada,
+stake pools cannot spend delegated Ada.
+
+Stake pools compete for delegators based on fees and performance.
+In order to achieve stable blockchain operation,
+the rewards are chosen such that they incentivize the system to evolve into a
+large, but fixed number of stake pools that attract most of the stake.
+For more details about the design and rationale of the rewards and delegation
+system, see Ref.~\parencite{shelley-delegation-design}.
```
carlostome, line 82:
```suggestion
system, see \textcite{shelley-delegation-design}.
```

carlostome (commented):
Some nitpicks on the text.

src/Ledger/Rewards.lagda
HeinrichApfelmus: Hard to fully explain at this position in the text — I have expanded it to "they (and their delegators) own" with a promise of explaining "delegators" below.

src/Ledger/Rewards.lagda
```diff
@@ -0,0 +1,176 @@
...
+      nonZeroz0 = nonZero-1/n nopt
+
+      nonZero-1+a0 : ℚ.NonZero (1+a0)
+      nonZero-1+a0 = nonZero-1+max0-x (pparams .PParams.a0)
+\end{code}
+\begin{code}
+    rewardℚ =
+        ((fromℕ rewardPot) ÷ 1+a0)
+        * (stake' + pledge' * a0 * (stake' - pledge' * (z0 - stake') ÷ z0) ÷ z0)
+    rewardℕ = posPart (floor rewardℚ)
```
WhatisRT, line 156: Interestingly, this expression can go negative, so `posPart` causes this to not be exactly the same as in the Shelley spec. This isn't a problem however, since if `stake >= pledge` it won't be negative, and this should be true.

But the chain of reasoning is a nightmare: `rewardOnePool` only uses the result of this function if `ostake >= pledge`, but `ostake` and `σ (stake)` aren't formally linked, so you have to go up to `reward`, but there the two relevant arguments are `s` and `Σ s/total` and `total` is again passed as an argument, which is called `circulation` in `createRUpd` which again depends on an argument called `total` but that one is finally hardcoded to `MaxLovelaceSupply` in the only call to it.

I feel like it'd be a good idea to clean that up (not in this PR) if you have the energy to do so.

WhatisRT (approved):
LGTM

Issue #59, opened 2023-03-14
Naming the Prelude modules


WhatisRT:
We have two `Prelude` modules, `Prelude` and `Ledger.Prelude`. The former publicly re-exports general purpose definitions and the latter is the same but includes definitions used in `Ledger.X` modules.

The purpose of this issue is to discuss whether they should be renamed and to what.

williamdemeo:
Superseded by #302.

PR #740, opened 2025-04-01
Cleanup agda code and organize preludes

carlostome:
# Description

Adds a new top-level folder `lib-exts` with two purposes:
1. Host extension files to libraries (e.g. the Agda standard library).
2. Host glue code between the library and the formal ledger Agda development. 

`lib-exts` is organized as a top-level directory for each library. 
```
lib-exts/
  stdlib
  stdlib-meta
  stdlib-classes
  iog-prelude
```

The name of Agda modules in these directories are prepended with the name of the library, e.g. `stdlib.Algebra.Literals`. This convention enables a smooth solution to both (1) and (2). We can reuse the name of the module as exported by the library (but prepended with the directory) both for extensions (see e.g. `stdlib.Algebra.Literals`) and glueing (see e.g. `iog-prelude.Prelude.InferenceRules`).

The idea is that modules in these folders cannot depend on ledger, but can depend on extensions (or glueing code) from the other libraries. This can be checked by commenting out `src` from `formal-ledger.agda-lib` and typechecking the modules under `lib-exts`.

In addition, this PR adds a dependency on `iog-prelude` as a first step to organize the Prelude situation in Ledger (see #59)

# Checklist

- [x] Self-reviewed the diff


carlostome:
I've split this PR into #742

lib-exts/stdlib-meta/Reflection.agda
```diff
@@ -1,14 +1,27 @@
-{-# OPTIONS --safe #-}
-module Reflection.Ext where
+{-# OPTIONS --safe --without-K #-}
```
williamdemeo, line 1: Is it possible to use `--cubical-compatible` here?  Going forward I think should try to use `--cubical-compatible` instead of `--without-K` whenever possible.

lib-exts/stdlib/Data/List/Subpermutations/Properties.agda
```diff
@@ -1,15 +1,15 @@
-{-# OPTIONS --safe #-}
+{-# OPTIONS --safe --without-K #-}
```
williamdemeo, line 1:
```suggestion
{-# OPTIONS --safe --cubical-compatible #-}
```
(if it type-checks)

williamdemeo (approved):
I think I understand the rationale here and I like it... but we can discuss during our meeting just to be sure.  It's good to have these libraries factored out, and then, as a side effect, we know each time they're imported exactly where they're coming from just by looking at the names.  Nice work!  LGTM!

PR #742, opened 2025-04-02
Cleanup agda

carlostome:
# Description

This PR cleans some Agda source code.

# Checklist

- [x] Self-reviewed the diff


src/Ledger/Conway/Foreign/HSLedger/Core.agda
```diff
@@ -80,13 +80,8 @@ module Implementation where
...
   ExUnits      = ℕ × ℕ
-  ExUnit-CommutativeMonoid = CommutativeMonoid 0ℓ 0ℓ ExUnits ∋ (Conversion.fromBundle record
-    { Carrier = ExUnits
-    ; _≈_ = _≈ᵖ_
-    ; _∙_ = _∙ᵖ_
-    ; ε = zero , zero
-    ; isCommutativeMonoid = pairOpRespectsComm +-0-isCommutativeMonoid
-    }) where open import Algebra.PairOp ℕ zero _≡_ _+_
+  ExUnit-CommutativeMonoid =
+    Conversion.fromBundle (commutativeMonoid +-0-commutativeMonoid +-0-commutativeMonoid)
```
williamdemeo, line 84: I'm not opposed to this change in principle, as I agree we should use stuff from stdlib or stdlib-classes when it simplifies... but I could understand what we had before at a glance, whereas using `Conversion.fromBundle` requires either familiarity with these conversions already or jumping to the `CommutativeModule.Core` module to see what's happening here... it's not terribly inconvenient, but does add an extra layer of abstraction.

src/Ledger/Gov.lagda
```diff
@@ -329,15 +330,15 @@ any?-connecting-subperm : ∀ {u} {v} → ∀ L → Dec (Any(λ l → Unique l 
...
 
 ∃?-connecting-subset : ∀ {u} {v} → ∀ L → Dec (∃[ l ] l ⊆ˡ L × Unique l × l connects u to v)
-∃?-connecting-subset L = from (map′⇔ ∃uniqueSubset⇔∃uniqueSubperm) (∃?-connecting-subperm L)
+∃?-connecting-subset L = map-Dec (sym-Equiv ∃uniqueSubset⇔∃uniqueSubperm) (∃?-connecting-subperm L)
 
 enactable? : ∀ eState aidPairs aidNew×st → Dec (enactable eState aidPairs aidNew×st)
 enactable? eState aidPairs (aidNew , as) with getHashES eState (GovActionState.action as .gaType)
 ... | nothing = yes tt
-... | just aidOld = from (map′⇔ ∃-sublist-⇔) (∃?-connecting-subset aidPairs)
+... | just aidOld = map-Dec (sym-Equiv ∃-sublist-⇔) (∃?-connecting-subset aidPairs) -- from (map′⇔ ∃-sublist-⇔) (∃?-connecting-subset aidPairs)
```
williamdemeo, line 341:
```suggestion
... | just aidOld = map-Dec (sym-Equiv ∃-sublist-⇔) (∃?-connecting-subset aidPairs)
```

src/Prelude.agda
```diff
@@ -81,10 +79,34 @@ import Data.Rational.Literals as ℚ
...
 _~_ = Connected _≡_
 
 instance
   ~? : {A : Type} {x y : Maybe A} → ⦃ DecEq A ⦄ → (x ~ y) ⁇
   ~? {A} {x} {y} ⦃ deqEq ⦄ = ⁇ (connected? (DecEq._≟_ deqEq) x y)
+
+-- Positive and negative part of integers
+open import Data.Integer using (sign; ∣_∣; _⊖_)
+open import Data.Integer.Properties using ([1+m]⊖[1+n]≡m⊖n)
+open import Data.Sign using (Sign)
```
williamdemeo, line 96: I prefer to have all imports at the top of the file.  It's more consistent with the rest of the code base, though there are exceptions to this rule of thumb.

williamdemeo (changes requested):
Nice work.  I made some minor comments, which you can disregard if you wish.  Though you probably do want to commit the suggestion that removes a comment of old code.  I'll mark this "Request changes" just in case you have it set to auto-merge (which I believe would then merge this without committing the removal of the comment).

src/Ledger/Conway/Foreign/HSLedger/Core.agda
williamdemeo: Ah, I see it's not set to auto-merge, so I'll approve it.

williamdemeo (approved):
After removing the comment on line 341 of `Ledger.Gov`, LGTM!

src/Prelude.agda
carlostome: I decided to leave at the top the modules that are reexported, while leave those only needed to define things close to those things.

Issue #722, opened 2025-03-21
Document proved and to-be-proved properties


williamdemeo:
Some of the to-be-proved properties are given as types at the bottom of `Ledger/Properties.agda`.

## To Do

- [ ] Add comments to the `Ledger/Properties.agda` file describing the properties in English prose.
- [ ] Start a list or table of the properties we've proved and those we'd like to prove; this could be done here (in this issue) and/or in a section of the ledger spec pdf (perhaps in an appendix or a table in the front matter).

Issue #23, opened 2022-10-28
Projections for the principal fields of records


WhatisRT:
As @Soupstraw pointed out, it's sometimes confusing to read something like `proj₁`, since it's not always clear if we're projecting out from something we think of as a pair, or if we're projecting away a proof. Related to this is the issue that we sometimes have records that have a 'principal' field, like the `Carrier` field in many algebraic structures for example. It would be nice to have a consistent way of projecting away less relevant parts of a record, for example with the following typeclass:

``` agda
record HasDowncast {a b} (A : Set a) (B : Set b) : Set (a ⊔ b) where
  field _↓ : A → B
```

PR #743, opened 2025-04-03
Document proved and to-be-proved properties

williamdemeo:
# Description

Addresses issue #722.

~**DO NOT MERGE**~
In order to keep this PR focused on properties, I'll make a new PR for the type class accessors (see notes about this below) and hopefully we can merge that before this one.

## type classes for easier access to fields

The type classes I add in this PR address issue #23, but I needed them to make the explanations of properties simpler.  I will also make a separate PR with just the type classes and hopefully we can get that merged relatively quickly, and before merging this PR.

The type classes for accessors allow us to, for example, get the `GovState` of a `ChainState` `cs` simply as `GovStateOf cs` instead of `cs .newEpochState .epochState .ls .govSt` or, worse, if the records types were not already opened, `LState.govSt (EpochState.ls (NewEpochState.epochState (ChainState.newEpochState cs)))`.

Type classes are natural here since our hierarchy of structures mean we will want to access the same field from multiple levels; for example, besides `GovStateOf cs` (for `cs : ChainState`), we can also do `GovStateOf nes` for an `nes : NewEpochState` (instead of `nes .epochState .ls .govSt`).

# Checklist

- [ ] Self-reviewed the diff


src/Ledger/GovernanceActions/Properties/ProposalChangePP.lagda
```diff
@@ -0,0 +1,56 @@
...
+\begin{property}[%
+  \LedgerMod{GovernanceActions/Properties/ProposalChangePP.lagda}{\AgdaModule{ProposalChangePP}}:
+  PParam updates have non-empty groups;
+  \textbf{to prove}%
+  ]\
+
+  \begin{itemize}
+    \item \textit{Informally}.
+      Suppose \ab{p} is a goveranance proposal (\GovProposal{}) whose \AgdaField{action}
+      (\GovAction{}) has \GovActionType{} \ChangePParams{} and \GovActionData{} \ab{up}.
```
carlostome, line 27:
```suggestion
      Suppose \ab{p} is a governance proposal (\GovProposal{}) whose \AgdaField{action} :
      \GovAction{} has \GovActionType{} \ChangePParams{} and \GovActionData{} \ab{up}.
```

carlostome, line 27:
```suggestion
      Suppose \ab{p} is a goveranance proposal (\GovProposal{}) whose \AgdaField{action}
      (\GovAction{}) has \GovActionType{} \ChangePParams{} and \GovActionData{} \ab{pu}.
```
params update?

src/Ledger/Chain/Properties/CHAINepochStep.lagda
```diff
@@ -0,0 +1,50 @@
...
+  \begin{itemize}
+    \item \textit{Informally}.
+      Let
+
+      \(\begin{array}{rcl}
+        \ab{s}&=&\text{a \AgdaRecord{ChainState}},\\
+        \ab{nes}&=&\text{the \AgdaField{newEpochState} of \ab{s}},\\
+        \ab{es}&=&\text{the \AgdaField{epochState} of \ab{nes} (= the \AgdaField{epochState} of \ab{s})},\\
+        \ab{le}&=&\text{the \AgdaField{lastEpoch} of \ab{nes}}.
+      \end{array}\)
```
carlostome, line 30: I like it more in the form of text as in the other properties than a list of bullet points.

src/Everything.agda
```diff
@@ -30,3 +30,19 @@ import Foreign.Convertible.DerivingTest
...
 import ScriptVerification.Lib
+
+-- ** Properties
+import Ledger.Utxo.Properties.MinSpend
+import Ledger.Utxo.Properties.UTXOpov
+import Ledger.Certs.Properties.CERTSpov
+import Ledger.Ledger.Properties.LEDGERpov
+import Ledger.Ledger.Properties.LEDGERgovDepsMatch
+import Ledger.Epoch.Properties.EPOCHgovDepsMatch
+import Ledger.Chain.Properties.CHAINgovDepsMatch
```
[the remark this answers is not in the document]
williamdemeo: I don't understand this comment  (but the missing i is a good catch!)

src/Ledger/Utxo/Properties/MinSpend.lagda
```diff
@@ -0,0 +1,289 @@
...
+      \ab{utxoSt'}~:~\UTxOState{} be two UTxO states such that
+      \ab{utxoSt}~\AgdaDatatype{⇀⦇}~\ab{tx}~\AgdaDatatype{,UTXO⦈}~\ab{utxoSt'}.
+      Then the coin consumed by \ab{tx} is at least the sum of the governance action deposits
+      of the proposals in \ab{tx}.
+
+
+    \item \textit{Formally}.
+\begin{AgdaMultiCode}
+\begin{code}
+  gmsc :  { Γ        : UTxOEnv    }
```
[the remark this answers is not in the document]
williamdemeo: "general minimum spending condition" was how we had been referring to it.  I'm happy to change the name `gmsc` though.

WhatisRT (commented):
I think it would be a good idea to take a step back and think about who is going to read this. I can imagine a few customers:
1. Someone who just wants to know that we have some good properties mostly cares about the high level. For that person, all of preservation of value except for `LEDGER` is complete noise.
2. Someone who is interested in the properties more deeply and knows Agda. That person would realistically look at the code anyway, and would maybe appreciate the links to the code and the high-level overview.
3. Someone who is interested in the properties more deeply and doesn't know Agda. For that person, the formal statements don't help them understand the theorem, and they are probably disappointed that the proof is completely inaccessible to them.

My worry here is that most of this work is really only for internal consumption - but we can all read the Agda code anyway and probably would be fine doing just that.

If we compare this e.g. to the Preservation of Value section of the Shelley spec (15.7), customers 1 and 3 would probably be happy. The main statement is introduced prominently on the first page, and all non-trivial statements have informal proofs. If it looked like that, together with links to the relevant Agda code, I'm sure customer 2 would be happy as well.

williamdemeo:
@WhatisRT Unfortunately, we didn't have time to discuss this PR before you looked at it, so I didn't get a change to make the goal of this work clear to you.  We want an organized, well-written and reasonably comprehensible (to Agda and non-Agda readers alike) list of the things we've proved about the Ledger, rather than having these properties strewn about the Agda code base.  The main customer I have in mind is James, who wants such a list, but I hope the properties section will be useful to other as well, and this is why I worked hard to make both the informal and formal statements of the properties as clean as possible.  This required a lot of revisions and additions to the Agda code (avoiding special syntax/symbols, making intuitively-named accessor functions for deeply embedded data, etc) which is why the PR is taking so long to finish.

latex/cardano-ledger.tex
```diff
@@ -31,13 +31,48 @@
...
-\includeAgda{Ledger/Rewards}
 \includeAgda{Ledger/Epoch}
 \includeAgda{Ledger/Chain}
 
 \section{Properties}
-\includeAgda{Ledger/Utxo/Properties}
+This section presents the properties of the ledger that we have formally proved in
+Agda or plan to do so in the near future.  We indicate in which Agda module each
+property is formally stated and (possibly) proved. A ``Claim'' is a property that is
+not yet proved, while a ``Theorem'' is one for which we have a formal proof.
```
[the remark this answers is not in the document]
williamdemeo: Using lemmas in addition to theorems seems reasonable for future consideration, but the idea here is to compile a list of the properties we've proved mainly for reference.  To get this PR over the finish line, I don't want to try to judge which are the important properties and highlight those, or hide some properties that a hypothetical reader might perceive as "complete noise."

williamdemeo:
I've added type classes for accessors so we can, for example, get the `GovState` of a `ChainState` `cs` simply as 

`getGovState cs` 

instead of

`cs .newEpochState .epochState .ls .govSt`

or, worse, if the records types were not already opened, 
 
`LState.govSt (EpochState.ls (NewEpochState.epochState (ChainState.newEpochState cs)))`.

The accessors should be type classes so we can do the same for other structures; e.g., we can now do

`getGovState nes`

for an inhabitant of `NewEpochState`, instead of 

`nes .epochState .ls .govSt`.

I believe this is a solution to issue #23, but since I needed this feature to make the explanations of properties simpler, I decided to include them in this PR.

latex/cardano-ledger.tex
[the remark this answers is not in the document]
williamdemeo: After reconsidering this, I realized it wasn't too hard to tell which "theorems" would naturally be called "lemmas," so I went ahead and incorporated your suggested change.

src/Ledger/Utxo/Properties/PoV.lagda
```diff
@@ -0,0 +1,67 @@
...
+    the first argument when the second argument is true, and returns zero when the second
+    argument is false. That is,
+        \[
+          \mbox{\AgdaFunction{φ}} (n , b) = \left\{
+            \begin{array}{ll}
+              n & \text{if } b = \mbox{true}, \\
+              0 & \text{if } b = \mbox{false}.
+            \end{array}\right.
+        \]     
+    \item \textit{Formally}.
```
carlostome, line 48: I'd get rid altogether of the phi function; it only adds noise and unnecessary indirection.

src/Ledger/Certs/Properties/PoV.lagda
```diff
@@ -0,0 +1,84 @@
...
+  \begin{itemize}
+    \item \textit{Informally}.
+      Let \ab{l} be a list of \DCert{}s, and let \ab{s₁}, \ab{sₙ} be \CertState{}s
+      such that \ab{s₁}~\AgdaDatatype{⇀⦇}~\ab{l}~\AgdaDatatype{,CERTS⦈}~\ab{sₙ}.
+      Then, the value of \ab{s₁} is equal to the value of \ab{sₙ} plus
+      the value of the withdrawals in \ab{Γ}.
+    \item \textit{Formally}.
+\begin{code}
+    CERTS-pov :  {Γ : CertEnv} {s₁ sₙ  : CertState}
+      → ∀[ a ∈ dom (CertEnv.wdrls Γ) ] NetworkIdOf a ≡ NetworkId
```
carlostome, line 68: This assumption is not mentioned in the prose. Why is it needed?

src/Ledger/Certs/Properties/PoV.lagda
```diff
@@ -0,0 +1,84 @@
...
+      →  Γ ⊢ s₁ ⇀⦇ l ,CERTS⦈ sₙ 
+      → getCoin s₁ ≡ getCoin sₙ + getCoin (CertEnv.wdrls Γ)
```
carlostome, line 70:
```suggestion
      → getCoin s₁ ≡ getCoin sₙ + getCoin (wdrlsOf Γ)
```
where `wdrlsOf` is a the method of a class to extract withdrawals?

src/Ledger/Utxo/Properties/PoV.lagda
```diff
@@ -0,0 +1,67 @@
...
+\begin{code}
+UTXOpov : {Γ : UTxOEnv} {tx : Tx} {s s' : UTxOState}
+  → txidOf tx ∉ mapˢ proj₁ (dom (UTxOOf s))
+  → Γ ⊢ s ⇀⦇ tx ,UTXO⦈ s'
+  → getCoin s + φ(getCoin (txwdrlsOf tx) , tx .isValid) ≡ getCoin s'
```
carlostome, line 53:
```suggestion
  → getCoin s + φ(getCoin (wdrlsOf tx) , tx .isValid) ≡ getCoin s'
```
where `wdrlsOf` is a class for getting withdrawals out of a tx?

src/Ledger/Certs/Properties/PoVLemmas.lagda
```diff
@@ -0,0 +1,245 @@
...
+      \ab{s}~\AgdaDatatype{⇀⦇}~\_~\AgdaDatatype{,CERTBASE⦈}~\ab{s'}.
+      Then, the value of \ab{s} is equal to the value of \ab{s'} plus the value of
+      the withdrawals in \ab{Γ}.  In other terms,
+      \\[4pt]
+      \AgdaField{getCoin}~\ab{s} $≡$ \AgdaField{getCoin}~\ab{s'}
+       + \AgdaField{getCoin}~(\ab{Γ} .\AgdaField{wdrls} ).
+    \item \textit{Formally}.
+\begin{code}
+    CERTBASE-pov : {Γ : CertEnv} {s s' : CertState}
+      → ∀[ a ∈ dom (CertEnv.wdrls Γ) ] NetworkIdOf a ≡ NetworkId
```
carlostome, line 157: Why this assumption?

src/Ledger/Certs/Properties/PoVLemmas.lagda
```diff
@@ -0,0 +1,245 @@
...
+    \item \textit{Proof}. See the
+      \LedgerMod{\CertsPoVL.lagda}{\AgdaModule{\CertsPoVL{}}}
+      module in the \href{\repourl}{formal ledger repository}.
+\begin{code}[hide]
+    -- Proof.
+    sts-pov (BS-base Id-nop) = refl
+    sts-pov (BS-ind x xs) = trans (CERT-pov x) (sts-pov xs)
+\end{code}
+  \end{itemize}
+\end{lemma}
```
carlostome, line 243: How is this lemma different from the theorem in Ledger.Certs.Properties.PoV? (Since CERTS is defined as the "reflexive-transitive closure" of CERTBASE)

latex/cardano-ledger.tex
```diff
@@ -36,20 +36,65 @@
...
+
+\subsection{Preservation of Value}
+\inputAgda{Ledger/Ledger/Properties/PoV}
+\inputAgda{Ledger/Utxo/Properties/PoV}
+\inputAgda{Ledger/Certs/Properties/PoV}
+\inputAgda{Ledger/Certs/Properties/PoVLemmas}
+
+\subsection{Invariance Properties}
+\Cref{thm:ChainGovDepsMatch} and \cref{lem:LedgerGovDepsMatch,lem:EpochGovDepsMatch} assert that
```
carlostome, line 51:
```suggestion
\Cref{thm:ChainGovDepsMatch,lem:LedgerGovDepsMatch,lem:EpochGovDepsMatch} assert that
```

latex/cardano-ledger.tex
```diff
@@ -36,20 +36,65 @@
...
+a certain relation is an invariant of the \CHAIN{}, \LEDGER{}, and \EPOCH{} rules, respectively.
```
carlostome, line 52:
```suggestion
a certain predicate is an invariant of the \CHAIN{}, \LEDGER{}, and \EPOCH{} rules, respectively.
```
Relation or predicate on states? Here is called relation but a bit below predicate.

latex/cardano-ledger.tex
```diff
@@ -36,20 +36,65 @@
...
+``the \AgdaFunction{govDepsMatch} relation is an invariant of the \LEDGER{} rule,''
+means the following:  if \govDepsMatch{}~\ab{s} and
+\ab{s}~\AgdaDatatype{⇀⦇}~\ab{tx}~\AgdaDatatype{,LEDGER⦈}~\ab{s'}, then
+\govDepsMatch{}~\ab{s'}.
+
+\inputAgda{Ledger/Chain/Properties/GovDepsMatch}
+\inputAgda{Ledger/Ledger/Properties/GovDepsMatch}
+\inputAgda{Ledger/Epoch/Properties/GovDepsMatch}
+\inputAgda{Ledger/Chain/Properties/CredDepsEqualDomRwds}
+\inputAgda{Ledger/Chain/Properties/PParamsWellFormed}
```
carlostome, line 71: I feel this bunch should be bundled together in a subsubsection called something like "Goverance action deposits match".
This will also help us be organized when we add other kinds of invariance properties.

latex/cardano-ledger.tex
```diff
@@ -36,20 +36,65 @@
...
+property is formally stated and (possibly) proved. A ``Claim'' is a property that is
+not yet proved, while a ``Theorem'' is one for which we have a formal proof.
+
+\subsection{Preservation of Value}
+\inputAgda{Ledger/Ledger/Properties/PoV}
+\inputAgda{Ledger/Utxo/Properties/PoV}
+\inputAgda{Ledger/Certs/Properties/PoV}
+\inputAgda{Ledger/Certs/Properties/PoVLemmas}
+
+\subsection{Invariance Properties}
```
carlostome, line 50: Maybe this subsection is missing a `govDepsMatch`-independent introduction to what is meant by invariance properties (given a predicate P on states and a rule like Ledger then blablabla), and this introduction corresponds to a subsubsection.

src/Ledger/Certs/Properties.agda
```diff
@@ -312,5 +312,3 @@ module _  {Γ : CertEnv}
                                                    (cong  (_+ getCoin (CertEnv.wdrls Γ))
                                                           (trans (CERT-pov x) (sts-pov sts)))
 
--- TODO: Prove the following property.
--- range vDelegs ⊆ map (credVoter DRep) (dom DReps)
```
carlostome, line 316: Why remove this comment? Is this proved somewhere already or is there a ticket for this (so we don't forget)

src/Ledger/Chain/Properties/GovDepsMatch.lagda
```diff
@@ -0,0 +1,104 @@
...
+      set to the \AgdaField{epochState} of \ab{cs} with the exception of the
+      \LState{} field, which is set to that of \ab{nes}.
+      \\[4pt]
+      Let \ab{utxoSt} and \ab{utxoSt'} be the respective \UTxOState{}s of the ledger
+      states of \ab{cs} and \AgdaFunction{cs'}, respectively, and let \ab{govSt} and \ab{govSt'}
+      be their respective \GovState{}s.
+      \\[4pt]
+      Assume the following conditions hold:
+      \begin{itemize}
+      \item the hypothesis about \AgdaFunction{removed'} needed for and described in \cref{lem:EpochGovDepsMatch},\\[4pt]
```
carlostome, line 68: I'm not a fan that this is a forward reference, instead of there being a backwards one (since this is the theorem maybe it'd make more sense to explain it here)

src/Ledger/Chain/Properties/GovDepsMatch.lagda
```diff
@@ -0,0 +1,104 @@
...
+      Under these conditions, if the governance action deposits of \ab{utxoSt}
+      equal those of \ab{govSt}, then the same holds for \ab{utxoSt'} and \ab{govSt'}.
+      In other terms,
+      \AgdaFunction{govDepsMatch}~\AgdaFunction{csLState} implies \AgdaFunction{govDepsMatch}~\AgdaFunction{nesState}.
+    \item \textit{Formally}.
+\begin{code}
+  CHAIN-govDepsMatch :
+    map (GovActionDeposit ∘ proj₁) removed' ⊆ map proj₁ (DepositsOf cs ˢ)
+    →  totalRefScriptsSize csLState ts ≤ maxRefScriptSizePerBlock
+    →  tt ⊢ cs ⇀⦇ b ,CHAIN⦈ cs'
```
carlostome, line 84:
```suggestion
    →  _ ⊢ cs ⇀⦇ b ,CHAIN⦈ cs'
```

src/Ledger/Chain/Properties/GovDepsMatch.lagda
```diff
@@ -0,0 +1,104 @@
...
+      \end{itemize}
+      Under these conditions, if the governance action deposits of \ab{utxoSt}
+      equal those of \ab{govSt}, then the same holds for \ab{utxoSt'} and \ab{govSt'}.
+      In other terms,
+      \AgdaFunction{govDepsMatch}~\AgdaFunction{csLState} implies \AgdaFunction{govDepsMatch}~\AgdaFunction{nesState}.
+    \item \textit{Formally}.
+\begin{code}
+  CHAIN-govDepsMatch :
+    map (GovActionDeposit ∘ proj₁) removed' ⊆ map proj₁ (DepositsOf cs ˢ)
+    →  totalRefScriptsSize csLState ts ≤ maxRefScriptSizePerBlock
```
carlostome, line 83: Maybe get rid of the projections?

src/Ledger/Epoch/Properties/GovDepsMatch.lagda
```diff
@@ -0,0 +1,184 @@
...
+      \item the governance actions in the \AgdaField{removed} field of the ratify
+    state of \ab{eps}, and
+      \item the orphaned governance actions in the \GovState{} of \ab{eps}.
+    \end{itemize}
+    Let $\mathcal{G}$ be the set
+    $\{\mbox{\GovActionDeposit{}~\ab{id}} : \mbox{\ab{id}} ∈ \mbox{proj}₁~\mbox{\AgdaFunction{removed'}}\}$.
+    The special hypothesis under which \govDepsMatch{} is an invariant of the
+    \EPOCH{} rule is the following: $\mathcal{G}$ is a subset of the set of deposits
+    of (the governance state of) \ab{eps}.
+    \item \textit{Formally}.
```
carlostome, line 85: Maybe get rid of the projections?

latex/cardano-ledger.tex
carlostome, line 71: These two should go in a different subsubsection

src/Ledger/Chain/Properties/EpochStep.lagda
```diff
@@ -0,0 +1,47 @@
...
+      Let \ab{cs} and \ab{cs'} be \ChainState{}s and \ab{b} a \Block{}.
+      If \ab{cs}~\AgdaDatatype{⇀⦇}~\ab{b}~\AgdaDatatype{,CHAIN⦈}~\ab{cs'}
+      and if the enact states of \ab{cs} and \ab{cs'} differ, then
+      the epoch of the slot of \ab{b} is the successor of the last epoch of \ab{cs}.
+    \item \textit{Formally}.
+\begin{AgdaMultiCode}
+\begin{code}
+enact-change⇒newEpoch : (b : Block) {cs cs'  : ChainState} 
+  → _ ⊢ cs ⇀⦇ b ,CHAIN⦈ cs' → EnactStateOf cs ≢ EnactStateOf cs'
+  → Type
```
carlostome, line 38: I think here we should be careful with what it means for two Enact states to differ. Since `EnactState` contains sets and maps I don't think propositional equality is the right choice to use here.

src/Ledger/Epoch/Properties/NoPropSameDReps.lagda
```diff
@@ -0,0 +1,38 @@
...
+\newcommand{\EpochPropConstDReps}{Epoch/Properties/NoPropSameDReps}
+
+
+\begin{claim}[%
+  \LedgerMod{\EpochPropConstDReps.lagda}{\AgdaModule{\EpochPropConstDReps{}}}:
+  DReps unchanged if no gov proposals%
+  ]\
+
+  \begin{itemize}
+    \item \textit{Informally}. TODO
```
carlostome, line 27: TODO

carlostome (commented):
I left some comments here and there. Overall the PR is looking good!

src/Ledger/Utxo/Properties/PoV.lagda
williamdemeo: Do you mean get rid of it in the code as well?  ...using an `if...then...else`?

williamdemeo: I seem to recall, previously, we had two separate properties, one was proved under the assumption that the transaction is valid, and the other.. invalid.  I think we decided it would be better to be more concise and use a function like phi.  (Perhaps it would have been better to name it chi instead of phi, but I don't think that really addresses your main concern.)

williamdemeo: I've decided to remove the formal definition of the phi function and rename it chi (since it's sort of a characteristic function).

src/Ledger/Certs/Properties/PoV.lagda
williamdemeo: Not sure. 🤷🏼

src/Ledger/Certs/Properties/PoVLemmas.lagda
williamdemeo: Actually, `CERTS` is the "refl-trans closure" of `CERT` (with `CERTSBASE` as the "base step" ...though it's not really a base step).  So, the proof of `CERTS-pov` uses both the `sts-pov` and the `CERTBASE-pov` lemmas.  If you think we should inline `sts-pov` and not present it as a separate lemma, that's probably a good idea.  However, I don't want to refactor proofs in this PR.

latex/cardano-ledger.tex
williamdemeo: I agree, although I think an extra subsubsection for a single introductory sentence is overkill.

src/Ledger/Certs/Properties.agda
williamdemeo: I'll make another property module for this.

src/Ledger/Chain/Properties/GovDepsMatch.lagda
williamdemeo: Let's save this for a later PR.

src/Ledger/Epoch/Properties/GovDepsMatch.lagda
williamdemeo: Actually, the projections on the right can be replaced by using `dom`, so I'll do that, but I'd like to leave attempts to remove the projection on the left for another PR.

latex/cardano-ledger.tex
williamdemeo: I think we can avoid making a subsubsection for these (what would be the title, anyway?) by simply putting them before the "govDepsMatch" subsubsection.

src/Ledger/Chain/Properties/EpochStep.lagda
williamdemeo: Let's take this up in a later issue/PR.

Issue #679, opened 2025-02-05
Complete the implementation of `stakeDistr`


carlostome:
As per @WhatisRT comment:

Ah, this function was only partially implemented. A correct implementation would need `PState`, so I don't think it's a good idea to remove it.

If you want to finish the implementation, look at Fig. 37 of the Shelley spec. Note that `addrPtr` has now been deprecated and removed, so that part is intentionally missing. This is also one of the few places where the spec really leans into set theory, so the inverse here turns a function into a relation and performs some operations on relations. I think `stakeRelation` is correct in this code.

_Originally posted by @WhatisRT in https://github.com/IntersectMBO/formal-ledger-specifications/pull/670#pullrequestreview-2592665435_

PR #752, opened 2025-04-10
Add `createRUpd` (PR)

HeinrichApfelmus:
# Description

This pull request adds the function `createRUpd` to the specification.

Comments:

* I have decided to stay very close to the notation from the original Shelley specification in order to make it straightforward to check that the two specifications agree.
* 6 lines in the `where` clause of `createRUpd` are devoted to extracting fields from records. This is somewhat lengthy, but I am under the impression that `where` clauses in Agda do not support pattern matching. It would be beneficial to join at least 3 lines into a single line `open Snapshot pstakego;`, but I am not quite sure what the policy on record modules is — `open` is used a couple of times in the specification, but it is never explained in Section 1 or 2 of the PDF.
* I have added `poolParameters` to the `SnapShot` type and the `SNAP` rule, as they are needed by the `createRUpd` function.
* I have also added the "Preservation of funds" diagram from the original Shelley specification for good measure.

Context:

#706, #679

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -8,13 +8,19 @@
 open import Data.Nat.Properties using (+-0-monoid; +-0-commutativeMonoid)
 open import Data.Integer using () renaming (+_ to pos)
 open import Data.Nat.GeneralisedArithmetic using (iterate)
+open import Data.Rational using (ℚ; floor; _*_; _÷_; _/_)
+open import Data.Rational.Literals using (number; fromℤ)
+import      Data.Rational as ℚ renaming (_⊓_ to min)
 
 open import Agda.Builtin.FromNat
 
-open import Ledger.Prelude hiding (iterate)
+open import Ledger.Prelude hiding (iterate; _/_; _*_)
 open Filter using (filter)
 open import Ledger.Abstract
 open import Ledger.Transaction
+open import Ledger.Types.Numeric.UnitInterval
+
+open Number number renaming (fromNat to fromℕ)
 
 module Ledger.Epoch
   (txs : _) (open TransactionStructure txs)
@@ -25,16 +31,9 @@ open import Ledger.Certs govStructure
 open import Ledger.Enact govStructure
 open import Ledger.Gov txs
 open import Ledger.Ledger txs abs
-open import Ledger.Prelude hiding (iterate)
 open import Ledger.Ratify txs
 open import Ledger.Rewards govStructure
 open import Ledger.Utxo txs abs
-
-open import Agda.Builtin.FromNat
-
-open import Data.Integer using () renaming (+_ to pos)
-open import Data.Nat.GeneralisedArithmetic using (iterate)
-open import Data.Nat.Properties using (+-0-monoid; +-0-commutativeMonoid)
 
 \end{code}
 \begin{figure*}[ht]
@@ -146,9 +145,86 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
 getStakeCred : TxOut → Maybe Credential
 getStakeCred (a , _ , _ , _) = stakeCred a
 
-open RwdAddr using (stake)
 open GovActionState using (returnAddr)
 \end{code}
+\begin{NoConway}
+\Cref{fig:functions:createRUpd} defines
+the function \AgdaFunction{createRUpd}
+which creates a \AgdaRecord{RewardUpdate},
+i.e.\@ the net flow of Ada due to paying out rewards
+after an epoch.
+Relevant quantities are:
+\begin{itemize}
+  \item \AgdaFunction{prevPp}: Previous protocol parameters,
+    which correspond to the parameters during the epoch for which we are creating rewards.
+  \item \AgdaFunction{ActiveSlotCoeff}:
+    Global constant, equal to the probability
+    that a party holding all the stake will be selected to be a leader for given slot.
+    Equals $1/20$ during the Shelley era on the Cardano Mainnet.
+  \item \AgdaFunction{Δr₁}: Ada taken out of the reserves for paying rewards,
+    as determined by the \AgdaField{monetaryExpansion} protocol parameter.
+  \item \AgdaFunction{rewardPot}: Total amount of coin available for rewards this epoch, as
+described in \textcite[\sectionname~6.4]{shelley-delegation-design}.
+  \item \AgdaFunction{feeSS}: The fee pot which, together with the reserves,
+    funds the \AgdaFunction{rewardPot}.
+    We use the fee pot that accumulated during the epoch
+    for which we now compute block production rewards.
+    Note that fees are not explicitly removed from any account:
+    the fees come from transactions paying them and are accounted for whenever transactions are processed.
+  \item \AgdaFunction{Δt₁}: The proportion of the reward pot that will move to the treasury,
+    as determined by the \AgdaField{treasuryCut} protocol parameter.
+    The remaining pot is called the \AgdaFunction{R},
+    just as in \textcite[\sectionname~6.5]{shelley-delegation-design}.
+  \item \AgdaFunction{pstakego}:
+    Stake distribution used for calculating the rewards.
+    This is the oldest stake distribution snapshot, labeled \AgdaField{go}.
+  \item \AgdaFunction{rs}: The rewards,
+    as calculated by the function \AgdaFunction{reward}.
+  \item \AgdaFunction{Δr₂}: The difference between the maximal amount of rewards
+    that could have been paid out if pools had been optimal,
+    and the actual rewards paid out.
+    This difference is returned to the reserves.
+  \item \AgdaFunction{÷₀}: Division operator that returns zero when the denominator is zero.
+\end{itemize}
+
+\begin{figure*}[h]
+\begin{AgdaMultiCode}
+\begin{code}
+
+createRUpd : ℕ → BlocksMade → EpochState → Coin → RewardUpdate
+createRUpd slotsPerEpoch b es total
+  = ⟦ Δt₁ , 0 - Δr₁ + Δr₂ , 0 - feeSS , rs ⟧
+  where
+    prevPp      = PParamsOf (es .EpochState.es)
+    reserves    = es .EpochState.acnt .Acnt.reserves
+    pstakego    = es .EpochState.ss .Snapshots.go
+    feeSS       = es .EpochState.ss .Snapshots.feeSS
+    stake       = pstakego .Snapshot.stake
+    delegs      = pstakego .Snapshot.delegations
+    poolParams  = ∅ᵐ -- FIXME: Add pool parameters to Snapshots
+
+    blocksMade = ∑[ m ← b ] m
+
+    rho = fromUnitInterval (prevPp .PParams.monetaryExpansion)
+    η = fromℕ blocksMade ÷₀ (fromℕ slotsPerEpoch * ActiveSlotCoeff)
+    Δr₁ = floor (ℚ.min 1 η * rho * fromℕ reserves)
+
+    rewardPot = pos feeSS + Δr₁
+    tau = fromUnitInterval (prevPp .PParams.treasuryCut)
+    Δt₁ = floor (tau * fromℤ rewardPot)
+    R = posPart (rewardPot - Δt₁)
+    circulation = total - reserves
+
+    rs = reward prevPp b R poolParams stake delegs circulation
+    Δr₂ = R - ∑[ c ← rs ] c
+
+\end{code}
+\end{AgdaMultiCode}
+\caption{RewardUpdate Creation}
+\label{fig:functions:createRUpd}
+\end{figure*}
+\end{NoConway}
+
 \begin{NoConway}
 \begin{figure*}[h]
 {\small
@@ -194,6 +270,9 @@ getOrphans es govSt = proj₁ $ iterate step ([] , govSt) (length govSt)
 
 \begin{figure*}[ht]
 \begin{AgdaSuppressSpace}
+\begin{code}[hide]
+open RwdAddr using (stake)
+\end{code}
 \begin{code}
 stakeDistr : UTxO → DState → PState → Snapshot
 stakeDistr utxo stᵈ pState = ⟦ aggregate₊ (stakeRelation ᶠˢ) , stakeDelegs ⟧
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -27,6 +27,7 @@ open import Ledger.Gov txs
 open import Ledger.Ledger txs abs
 open import Ledger.Prelude hiding (iterate)
 open import Ledger.Ratify txs
+open import Ledger.Rewards govStructure
 open import Ledger.Utxo txs abs
 
 open import Agda.Builtin.FromNat
@@ -36,24 +37,6 @@ open import Data.Nat.GeneralisedArithmetic using (iterate)
 open import Data.Nat.Properties using (+-0-monoid; +-0-commutativeMonoid)
 
 \end{code}
-\begin{NoConway}
-\begin{figure*}[ht]
-\begin{AgdaMultiCode}
-\begin{code}
-record RewardUpdate : Set where
-\end{code}
-\begin{code}[hide]
-  constructor ⟦_,_,_,_⟧ʳᵘ
-\end{code}
-\begin{code}
-  field
-    Δt Δr Δf : ℤ
-    rs : Credential ⇀ Coin
-\end{code}
-\end{AgdaMultiCode}
-\end{figure*}
-\end{NoConway}
-
 \begin{figure*}[ht]
 \begin{AgdaMultiCode}
 \begin{NoConway}
--- a/src/Ledger/Epoch.lagda
+++ b/src/Ledger/Epoch.lagda
@@ -42,9 +42,9 @@ open import Ledger.Utxo txs abs
 \begin{code}
 record Snapshot : Set where
   field
-    stake        : Credential ⇀ Coin
-    delegations  : Credential ⇀ KeyHash
-    -- poolParameters : KeyHash ⇀ PoolParam
+    stake           : Credential ⇀ Coin
+    delegations     : Credential ⇀ KeyHash
+    poolParameters  : KeyHash ⇀ PoolParams
 
 record Snapshots : Set where
   field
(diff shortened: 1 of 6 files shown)


src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+    determined by its performance.
+  \item \AgdaFunction{Δr₂}: The difference between the maximal amount of rewards
+    that could have been paid out if pools had been optimal,
+    and the actual rewards paid out.
+    This difference is returned to the reserves.
+  \item \AgdaFunction{÷₀}: Division operator that returns zero when the denominator is zero.
+\end{itemize}
+
+\begin{figure*}[h]
+\begin{code}
```
carlostome, line 187:
```suggestion
\begin{figure*}[h]
\begin{AgdaMultiCode}
\begin{code}
```

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+ActiveSlotCoeff : ℚ
+ActiveSlotCoeff = (1 / 20)
+\end{code}
```
carlostome, line 190:
```suggestion
ActiveSlotCoeff = (1 / 20)

\end{code}
```

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+    rewardPot = pos feeSS + Δr₁
+    tau = fromUnitInterval (prevPp .PParams.treasuryCut)
+    Δt₁ = floor (tau * fromℤ rewardPot)
+    R = posPart (rewardPot - Δt₁)
+    circulation = total - reserves
+
+    rs = reward prevPp b R poolParams stake delegs circulation
+    Δr₂ = R - ∑[ c ← rs ] c
+\end{code}
+\caption{RewardUpdate Creation}
```
carlostome, line 221:
```suggestion
\end{code}
\end{AgdaMultiCode}
\caption{RewardUpdate Creation}
```
So that there is one blank like between `ActiveSlotCoeff` and `createRUpd` instead of 2

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+  \item \AgdaFunction{rewardPot}: Total amount of coin available for rewards this epoch, as
+described in section 6.4 of \textcite{shelley-delegation-design}.
+  \item \AgdaFunction{feeSS}: The fee pot, containing the transaction fees from the epoch.
+    We use the fee pot from the snapshot taken at the epoch boundary.
+    Note that fees are not explicitly removed from any account:
+    the fees come from transactions paying them and are accounted for whenever transactions are processed.
+  \item \AgdaFunction{Δt₁}: The proportion of the reward pot that will move to the treasury,
+    as determined by the \AgdaField{treasuryCut} protocol parameter.
+    The remaining pot is called the \AgdaFunction{R},
+    just as in section 6.5 of \textcite{shelley-delegation-design}.
```
carlostome, line 172:
```suggestion
    just as in \textcite[\sectionname~6.5]{shelley-delegation-design}.
```

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+i.e.\ the net flow of Ada due to paying out rewards
+after an epoch.
+Relevant quantities are:
+\begin{itemize}
+  \item \AgdaArgument{prevPp}: Previous protocol parameters,
+    which correspond to the parameters during the epoch for which we are creating rewards.
+  \item \AgdaFunction{Δr₁}: Ada taken out of the reserves for paying rewards,
+    as determined by the \AgdaField{monetaryExpansion} protocol parameter.
+  \item \AgdaFunction{rewardPot}: Total amount of coin available for rewards this epoch, as
+described in section 6.4 of \textcite{shelley-delegation-design}.
```
carlostome, line 164:
```suggestion
described in \textcite[\sectionname~6.4]{shelley-delegation-design}.
```

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
 getStakeCred (a , _ , _ , _) = stakeCred a
 
-open RwdAddr using (stake)
 open GovActionState using (returnAddr)
 \end{code}
+\begin{NoConway}
+\Cref{fig:functions:createRUpd} defines
+the function \AgdaFunction{createRUpd}
+which creates a \AgdaRecord{RewardUpdate},
+i.e.\ the net flow of Ada due to paying out rewards
```
carlostome, line 155:
```suggestion
i.e.\@ the net flow of Ada due to paying out rewards
```
`\@` tells latex that the preceding period doesn't end a sentence.

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+    We use the fee pot from the snapshot taken at the epoch boundary.
+    Note that fees are not explicitly removed from any account:
+    the fees come from transactions paying them and are accounted for whenever transactions are processed.
+  \item \AgdaFunction{Δt₁}: The proportion of the reward pot that will move to the treasury,
+    as determined by the \AgdaField{treasuryCut} protocol parameter.
+    The remaining pot is called the \AgdaFunction{R},
+    just as in section 6.5 of \textcite{shelley-delegation-design}.
+  \item \AgdaFunction{pstakego}:
+    Stake distribution used for calculating the rewards.
+    This is the oldest stake distribution snapshot, labeled ``go''.
```
carlostome, line 175:
```suggestion
    This is the oldest stake distribution snapshot, labeled \AgdaField{go}.
```

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+  \item \AgdaFunction{rs}: The calculated rewards.
+    As given by \AgdaFunction{maxPool}, each pool can receive a maximal amount,
+    determined by its performance.
```
carlostome, line 178: This phrase is a bit difficult to understand the way its written. Also, it talks about this quantity determined by `maxPool`, how are these connected?

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+  \item \AgdaFunction{Δr₂}: The difference between the maximal amount of rewards
+    that could have been paid out if pools had been optimal,
+    and the actual rewards paid out.
+    This difference is returned to the reserves.
+  \item \AgdaFunction{÷₀}: Division operator that returns zero when the denominator is zero.
```
carlostome, line 183:
```suggestion
  \item \AgdaFunction{÷₀}: Division operator that returns zero when the denominator is zero.
```
This item seems to be of a different kind than the rest in the itemize. I'd suggest to change it to a footnote.

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+createRUpd : PParams → ℕ → BlocksMade → EpochState → Coin → RewardUpdate
+createRUpd prevPp slotsPerEpoch b es total
+  = ⟦ Δt₁ , 0 - Δr₁ + Δr₂ , 0 - feeSS , rs ⟧
+  where
+    reserves = es .EpochState.acnt .Acnt.reserves
+    pstakego  = es .EpochState.ss .Snapshots.go
+    feeSS    = es .EpochState.ss .Snapshots.feeSS
+    stake = pstakego .Snapshot.stake
+    delegs  = pstakego .Snapshot.delegations
+    poolParams = pstakego .Snapshot.poolParameters
```
carlostome, line 204:
```suggestion
    reserves = es .EpochState.acnt .Acnt.reserves
    pstakego = es .EpochState.ss .Snapshots.go
    feeSS = es .EpochState.ss .Snapshots.feeSS
    stake = pstakego .Snapshot.stake
    delegs = pstakego .Snapshot.delegations
    poolParams = pstakego .Snapshot.poolParameters
```

src/Ledger/Rewards.lagda
```diff
@@ -424,13 +424,101 @@ reward pp blocks rewardPot poolParams stake delegs total = rewards
...
 after an epoch.
-NOTE: The function \AgdaFunction{createRUpd} calculates the
+This type is defined in \Cref{fig:functions:RewardUpdate}.
+The update consists of four net flows:
+\begin{itemize}
+  \item \AgdaField{Δt}: The change to the treasury. This will be a positive value.
+  \item \AgdaField{Δr}: The change to the reserves. This will be a negative value.
+  \item \AgdaField{Δf}: The change to the fee pot. This will be a negative value.
+  \item \AgdaField{rs}: The map of new individual rewards,
+    to be added to the existing rewards.
```
carlostome, line 437: If we know for a fact that some of these fields are to be interpreted as positive/negative numbers which not make them correct by construction using naturals?

src/Ledger/Rewards.lagda
```diff
@@ -424,13 +424,101 @@ reward pp blocks rewardPot poolParams stake delegs total = rewards
...
+\begin{AgdaMultiCode}
+\begin{code}
+record RewardUpdate : Set where
+\end{code}
+\begin{code}[hide]
+  constructor ⟦_,_,_,_⟧ʳᵘ
+\end{code}
+\begin{code}
+  field
+    Δt Δr Δf : ℤ
```
carlostome, line 450: See previous comment.

carlostome (commented):
Left some comments here and there. So far it looks good. Thank you!

src/Ledger/Rewards.lagda
WhatisRT: There's a second issue here, which is that the claim for the reserves isn't true: https://github.com/IntersectMBO/cardano-ledger/issues/4771

src/Ledger/Epoch.lagda
HeinrichApfelmus: Unfortunately, making this change runs into error #753 .

However, it appears that by adding one empty newline after the function `createRUpd`, I can align these equal signs. Trying to align the other equal signs in the same function runs into the same error again, but at least these equal signs are aligned.

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+after an epoch.
+Relevant quantities are:
+\begin{itemize}
+  \item \AgdaArgument{prevPp}: Previous protocol parameters,
+    which correspond to the parameters during the epoch for which we are creating rewards.
+  \item \AgdaFunction{Δr₁}: Ada taken out of the reserves for paying rewards,
+    as determined by the \AgdaField{monetaryExpansion} protocol parameter.
+  \item \AgdaFunction{rewardPot}: Total amount of coin available for rewards this epoch, as
+described in section 6.4 of \textcite{shelley-delegation-design}.
+  \item \AgdaFunction{feeSS}: The fee pot, containing the transaction fees from the epoch.
```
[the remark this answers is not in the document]
HeinrichApfelmus: I have changed the sentence to describe "the epoch" more specifically.

src/Ledger/Epoch.lagda
HeinrichApfelmus: I have removed the mention of `maxPool` in favor of simply delegating the discussion to `reward`.

src/Ledger/Epoch.lagda
HeinrichApfelmus: I agree that this item feels different — but the problem is that I don't see a good location to which I could _attach_ the footnote. 🤔

My design rule for these itemizations is that they are equivalent to source code comments that I would write next to the definitions or usage sites of functions such as `Δr₁` or `÷₀`. In the case of `Δr₁`, it's a definition, but the case of `÷₀` feels different, because it's a usage site.

@williamdemeo — I think that these itemizations are of general interest to the topic of "conversion to printed formats" and "levels of detail" that we briefly discussed during the last meeting. 🤔 I see two things here:

* _Usage sites_. Here, the identifier `÷₀` is at a usage site. I think it would be great if this identifier could be hyperlinked to its definition. That would eliminate the need to describe it in the itemization. Of course, the drawback of excessive hyperlinking is that following the specifications becomes a nonlinear activity.
* _Levels of detail_. While I think that the itemization is helpful for the specification as is, I am not sure if it's at the right level of detail. I would assume that a high-level description of the `createRUpd` function does not show this itemization, but that a more detailed look inside the function would indeed show descriptions for all quantities such as `Δr₁`. Perhaps we can make progress by giving names to different levels:
  * "prose"
  * "definitions, abbreviated" — abbreviated notation using type class magic such as `+`.
  * "definitions, full" — precise notation such as `ℚ.+`.
  * "proofs, full" — also shows proofs. For the purpose of a specification, humans never need to look at proofs, because the compiler checks them for us; looking at the statements should be sufficient.

I suppose that "hyperlinking usage sites to their definitions" is a well-defined feature request, whereas I'm less clear on the "levels of detail". However, for the general problem of writing software that we can understand and verify, organizing the "levels of detail" effectively, be it through a formal specification or in code, ranks very high in importance.

src/Ledger/Rewards.lagda
HeinrichApfelmus: I have weakened the claim to "We typically expect this to be a negative value".

src/Ledger/Rewards.lagda
HeinrichApfelmus: Yes, that would be nice — though it also depends on how difficult the proof obligations are, and how much we care about them for now.

However, I would like to postpone this question to a later pull request when I consider the problem of adding the invariant that the quantities sum up to `0`.

For that later point in time, one design choice would be to use refinement types

```agda
record RewardUpdate : Set where
  field
    Δt : [ n ∈ ℤ ∣ 0 ≤ n ]
    Δr : ℤ
    Δf : [ n ∈ ℤ ∣ n ≤ 0 ]
    rs : Credential ⇀ Coin
    
    flowConservation : Δt + Δr + Δf + (∑[ c ← rs ] c) ≡ 0
```

though the statement of `flowConservation` now needs implicit conversions.

Another design choice would be to use ℤ and state the properties as part of the `record`:

```agda

record RewardUpdate : Set where
  field
    Δt Δr Δf : ℤ
    rs : Credential ⇀ Coin
    
    flowConservation : Δt + Δr + Δf + (∑[ c ← rs ] c) ≡ 0
    . Δt-positive    : 0 ≤ Δt
    . Δf-negative    : Δf ≤ 0
```

but then I may need to add the barely visible, but potentially important irrelevance annotations to the inequality proofs.

My current viewpoint is that it is generally beneficial to combine two different universes of discourse: a) `Type` — stuff that computes but has rather basic types and b) `Prop` — logical statement about that stuff, but which are fully erased at computation time. Refinement types like `RewardUpdate` do indicate that we need to be able to use b) inside a), but my feeling is that too much intermingling just makes things more difficult to understand again.

WhatisRT: I think I'd prefer the second option because it separates things more nicely, but it's hard do say. Anyway, for this case irrelevance is probably unnecessary, because `_≤_` already is irrelevant.

WhatisRT (approved):
Looks pretty good! Just some minor comments

src/Ledger/Rewards.lagda
HeinrichApfelmus: The second option would be my preference as well. I'll give it a try in the future.

src/Ledger/Epoch.lagda
```diff
@@ -156,9 +146,83 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+    that could have been paid out if pools had been optimal,
+    and the actual rewards paid out.
+    This difference is returned to the reserves.
+  \item \AgdaFunction{÷₀}: Division operator that returns zero when the denominator is zero.
+\end{itemize}
+
+\begin{figure*}[h]
+\begin{code}
+ActiveSlotCoeff : ℚ
+ActiveSlotCoeff = (1 / 20)
```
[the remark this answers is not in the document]
HeinrichApfelmus: I have added it to the `GlobalConstants` record. The only trouble with this is that these constants are currently not explained in the document. As a compromise, I have added a local explanation, to be moved elsewhere.

src/Ledger/Epoch.lagda
```diff
@@ -163,9 +145,88 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+\end{code}
+\begin{code}[hide]
+  {{nonZero}}
+\end{code}
+\begin{code}
+createRUpd : PParams → ℕ → BlocksMade → EpochState → Coin → RewardUpdate
+createRUpd prevPp slotsPerEpoch b es total
+  = ⟦ Δt₁ , 0 - Δr₁ + Δr₂ , 0 - feeSS , rs ⟧
+  where
+    reserves    = es .EpochState.acnt .Acnt.reserves
```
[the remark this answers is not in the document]
HeinrichApfelmus: Probably. In the absence of a definitive design decision on lenses, I would keep it in a state of "I can understand this well enough, but it could be prettier".

src/Ledger/Epoch.lagda
```diff
@@ -163,9 +145,88 @@ toRwdAddr x = record { net = NetworkId ; stake = x }
...
+\begin{code}
+ActiveSlotCoeff : ℚ
+ActiveSlotCoeff = (1 / 20)
+
+\end{code}
+\begin{code}[hide]
+  {{nonZero}}
+\end{code}
+\begin{code}
+createRUpd : PParams → ℕ → BlocksMade → EpochState → Coin → RewardUpdate
```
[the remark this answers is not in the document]
HeinrichApfelmus: Ah, thanks!

I had a misunderstanding about `HashProtected` — I thought that this type constructor would be a phantom type similar to `Hash` in the `cardano-ledger`, and thus not contain the actual data, but only its hash. But apparently, I can get the actual `PParams` out of a `HashProtected PParams` after all, thanks!

PR #755, opened 2025-04-14
Fix pp scripts in fls-shake

### LINKED ISSUE #754: Vertical vector regression
It seems we have a regression with post-processing. I didn't go and bisect the issue, but it seems the script is never being run (at least in CI).

carlostome:
# Description

Closes #754 

We should however think of a better approach to the way we do hldiff.

Right now the logic is handled by the python script `hldiff.py` while the replacement words are stored in `latex/hldiff_list.txt`.

A good solution might be to do this replacement directly in `fls-shake`.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/fls-shake/Main.hs
+++ b/fls-shake/Main.hs
@@ -12,7 +12,7 @@ Portability : POSIX
 module Main where
 
 import Development.Shake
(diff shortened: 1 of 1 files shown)


WhatisRT (approved):
LGTM

Issue #711, opened 2025-03-11
Value-TokenAlgebra cannot be instantiated


Ali-Hill:
This issue addresses problems encountered when trying to create a `TokenAlgebra` using `Value-TokenAlgebra ` found in `TokenAlgebra/ValueSet.lagda`. It is not currently possible to instantiate a TokenAlgebra using this library. This is needed in order to implement NFT's in concrete examples.  

**Blocking** 

1. Concrete implementation of contracts that have NFT's
2. Implementation of example's using structured contracts as they require NFT's

**Issue**

The first problem is that decidable equality is needed to be provided for total maps: `DecEq (TotalMap AssetId ℕ)`. This is not possible to construct.

If as a temporary solution it is postulated that setoid equality is needed instead of DecEq:

e.g. 
```
postulate
  setoidIsEq : (n m : TotalMap AssetId ℕ) → Dec ((TotalMap.rel n) ≡ᵉ (TotalMap.rel m)) → Dec (n ≡ m)  
```

Then it is possible to define: 

``` 
  setoidEqualityIsDecidable : ∀ (n m : TotalMap AssetId ℕ) → Dec ((TotalMap.rel n) ≡ᵉ (TotalMap.rel m))
This allowed me to define: DecEq (TotalMap AssetId ℕ).
```

Assuming this issue is solved, a TokenAlgebra can be constructed with Value-TokenAlgebra by passing in 6 arguments :
1. `ℙ AssetId`: A set of AssetId’s
2. `isMaximal assetIds`: A proof that the set of AssetId’s isMaximal
3. `(Dec-lookup≤ : ∀ {u v : AssetId ⇒ ℕ}
    → (∀ {a p q} → lookup u (a , p) ≤ lookup v (a , q)) ⁇)` 
4. `PolicyId` : A policy id for the TokenAlgebra
5. `AssetName` : The asset name for the TokenAlgebra.
6. `TotalMap AssetId Quantity → ℕ` : A size function for the TotalMap

For the purpose of running traces on the ledger implementation, I think it is currently possible to define 1,4,5,6 with 6 being a dummy implementation. It is not possible to construct 2 (isMaximal assetIds). I have not tried with 3 so am unaware how difficult this would be.

The `setoidIsEq` postulate may be discharged once we address: Issue #151 

**Example**

[...]

PR #757, opened 2025-04-15
Cleanup token algebra

carlostome:
# Description

This PR aims to organize and instantiate TokenAlgebra (possibly solving #711).

In particular,

- Moves Coin instance of TA to its own module.

~- Removes the less-or-equals than relation from TokenAlgebra.~

~This relation is used by the UTxO rule to check that every output of a transaction has at least `(overhead + utxoEntrySize txout) * coinsPerUTxOByte pp` coin. The way this check is currently performed is by comparing the coin at the level of Value, via injection. This relies somewhat implicitly on two things: `inject` does the "right thing" for assets other than Ada; and the less-or-equals relation for TA agrees with that of Coin. These requirements are not part of the mechanization.~

~Another issue is to define what kind of relation should the less-or-equals be. Since Value (as per the shelley-ma formal spec.) is meant to be an instance, the relation can not be a total order but at most a pre/partialorder.~

~To side-step these shortcomings, this PR proposes to remove the relation altogether from TA and adapt the check in the UTxO rule.~

- Removes `AssetName` and `specialAsset` from TokenAlgebra

These fields are not necessary since by construction Coin is a retract of TokenAlgebra, via `inject` and `coin`, and hence we already know how to "extract" Coin from a TA.
 
- Rename `property`

- Add instance for fixed number of multi-assets (in `ValueVector`)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Conformance/Utxo.agda
+++ b/src/Ledger/Conway/Conformance/Utxo.agda
@@ -95,7 +95,7 @@ data _⊢_⇀⦇_,UTXO⦈_ : UTxOEnv → UTxOState → Tx → UTxOState → Type
     ∙ L.refScriptsSize utxo tx ≤ pp .maxRefScriptSizePerTx
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ coin (getValueʰ txout)
+        inject ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
         serSize (getValueʰ txout) ≤ maxValSize pp
     ∙ ∀[ (a , _) ∈ range txoutsʰ ]
--- a/src/Ledger/Conway/Conformance/Utxo.agda
+++ b/src/Ledger/Conway/Conformance/Utxo.agda
@@ -95,7 +95,7 @@ data _⊢_⇀⦇_,UTXO⦈_ : UTxOEnv → UTxOState → Tx → UTxOState → Type
     ∙ L.refScriptsSize utxo tx ≤ pp .maxRefScriptSizePerTx
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
+        ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ coin (getValueʰ txout)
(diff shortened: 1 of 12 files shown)


src/Ledger/Conway/Conformance/Utxo.agda
```diff
@@ -95,7 +95,7 @@ data _⊢_⇀⦇_,UTXO⦈_ : UTxOEnv → UTxOState → Tx → UTxOState → Type
     ∙ L.refScriptsSize utxo tx ≤ pp .maxRefScriptSizePerTx
 
     ∙ ∀[ (_ , txout) ∈ txoutsʰ .proj₁ ]
-        inject ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ᵗ getValueʰ txout
+        ((overhead + L.utxoEntrySize txout) * coinsPerUTxOByte pp) ≤ coin (getValueʰ txout)
```
WhatisRT, line 98: This isn't equivalent to the previous line, since that one also ensures that no negative amounts are put in an output.

We either need a `TokenAlgebra` instance with negative amounts (at least being supported) because `mint` must be able to contain them to burn tokens, or we have to do what the implementation does (I think) which is to have a bunch of different types. I much prefer having this inequality over another type.

WhatisRT:
> Another issue is to define what kind of relation should the less-or-equals be. Since Value (as per the shelley-ma formal spec.) is meant to be an instance, the relation can not be a total order but at most a pre/partialorder.

That's fine, we have a `HasPreorder` typeclass that can be used here.

carlostome:
I reverted the removal of ≤ᵗ from this PR, and changed `VectorValue` accordingly.

The discussion about removing ≤ᵗ from Token Algebra has been moved to its own issue (#761)

src/Ledger/TokenAlgebra/ValueVector.agda
```diff
@@ -0,0 +1,50 @@
...
+
+Quantity = ℕ
+
+Vec-commutativeMonoid = Vec.commutativeMonoid n +-0-commutativeMonoid
+
+module +-0-commutativeMonoid = Alg.CommutativeMonoid +-0-commutativeMonoid
+module Vec-commutativeMonoid = Alg.CommutativeMonoid Vec-commutativeMonoid
+
+module _ (Policies : Vec PolicyId n) where
+  Value-TokenAlgebra : TokenAlgebra
```
WhatisRT, line 26: I think it would be good to also implement the FFI for this type in this PR, since that seems to be the main point of this particular algebra. I'd say it should map to a `Coin × List (PolicyId × String × ℕ)`, and the conversion to the Agda type should error if any of the `String`s are nonempty or there are any duplicate `PolicyId`s.

WhatisRT (approved):
LGTM, I've approved it in case you want to merge it now and have other plans for the FFI

WhatisRT:
Ah, I didn't see the auto merge, I guess it'll be future work then

PR #781, opened 2025-05-13
Add invariants to `RewardUpdate`

HeinrichApfelmus:
# Description

This pull requests adds invariants to the type `RewardUpdate` and adds the proofs for these invariants to `createRUpd`.

### Comments

* I managed to confine the use of proof-irrelevance to the module `Ledger.Conway.Types.Numeric.UnitInterval` by using decidability via the function `fromUnitInterval-inUnitInterval`. In this way, we can avoid any discussion of this topic in the human-readable definition of `RewardUpdate`.
* The module `Data.Rational.Properties` uses special type classes such as `NonNegative` for the predicate `0 ≤`. I wasn't quite sure how to use these classes effectively — their main purpose seems to hide/automate the proof obligation that `NonZero y` in the division `x ÷ y`, but these classes have enough lemmas associated with them that they can compete with the properties of `≤`. I'm not entirely convinced about the design of the std-lib here, I have tried to formulate properties in terms of `0 ≤` for the sake of readability while using what is available for proofs.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/lib-exts/stdlib/Data/Rational/Properties.agda
+++ b/lib-exts/stdlib/Data/Rational/Properties.agda
@@ -65,8 +65,8 @@ fromℕ-0≤ : ∀ (n : ℕ) → 0 ≤ fromℕ n
 fromℕ-0≤ n = fromℤ-0≤ (ℤ.+ n) (+≤+ z≤n)
 
 -- The product of 2 non-negative numbers is non-negative.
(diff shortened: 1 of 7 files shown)


williamdemeo (approved):
This is great work! Nice job.  Instead of making comments and suggestions, I simply committed some minor (mostly stylistic) changes so the code is more consistent with the rest of the ledger code base and to improve some indentation, so you don't have to make the changes yourself.  I hope you don't mind.

PR #802, opened 2025-06-20
Complete implementation of `stakeDistr`

### LINKED ISSUE #679: Complete the implementation of `stakeDistr`

HeinrichApfelmus:
# Description

This pull request completes the implementation of `stakeDistr`. Resolve #679.

**Comments**

* I have introduced the value `activeDelegs` as it appears in Figure 37 of the original Shelley specification.
* I have used `mapWithKeys` and removed `∪` in order to avoid a potential situation where the set-theoretical union could merge identical coin values.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Specification/Rewards.lagda
+++ b/src/Ledger/Conway/Specification/Rewards.lagda
@@ -621,8 +621,8 @@ opaque
       activeDelegs  : Credential ⇀ KeyHash
       activeRewards : Credential ⇀ Coin
       activeStake   : Credential ⇀ Coin
- \end{code}
- \begin{code}
+\end{code}
+\begin{code}
       poolParams     = pState .PState.pools
       open DState dState using (stakeDelegs; rewards)
       utxoBalance    = λ cred → cbalance (utxo ∣^' λ txout → getStakeCred txout ≡ just cred)


WhatisRT:
I'm really surprised a change underneath `opaque` can affect unification (which I assume in the issue here) outside of `opaque`.

With `--profile=definitions`, I get `Ledger.Conway.Epoch.Properties._.EPOCH-complete    23,210ms` before this change (whether `stakeDistr` is `opaque` or not) which is already quite much. My feeling is that there is some kind of multiplicative buildup - adding a factor of 200 is enough to get to the time you're seeing.

So I think there are two approaches here, and fixing one might be good enough:
1. investigate why the original definition is so slow and speed that up
2. investigate why this change slows things down

I feel like 2 might be an Agda bug actually. Changing code under `opaque` should never slow down typechecking (except of the changed definition itself of course). Maybe my internal model is wrong here but I can't think of a good reason why. So it might be worth trying to minimize this (but this might be a lot of work!)

Using interactive highlighting shows that the typechecker is stuck on `cong₂` for a really long time (both in the original and with this change) so it's probably one of the implicit arguments to `cong₂` being inferred that's so slow.

carlostome:
I created an issue (#850) so (hopefully) somebody investigates why my "Refactor EPOCH-complete" commit fixes the typechecking time problem.

williamdemeo (approved):
Nice work Heinrich.  And great job refactoring the proofs and getting the type-checking time down, Carlos!

PR #835, opened 2025-07-21
Fix build on Mac

### LINKED ISSUE #827: Error in `nix-build` on macos
I get this error when running `nix-build` or `nix-build -A formal-ledger`:
```
error: cycle detected in build of '/nix/store/kklkks1vpg3i0wvdkx8ryl9nfhlfv2j7-fls-agda-2.7.0.1.drv' in the references of output 'bin' from output 'out'
```

I get this warning a bit earlier in the build, which doesn't look relevant, but could be:
```
Warning: supplying the --target arm64-apple-darwin != aarch64-apple-darwin argument to a nix-wrapped compiler may not work correctly - cc-wrapper is currently not designed with multi-target compilers in mind. You may want to use an un-wrapped compiler instead.
```

Nix version is `2.24.14`.

vrom911:
# Description

Resolves #827 

As spotted in the issue discussion, this is related to the circular dependency due to the naming differences for outputs on Macos.

This is the attempt of fixing this.
I would need CI to make sure that this is still working on Linux though 🙏🏼 

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/build-tools/agda/nix/fls-agda.nix
+++ b/build-tools/agda/nix/fls-agda.nix
@@ -2,15 +2,8 @@
   sources ? import ../../nix/sources.nix,
   nixpkgs ? import sources.nixpkgs { },
 }:
-let
-  fls-agda-override = oldAttrs: {
-    outputs = [ "bin" ] ++ (oldAttrs.outputs or [ ]);
-    postInstall =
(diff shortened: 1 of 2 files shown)


carlostome (approved):
LGTM, thank you!

Issue #675, opened 2025-02-04
Remaining Shelley features/proofs


williamdemeo:
This issue itemizes features and proofs that are missing (i.e., not yet formalized and/or described in the full cardano-ledger.pdf) from the Shelley era:

- [ ] [CHAIN](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.75): incomplete (out of scope)
    + Dependencies:
      - PRTCL: missing - now part of Consensus
      - TICKN: missing - now part of Consensus
      - TICK: partially missing (not missing anymore since #869)
          - The NEWEPOCH logic is partially inlined in CHAIN (it has been moved to TICK)
          - CHAIN is missing the logic of RUPD (implemented in TICK since #869)
   + [Functions](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.74):
        - `chainChecks`: missing - now part of Consensus
        - `prtlSeqChecks`: missing - now part of Consensus
        - `adoptGenesisDelegs`: missing - deprecated

- [x] BBODY: partially missing. Done in #884
   + LEDGERS: logic is inlined
   + CHAIN is missing preconditions:
     - `bBodySize txs = hBbsize bhb`
     - `bbodyhash txs = bhash`
   + CHAIN is missing the logic to update the block via `incrBlocks`

- [x] [TICK](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.64): missing (5 days) Done in #869
    + Dependencies:
        - RUPD: missing
        - NEWEPOCH: partially inlined in CHAIN
    + [Functions](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.63):
        - `adoptGenesisDelegs`: missing - deprecated

- [x] [RUPD](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.62): missing (10 days) Done in #869
    + Dependencies:
    + [Functions](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.63):
        - `createRUpd`: missing (#706)
        - monetary and treasury expansion parameters (#707)

- [x] [NEWEPOCH](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.57): incomplete. Done in #854
[...]

PR #854, opened 2025-08-08
Implement calculatePoolDistr

facundominguez:
This PR implements the `calculatePoolDistr` function of epoch transitions from Shelley.  (See [issue 675](https://github.com/IntersectMBO/formal-ledger-specifications/issues/675), and [the shelley spec](https://github.com/intersectmbo/cardano-ledger/releases/latest/download/shelley-ledger.pdf#figure.56), Fig. 56.)

calculatePoolDistr

### DIFF
--- a/src/Ledger/Conway/Specification/Rewards.lagda.md
+++ b/src/Ledger/Conway/Specification/Rewards.lagda.md
@@ -565,10 +565,10 @@ distribution snapshot.  Such a snapshot contains the essential data needed to co
 rewards.
 
 - `stake`{.AgdaField} A stake distribution, that is a mapping from
-  credentials to coin.
+  stake credentials to the active stake that they own in coins.
 
 - `delegations`{.AgdaField}: A delegation map, that is a mapping from
-  credentials to the stake pools that they delegate to.
+  stake credentials to the stake pools that they delegate to.
 
 - `poolParameters`{.AgdaField}: A mapping that stores the pool
   parameters of each stake pool.


src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -446,6 +447,21 @@ its results by carrying out each of the following tasks.
 
 \begin{NoConway}
 \begin{figure*}[ht]
+\begin{code}[hide]
+calculatePoolDistr : Snapshot → PoolDistr
+calculatePoolDistr ss =
+    zipWithᵐ (λ c pp → (? , ?) ) sd (poolParameters ss)
```
facundominguez, line 453: One challenge here is that I need to prove for the first component of the pair that c/totalStake is a number between 0 and 1.

Another challenge is that pool parameters don't seem to have the `poolVRF` field that the spec requires. I need to find out what to use instead.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -446,6 +447,21 @@ its results by carrying out each of the following tasks.
...
+  where
+    open Snapshot
+    totalStake = ∑[ c ← stake ss ] c
+
+    sd : KeyHash ⇀ Coin
+    sd = aggregateBy (delegations ss ˢ) (stake ss)
+
+    zipWithᵐ : {A B C D : Type} → (B → C → D) → A ⇀ B → A ⇀ C → A ⇀ D
+    zipWithᵐ = ?
```
facundominguez, line 462: Perhaps I need to implement this from scratch (?)

facundominguez: One idea to implement it:
```
map fromJust $ filter isJust $ unionWith (fold (const nothing) (const nothing) (just . uncurry _,_))
```
Some fiddling will be necessary to justify the `fromJust` call.

facundominguez: Went with the later one.

src/Ledger/Conway/Specification/Epoch.lagda
facundominguez: I'm clamping for now.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -446,6 +447,34 @@ its results by carrying out each of the following tasks.
...
+      totalStake : Coin
+      totalStake = ∑[ c ← ss .stake ] c
+
+      normalizeStake : Coin → UnitInterval
+      normalizeStake c = clamp $ case totalStake of λ where
+        zero → Data.Rational.normalize 0 (suc zero)
+        (suc n) → Data.Rational.normalize c (suc n)
+
+      sd : KeyHash ⇀ Coin
+      sd = aggregateBy (ss .delegations ˢ) (ss .stake)
```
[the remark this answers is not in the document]
williamdemeo: `aggregateBy` is also used in `Rewards.lagda`, but I think the `Epoch.lagda` module comes first, so this is as good a place as any to explain it, imo.

williamdemeo:
It might be helpful to add an image of (the Shelley pdf definition of) the function this PR defines to the PR description above (in the message "This PR defines...").  You should be able to add the screenshot I created by editing the description message and pasting in the following line:
```
<img width="1362" height="611" alt="calculatePoolDistr" src="https://github.com/user-attachments/assets/d425f044-fbcd-4af3-a232-241ee90547e9" />
```

src/Ledger/Conway/Specification/Epoch.lagda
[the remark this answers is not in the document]
facundominguez: Added.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -445,7 +460,38 @@ its results by carrying out each of the following tasks.
...
```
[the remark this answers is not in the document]
facundominguez: Ok. Do you mean that once we find out, we should do exactly as the implementation?

facundominguez:
I rebased master once more to solve conflicts. The last four commits are new with respect to the last review.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -85,13 +85,20 @@ instance
   HasPParams-EpochState .PParamsOf = PParamsOf ∘ EnactStateOf
 \end{code}
 \begin{NoConway}
+
+`PoolDistr`{.AgdaBound} in the Shelley spec results from dividing the coins in
+`PoolDelegatedStake`{.AgdaBound} by the total stake in
+`PoolDelegatedStake`{.AgdaBound}.
```
carlostome, line 91: - Since this file is still latex, I'd suggest to use latex markup so the pipeline doesn't get confused.
```suggestion
\AgdaBound{PoolDistr} in the Shelley spec results from dividing the coins in
\AgdaBound{PoolDelegatedStake} by the total stake in
\AgdaBound{PoolDelegatedStake}.
```
- I'd add a reference to the definition of PoolDistr in the Shelley spec
- If I understand correctly, this paragraph aims to differentiate the type PoolDelegatedStake introduced here from the type used in Shelley spec (PoolDistr). I suggest to expand on the explanation, because without context is hard to understand.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -85,13 +85,20 @@ instance
...
+
 \begin{code}
+PoolDelegatedStake = KeyHash ⇀ (Coin × KeyHash)
```
carlostome, line 94: ATM we don't have VRF's as part of the Agda spec, so I'd suggest to remove it from the codomain of this type.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -344,6 +351,20 @@ opaque
...
+and returns a function that maps each \AgdaBound{a} in the domain of
+\AgdaBound{m} to the sum of all \AgdaBound{b} such that
+(\AgdaBound{a}, \AgdaBound{b}) ∈ \AgdaBound{R}.
+
+In the definition of \AgdaFunction{mkStakeDistrs}, the relation and map passed to
+\AgdaFunction{aggregateBy} are
+\AgdaFunction{∣} \AgdaBound{delegations} \AgdaFunction{∣} :
+ℙ \AgdaDatatype{Credential} \AgdaFunction{×} \AgdaDatatype{VDeleg} and
+\AgdaFunction{stakeOf} \AgdaBound{ss} \AgdaFunction{∪⁺}
+\AgdaFunction{gaDepositStake} \AgdaBound{govSt} \AgdaBound{ds}, respectively.
```
carlostome, line 366: This explanation seems a bit out of place.

src/Ledger/Conway/Specification/Epoch.lagda
```diff
@@ -535,7 +557,29 @@ its results by carrying out each of the following tasks.
...
 
 \begin{NoConway}
+The \AgdaFunction{calculatePoolDistr} produces a new pool distribution from the
+delegation map and stake allocation of the previous epoch.
+
 \begin{figure*}[ht]
+\begin{code}
+opaque
+  calculatePoolDistr : Snapshot → PoolDelegatedStake
+  calculatePoolDistr ss =
```
carlostome, line 567: I wonder if we should keep or change the name of this function since it is not calculating the PoolDistr as in the Shelley spec, but something else.

facundominguez: The name makes easy to find the corresponding spot in the Agda spec. We could still change it to something like
```
\AgdaFunction{calculatePoolDelegatedStake} performs the computation of
\AgdaFunction{calculatePoolDistr} in the Shelley spec, without normalizing the
stakes to be between 0 and 1.
\begin{code}
calculatePoolDelegatedStake : Snapshot → PoolDelegatedStake
...
```

src/Ledger/Conway/Specification/Epoch.lagda
williamdemeo: I think we can put the explanation there for now and move it to a more appropriate place later.  It's good to have it *somewhere* and we shouldn't let its placement hold up Facundo's PR any longer.

src/Ledger/Conway/Specification/Epoch.lagda
williamdemeo: I'm confused. Don't you need `poolVRF` in order to define the `calculatePoolDistr` function, as in the Shelley spec?  Is the absence of `VRF` explained by the comment you added in your last commit?  I'll invite you to the call with Carlos tomorrow so we can discuss this and get this PR approved.

williamdemeo (approved):
Looks good.  Nice work!  I think you can merge this once you rebase one more time on master, and make sure it still works with the new upgrade to Agda 2.8!

PR #857, opened 2025-08-14
Ensure stake distribution only contains valid SPO delegations

Soupstraw:
# Description

# Checklist

- [ ] Self-reviewed the diff


carlostome:
The changes introduced in this PR seem to be part already of the Specification (introduced by #679) ([source](https://github.com/IntersectMBO/formal-ledger-specifications/blob/201d6f311e33ec26d35059cdc8464df192aaf8e1/src/Ledger/Conway/Specification/Rewards.lagda#L613). However, ATM, these are not being used by the Conformance modules.

I have opened an issue regarding this and other changes that we need to integrate with the Conformance modules: #873

Soupstraw:
After meeting with @carlostome, we figured that it's best to make the spec and implementation match up during tranlation. The reason for the difference in the outputs was actually that the spec stores both DRep delegations and SPO delegations in the snapshots, while the implementations stores the DRep delegations separately in the GovState.

PR #871, opened 2025-08-22
Fix the broken table in the `Gov.Actions` module

williamdemeo:
# Description

The LaTeX `longtable` (and even the simple `table`) environment causes pandoc to put the markdown version of the table inside a `<div>` `</div>` block and this apparently prevents mkdocs from rendering the table correctly, despite the fact that we are using the markdown extension `md_in_html`.  Until a better solution is found, a workaround is to simply omit the table environment altogether, just use a tabular environment, and move the caption ("Types of Governance Actions") to a subsection title called "Table: Types of Governance Actions".

Also, for some reason, pandoc silently deletes the footnote in the table.

To resolve both the table rendering issue and the lost footnote, I suggest we hand-tweak the `Actions.lagda.md` file and commit it to the repo soon.  For now, however, the workaround in the PR is an easy improvement.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Specification/Gov/Actions.lagda
+++ b/src/Ledger/Conway/Specification/Gov/Actions.lagda
@@ -110,24 +110,23 @@ The governance actions carry the following information:
   \item \ChangePParams{}: the updates to the parameters; and
   \item \TreasuryWdrl{}: a map of withdrawals.
 \end{itemize}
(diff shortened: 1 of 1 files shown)


carlostome (approved):
LGTM!

PR #925, opened 2025-09-22
Define a step function to run the EPOCH rule from the specification

facundominguez:


facundominguez:
I reached the point where I can run
```
 --match "/Conway/Imp/Conformance with constrained generators/Ticks transition graph/EPOCH/"
```
and get a long error message describing a failure. I need to make sense of it next.

PR #953, opened 2025-10-16
Implement fPoolParams as in the Shelley specification

### LINKED ISSUE #951: Implement fPoolParams from the Shelley spec
`fPoolParams` is used in the Pool transition relation. It is needed to address #948.

In the implementation, `fPoolParams` materializes as `psFutureStakePools`. Here's the implementation of the update in the `Pool-reReg` rule [[link]](https://github.com/IntersectMBO/cardano-ledger/blob/28d98878cedd967e640664bfa3f742648dc86dd4/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs#L298).

facundominguez:
Fixes #951

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Conway spec
 
 ### WIP
-
+- Implement the fPoolParams field of PState as in the Shelley specification
 - Add the BBODY transition relation from Shelley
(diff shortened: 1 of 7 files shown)


carlostome (approved):
LGTM!

PR #973, opened 2025-10-31
Don't count pool deposits when reregistering pools

facundominguez:
This PR modifies updates `updateCertDeposits` so pool deposits are not counted again when reregistering a pool. This is the function used by the UTXO rule to update deposits.

The change to the function is fairly simple, but then there are a few broken proofs that are fixed in later commits.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -701,6 +701,27 @@ See [the `conformance-example` directory][conformance-example].
 ## 🗃️ Miscellanea
 
 
+### Updating Agda dependencies in the Nix configuration
+
+The following example ilustrates the procedure
+
+```
+nix flake update agda-nix/abstract-set-theory \
+  --override-input agda-nix/abstract-set-theory \
+  github:input-output-hk/agda-sets/bbaa00abc4aef061896ae5d3cdec148bfbf5029f
+nix build ./#fls-agdaWithPackages -o ~/ledger-agda
+```
+
+The first line updates the commit hash to use for a dependency. In the example,
+it updates the dependency `agda-nix/abstract-set-theory` to point at the commit
+`bbaa00abc4aef061896ae5d3cdec148bfbf5029f` or the repository
+`github:input-output-hk/agda-sets`.
+
+The second line rebuilds the Agda mode to use with emacs. This step is necessary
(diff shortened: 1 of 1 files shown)


src/Ledger/Conway/Specification/Ledger/Properties/Base.lagda.md
```diff
@@ -113,6 +113,88 @@ module LEDGER-PROPS (tx : Tx) (Γ : LEnv) (s : LState) where
...
+        { Carrier = ℙ A
+        ; _≈_ = _≡ᵉ_
+        ; _≤_ = _⊆_
+        ; isPartialOrder = ⊆-PartialOrder
+        }
+      open import Relation.Binary.Reasoning.PartialOrder ⊆-Poset
+      open import Relation.Binary.Structures using (IsEquivalence)
+      module ≡ᵉA = IsEquivalence (≡ᵉ-isEquivalence {A = A})
+      open import Relation.Binary.Reasoning.Syntax
+      open ⊆-syntax _IsRelatedTo_ _IsRelatedTo_ ≤-go public
```
facundominguez: The imports are as long as the proof. Maybe we should create a module
`Axiom.Set.Reasoning` to reexport these?

src/Ledger/Conway/Specification/Ledger/Properties/Base.lagda.md
```diff
@@ -113,6 +113,88 @@ module LEDGER-PROPS (tx : Tx) (Γ : LEnv) (s : LState) where
...
+  _∎→ : ∀ (A : Type) → A → A
+  A ∎→ = id
+  infix 3 _∎→
+
+  _→⟨_⟩_ : ∀ (x : Type) {y z : Type} → (x → y) → (y → z) → x → z
+  _ →⟨ f ⟩ g = g ∘ f
+  infixr 2 _→⟨_⟩_
+
+  begin→ : ∀ {A B : Type} → A → (A → B) → B
+  begin→ a f = f a
```
facundominguez: These are used in the proofs below. Do they belong here or in some more general place?

facundominguez:
I rebased master and updated agda-sets. Maybe this is ready to merge now.

Issue #723, opened 2025-03-21
Remaining Alonzo features/proofs


Ali-Hill:
The current Alonzo implementation is missing definitions that reference `serialization` and `POSIXTIME` in general. Also, some features that are no longer supported such as `MIR Certificates`  are not included. A detailed list it provided below:

Figure 1: Definitions Used in Protocol Parameters

- Known issue where we can't make costmdls `Language →/⇀ CostModel` as it does not work with DecEq
`
- All implemented except for `getLanguageView`

Figure 2: Definitions for Transactions 

- All implemented except for the `hashScriptIntegrity` function

Figure 5: Indexing script and data objects

- Need to add notion of `posixtime`, `EpochInfo`, `SystemStartSystem` and `UTCTimeSystem`

Figure 6: Script Validation, cont.
 
- Add `epochInfoSlotToUTCTime` function

Figure 7: Scripts and their arguments

- Update `collectTwoPhaseScriptInputs` function to include additional arguments:
  - `EpochInfo`
  - `SystemStart`

Figure 9: State update rules

- Add additional arguments to `collectTwoPhaseScriptInputs`:
  - `EpochInfo`
  - `SystemStart`

Figure 10: UTxO inference rules

- Update condition referencing `epochInfoSlotToUTCTime`

Figure 13: UTxO with witnesses inference rules for Tx

- Add `hashScriptIntegrity` case to the UTxOW rule

Figure 14: Ledger inference rules 

Is LEDGER-I Equivalent to LEDGER-NV?

Figure 15: BBody rules 

Add whole figure 

Figure 16: Shelley to Alonzo State Transition 

Add whole figure

Figure 17/18/19/20 : MIR Certificates and Rules

None of this is in the ledger but it is intentional according to comment in Ledger/Certs: 

`Support for pointer addresses, genesis delegations and MIR
certificates is removed (see \hrefCIP{1694} and \textcite{cip1694}).`

Figure 22: TxInfo and Constituent Types

- Update TxInfo to use `POSIXTIME`

Figure 23: Types and Functions Used in Time Conversion 

Add whole figure 

Figure 24: TxInfo Constituent Type Translation Functions

Add whole figure 

Figure 26: Transaction Summarization Functions

- Update txInfo to `POSIXTIME` and include `EPochInfo` and `SystemStart`

PR #1035, opened 2026-01-15
Make costmdls a map as in the Alonzo spec

facundominguez:
Addresses part of #723

### DIFF
--- a/src/Ledger/Conway/Conformance/Utxow.agda
+++ b/src/Ledger/Conway/Conformance/Utxow.agda
@@ -53,7 +53,7 @@ data _⊢_⇀⦇_,UTXOW⦈_ where
     ∙  neededScriptHashes - refScriptHashes ≡ᵉ witsScriptHashes
     ∙  inputsDataHashes ⊆ txdatsHashes
     ∙  txdatsHashes ⊆ inputsDataHashes ∪ outputsDataHashes ∪ refInputsDataHashes
-    ∙  L.languages tx utxo ⊆ L.allowedLanguages tx utxo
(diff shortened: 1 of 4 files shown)


williamdemeo (changes requested):
## key issue/possible bug

Using `lookupᵐ?` and `mapPartial` in `collectP2ScriptsWithContext` might result in silently skipping scripts.

In both Conway and Dijkstra we define `collectP2ScriptsWithContext` with a helper,

`toScriptInput : ScriptPurpose → ScriptHash → Maybe (P2Script × List Data × ExUnits × CostModel)`

which (in this PR) uses `costModel ← lookupᵐ? (PParams.costmdls pp) (language p2s)`

`collectP2ScriptsWithContext` builds the list via:

```agda
setToList
    $ mapPartial (λ (sp , c) → if isScriptObj c
                                then (λ {sh} → toScriptInput sp sh)
                                else nothing)
    $ credsNeeded utxo (tx .Tx.body)
```

So, if a cost model is missing for some language, then `lookupᵐ?` returns `nothing`, which makes `toScriptInput` return `nothing`, which makes `mapPartial` drop that script from the list, which makes `evalP2Scripts` run `all` over a list that is missing a required script---i.e., the transaction could incorrectly pass because a script never ran.

This is a real risk, unless we ensure that a missing cost model is impossible.

### Suggested Fix

I suggest we add the following to the list of premises of the `UTXOW` rule:

```agda
    ∙  languages tx utxo ⊆ dom (PParams.costmdls (PParamsOf Γ))
```

If we put that in `UTXOW-inductive`, then `lookupᵐ?` should never fail in reachable states, and the current implementation is okay (though maybe still a bit too permissive in the `toScriptInput` helper).

williamdemeo (approved):
Looks good.  LGTM! ...but let's remember to make an issue to add cost models to Dijkstra.

PR #1134, opened 2026-03-23
Add Dijkstra Foreign modules

carlostome:
# Description

This PR adds the `Foreign` hierarchy of modules for the Dijkstra era.

In addition,
- changes `DecEq Network` module parameter from a typeclass instance to an explicit parameter to work around Agda bug.
- it removes the `TxLevel` type parameter from transaction witnesses as it is not necessary
- adds `DecEq` instances for several types (needed for conformance)

~Blocked by #1133.~

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Foreign/Account.lagda.md
+++ b/src/Ledger/Dijkstra/Foreign/Account.lagda.md
@@ -24,7 +24,7 @@ private
     if      s ≟ "⟦_,_⦆" then "Both"
     else if s ≟ "⟦_,∞⦆" then "Lower"
     else if s ≟ "⟦0,_⦆" then "Upper"
-    else ""
+    else s
 
 instance
   HsTy-BalanceInterval = autoHsType BalanceInterval ⊣ onConstructors toHaskellConstr
--- a/src/Ledger/Dijkstra/Foreign/Account.lagda.md
+++ b/src/Ledger/Dijkstra/Foreign/Account.lagda.md
@@ -1,0 +1,31 @@
+---
+source_branch: master
+source_path: src/Ledger/Dijkstra/Foreign/Account.lagda.md
+---
+```agda
+module Ledger.Dijkstra.Foreign.Account where
+
+open import Foreign.Convertible
+open import Foreign.Convertible.Deriving
+open import Foreign.HaskellTypes
+open import Foreign.HaskellTypes.Deriving
+import Data.String as S
+
+open import Ledger.Prelude
+open import Ledger.Prelude.Foreign.HSTypes
+
+open import Ledger.Core.Foreign.Address
+open import Ledger.Core.Foreign.Crypto.Base
+open import Ledger.Dijkstra.Foreign.HSStructures
+
+private
+  toHaskellConstr : String → String
+  toHaskellConstr s =
+    if      s ≟ "⟦_,_⦆" then "Both"
+    else if s ≟ "⟦_,∞⦆" then "Lower"
+    else if s ≟ "⟦0,_⦆" then "Upper"
+    else ""
+
+instance
+  HsTy-BalanceInterval = autoHsType BalanceInterval ⊣ onConstructors toHaskellConstr
+  Conv-BalanceInterval = autoConvert BalanceInterval
--- a/src/Ledger/Dijkstra/Foreign/Chain.lagda.md
+++ b/src/Ledger/Dijkstra/Foreign/Chain.lagda.md
@@ -51,25 +51,6 @@ record HSBlock : Type where
 
 open HSBlock
 
-convBlock : HSBlock → Block
-convBlock b
-  with bBodySize b ≟ BHBody.hBbsize (BHeader.bhbody (bheader b))
-... | no _ = error $ "bBodySize check failed: " S.++
-               show (bBodySize b) S.++ " ≠ " S.++
-               show (BHBody.hBbsize (BHeader.bhbody (bheader b)))
-... | yes p
-  with bBodyHash b ≟ BHBody.bhash (BHeader.bhbody (bheader b))
-... | no _ = error $ "BBodyHash check failed: " S.++
-               show (bBodyHash b) S.++ " ≠ " S.++
-               show (BHBody.bhash (BHeader.bhbody (bheader b)))
-... | yes q = record
-  { bheader = bheader b
-  ; ts = ts b
-  ; bBodySize = bBodySize b
-  ; ≡-bBodySize = p
-  ; ≡-bBodyHash = q
-  }
-
 instance
   Conv-Block-HSBlock : Convertible Block HSBlock
   Conv-Block-HSBlock .to b = record { Block b }
--- a/src/Ledger/Dijkstra/Foreign/Chain.lagda.md
+++ b/src/Ledger/Dijkstra/Foreign/Chain.lagda.md
@@ -1,0 +1,113 @@
+---
+source_branch: master
+source_path: src/Ledger/Dijkstra/Foreign/Chain.lagda.md
+---
+```agda
+module Ledger.Dijkstra.Foreign.Chain where
+
+import Data.String as S
+open import Foreign.Convertible
+open import Foreign.Convertible.Deriving
+open import Foreign.HaskellTypes
+open import Foreign.HaskellTypes.Deriving
+
+open import Ledger.Prelude
+open import Ledger.Prelude.Foreign.Util
+open import Ledger.Prelude.Foreign.HSTypes
+
+open import Ledger.Core.Foreign.Address
+open import Ledger.Core.Foreign.Crypto.Base
+open import Ledger.Dijkstra.Foreign.HSStructures
+open import Ledger.Dijkstra.Foreign.PParams
+open import Ledger.Dijkstra.Foreign.Cert
+open import Ledger.Dijkstra.Foreign.Enact
+open import Ledger.Dijkstra.Foreign.Gov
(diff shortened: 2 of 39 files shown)


src/Ledger/Dijkstra/Specification/Utxow.lagda.md
```diff
@@ -139,11 +139,11 @@ hashScriptIntegrity
   → RedeemerPtr ⇀ (Redeemer × ExUnits)
   → ℙ Datum
   → Maybe ScriptHash
-hashScriptIntegrity pp langs rdrms dats
-  with rdrms ˢ ≟ ∅ˢ | langs ≟ ∅ˢ | dats ≟ ∅ˢ
-...  | yes _        | yes _      | yes _ = nothing
-...  | _            | _          | _     =
-    just $ hash (dats , rdrms , mapˢ (getLanguageView pp) langs)
+hashScriptIntegrity pp langs rdrms dats = nothing
```
williamdemeo: This change effectively disables script-integrity hashing for the entire Dijkstra spec, not just for foreign extraction.

If this is a deliberate placeholder until `hash` is wired through to a real Haskell hash function, consider replacing the commented-out code with a `-- TODO(#NNNN): …` referencing a tracking issue and explain what the blocker is.

If it's a leftover from local debugging, can you restore the original definition?

(A stubbed `hashScriptIntegrity` together with the conformance scaffolding will likely make the first round of conformance tests pass for the wrong reason.)

src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -388,7 +385,8 @@ module PParamsUpdate where
       ; Emax                        = U.Emax ?↗ P.Emax
       ; nopt                        = U.nopt ?↗ P.nopt
       ; collateralPercentage        = U.collateralPercentage ?↗ P.collateralPercentage
-      ; costmdlsAssoc               = U.costmdls ?↗ P.costmdlsAssoc
+      ; costmdlsAssoc               = if U.costmdls then (λ {cm} → cm ∪ˡᶜᵐ P.costmdlsAssoc)
+                                                    else P.costmdlsAssoc
```
williamdemeo, line 389: This changes the semantics of `ChangePParams` for cost models from "replace the entire `costmdls` map if the update is `Just`" to "merge per-language with the update taking precedence." That's a meaningful protocol-level change and I don't see it called out anywhere.

**Questions**

1. Is this an intentional Dijkstra-era change (matching something in the CIP), or motivated purely by the foreign-extraction refactor of `LanguageCostModels` from a type alias to a record wrapper?
2. If intentional, please add a paragraph to the PParams documentation explaining the new semantics (and maybe note it in the CHANGELOG?)
3. If it's incidental to the wrapper-record refactor, the simplest fix is to keep `?↗` semantics: `costmdlsAssoc = U.costmdls ?↗ P.costmdlsAssoc` should still type-check once `LanguageCostModels` is a record (the wrapper is transparent under `?↗`).

Also: `if U.costmdls then …` reads like boolean-`if` over a `Maybe`, which I assume relies on a `Maybe`→`Bool` coercion somewhere upstream. `maybe (λ cm → cm ∪ˡᶜᵐ P.costmdlsAssoc) P.costmdlsAssoc U.costmdls` would make the `Maybe` handling explicit.

src/Ledger/Dijkstra/Foreign/Chain.lagda.md
```diff
@@ -0,0 +1,113 @@
...
+
+record HSBlock : Type where
+  field
+    bheader   : BHeader
+    ts        : List TopLevelTx
+    bBodySize : ℕ
+    bBodyHash : KeyHash
+
+postulate
+  trustMe-≡ : ∀ {a} {A : Type a} {x y : A} → x ≡ y
```
williamdemeo: The `trustMe-≡` postulate doesn't seem to be used anywhere in the file; `convBlock` and `Conv-Block-HSBlock` both compute proofs via decidable equality.

As written, this postulate is `≡-of-anything-to-anything`, which is inconsistent.  If it's dead code, we should delete it; if it's intended for future use, we should mark it `--FIXME` or `--TODO`, with a reference to what should replace it.

src/Ledger/Dijkstra/Foreign/Account.lagda.md
```diff
@@ -0,0 +1,31 @@
+---
+source_branch: master
+source_path: src/Ledger/Dijkstra/Foreign/Account.lagda.md
+---
+```agda
+module Ledger.Dijkstra.Foreign.Account where
```
williamdemeo, line 6: Places where this module (`Ledger.Dijkstra.Foreign.Account`) is not yet included, but probably should be:

1. `mkdocs.yml` (add a `- Account: Ledger.Dijkstra.Foreign.Account.md` entry under `Foreign/`);
2. `src/Ledger/Dijkstra/Foreign.lagda.md` (add `open import Ledger.Dijkstra.Foreign.Account public`);
3. `MAlonzo/Code/Ledger/Dijkstra/Foreign/API.hs` (re-export `BalanceInterval(..)` so Haskell-side test code can construct `txBalanceIntervals` values).

williamdemeo (approved):
Excellent work, Carlos!

Would you mind adding a PR description summarizing the changes?  In particular it would help to call out some or all of the following:

1. which Specification-side changes are forced by foreign extraction vs. which are independent semantic changes;
2. which rules have step functions wired up and which are deferred;
3. any of the `trustMe`/`postulate` shortcuts that you intend to come back to.

The diff modifies enough non-`Foreign/` files that a "roadmap," or at least a few sentences, up front would make review faster on this and follow-up Issues/PRs.

I know that's a tedious ask, but it would help me understand better, and also make follow-up Issue/PR creation easier to plan and develop.  (Maybe ask Claude or Copilot to draft a PR description addressing the points above...?)

williamdemeo:
I asked Claude to help me review this PR and it caught the issues I mentioned above.

I included in my Claude query the fact that we like well-scoped, manageable PRs, so missing things could be flagged for follow-up issues rather than marking them as deficiencies of the current PR.

Claude suggests five follow-up issues.  I'll look at them more carefully and decide whether they make sense and deserve tickets... in the meantime, here are four of the five comments that Claude suggested I include with the PR review pointing to those issues (the fifth comment was about using `it DummyAbstractFunctions` directly, instead of going through `ExternalFunctions`, and whether that is intentional; since we already discussed this earlier today, I won't rehash it here).

1. Per-rule conformance for the new Dijkstra `SUBLEDGER`/`SUBLEDGERS`/`SUBUTXO`/`SUBUTXOW` rules isn't wired up — the env types are exposed but there are no step functions. End-to-end via `ledgers-step` will still work, so this is fine for this PR.
2. The `trustMe-flowConservation`/`trustMe-nonneg`/`trustMe-nonpos` postulates in `Foreign/Rewards.lagda.md` are conceptually similar to the Conway-side shortcuts but lack a FIXME tag, and `trustMe-nonneg` + `trustMe-nonpos` are jointly inconsistent if instantiated on the same `ℤ`. They're only used in the `from`-direction of `Convertible RewardUpdate HsRewardUpdate`, where we trust that the Haskell-side `RewardUpdate` value really does satisfy the invariants — so functionally they're fine, but they should at minimum carry a FIXME and ideally have a tracking issue. Same for `ppWF` in `ExternalStructures.lagda.md` (which already has a FIXME).
3. Heads up: in moving from Conway's manual `case r of …` to `to (compute …)`, the foreign step functions for EPOCH/NEWEPOCH lose the `(state × String)` proof-summary that Conway's versions returned. If we want to keep that debug context for conformance failure messages — same way Conway has it — we'll need to either keep the manual case-match wrapper or extend `Computational` to expose a show-of-proof. This is fine to defer and file an issue.
4. Minor: this imports `Coin-TokenAlgebra` from `Ledger.Conway.Specification.TokenAlgebra.Coin`, which makes Dijkstra depend on a Conway-namespaced module. Since the Coin instance is era-agnostic, it'd be cleaner to either move it to a shared `Ledger.Core.*` location or define a parallel `Ledger.Dijkstra.Specification.TokenAlgebra.Coin`. Not a blocker; file an issue.

williamdemeo:
Three more minor notes from Claude, which I'll share just in case they're of interest/helpful to you (but you are free to ignore them of course).

- `Foreign/Account.lagda.md`'s `toHaskellConstr` falls through to `else ""` on any unrecognized constructor name, which would silently produce broken Haskell if `BalanceInterval` ever gains a constructor. `else error $ "unknown BalanceInterval constructor: " S.++ s` would fail loudly.
- The removal of the unused `(txLevel : TxLevel)` parameter from `TxWitnesses` is fine — the field types didn't depend on it — but it's a Specification change that'd be worth one line in the PR description.
- The `derive-DecEq` additions for `StakePoolParams`/`DCert`/`Anchor` and the `==-GovProposal`/`==-GovAction` machinery in `Specification/Gov/Actions.lagda.md` are pure infrastructure for foreign deriving. Same: a sentence in the PR description placing them is enough.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
carlostome: 1. It is not a Dijkstra-era change but a retrofit into Dijkstra of a (somewhat) recent change in Conway (see https://github.com/IntersectMBO/formal-ledger-specifications/pull/1107).
2. See (1)
3. The wrapper refactor is to better deal with Haskell instance in conformance testing.

Issue #1119, opened 2026-03-12
[Dijkstra] CIP-159-07: Version gating — restrict CIP-159 fields based on script language version


williamdemeo:
## Description

**Priority**: High (Phase 1)
**Depends on**: Issue #1114

CIP-159 introduces fields (`direct_deposits`, `account_balance_intervals`) that older Plutus scripts (v1–v3) cannot understand. These fields must be gated so that transactions using older scripts don't encounter unknown fields. CIP-118's nested transactions provide the escape hatch: older scripts run in the top-level transaction, while CIP-159 fields are isolated in sub-transactions.

This issue covers the `UsesV4Features` / `UsesV5Features` predicates in `Utxow` and the `allowedLanguages` logic.

### 1. Extend `UsesV4Features` or Define `UsesV5Features`

Currently in `Ledger.Dijkstra.Specification.Utxow`,

```agda
data UsesV4Features : Set where
  hasSubtransactions  : ¬ (Is-[] txSubTransactions) → UsesV4Features
  hasGuards           : ¬ (Is-∅ txGuards)           → UsesV4Features
```

CIP-159 fields should gate to at least PlutusV4, or a new PlutusV5 if the Dijkstra era were to define one (**it likely will not**).

**Design decision**.  Whether CIP-159 features are gated as PlutusV4 or a new PlutusV5 depends on whether the Dijkstra era introduces these features alongside CIP-118 (same Plutus version) or in a later version bump.  (**Answer**: later)
Based on CIP-159's text, it references "PlutusV4" as the boundary, so CIP-159 features require PlutusV4, making them part of `UsesV4Features`.

So, we should extend `UsesV4Features`, as follows:
```agda
data UsesV4Features : Set where
  hasSubtransactions     : ¬ (Is-[] txSubTransactions)       → UsesV4Features
  hasGuards              : ¬ (Is-∅ txGuards)                 → UsesV4Features
  hasDirectDeposits      : ¬ (DirectDepositsOf tx ≡ ∅)       → UsesV4Features  -- NEW
  hasBalanceIntervals    : ¬ (BalanceIntervalsOf tx ≡ ∅)     → UsesV4Features  -- NEW
```

### 2. Update `allowedLanguages`

The `allowedLanguages` function determines which Plutus language versions are permitted in a transaction.  If a transaction uses CIP-159 fields, only PlutusV4 scripts should be allowed at that transaction level.

[...]

Issue #1115, opened 2026-03-12
[Dijkstra] CIP-159-03: Extend `DState` to support account balances and direct deposits


williamdemeo:
## Description

**Depends on**: Issue #1113 (core types)

CIP-159 changes the account state from a simple `Coin` balance (used for staking rewards) to a richer structure (that later will hold multi-asset balances and) that supports direct deposits. In Dijkstra, the balance remains ADA-only, but the state representation must support the new deposit/withdrawal semantics.

Currently in the Dijkstra `Certs` module:

```agda
Rewards : Type
Rewards = Credential ⇀ Coin

record DState : Type where
  constructor ⟦_,_,_,_⟧ᵈ
  field
    voteDelegs   : VoteDelegs
    stakeDelegs  : StakeDelegs
    rewards      : Rewards
    deposits     : Credential ⇀ Coin
```

CIP-159 proposes the following Haskell-level change:

```haskell
data AccountState era = AccountState
  { balance             :: AccountValue era  -- was: Coin
  , deposit             :: Coin              -- registration deposit + whitelist deposit
  , stakePoolDelegation :: StrictMaybe (KeyHash 'StakePool)
  , dRepDelegation      :: StrictMaybe DRep
  }
```

## Changes Required in Dijkstra (ADA-only)

The `rewards` field in `DState` already maps `Credential ⇀ Coin`. The key semantic change is:

1. **Rewards now represent "account balances"**, not just staking rewards. Direct deposits add to this balance. The field name should arguably be generalized (e.g., `accountBalances` or keep `rewards` with updated documentation).

2. **Partial withdrawals** must be supported: the `DELEG-dereg` precondition `(c , 0) ∈ rwds` (requiring zero balance for deregistration) and the withdrawal rule in `PRE-CERT` both need updating. This is addressed in Issue 4 but depends on the state representation chosen here.

3. **Direct deposit application**: Define a function that applies direct deposits to the DState:
   ```agda
   applyDirectDeposits : DirectDeposits → DState → DState
   applyDirectDeposits dd ds = record ds { rewards = rewards ds ∪⁺ dd }
   ```
   This function will be called from the UTXO/LEDGER rules.

## Acceptance Criteria

- [ ] `DState` is updated (or documented) to clarify that `rewards` represents account balances.
- [ ] `applyDirectDeposits` function is defined.
- [ ] `rewardsBalance` is updated if the rewards type changes.
[...]

PR #1147, opened 2026-04-08
[Dijkstra] CIP-159-04: Update cert rules for partial withdrawals (#1116)

### LINKED ISSUE #1116: [Dijkstra] CIP-159-04: Update certificate rules for partial withdrawals
## Description

**Depends on**: Issues #1113, #1115

CIP-159 changes the withdrawal semantics and introduces a new certificate type. This issue covers updating the `DELEG`, `PRE-CERT`, and `CERT` transition rules in `Ledger.Dijkstra.Specification.Certs`.

### Partial Withdrawals

Currently, the `PRE-CERT` rule enforces that withdrawals drain the *full* account balance:

```agda
∙ mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ
```

This premise checks that each `(credential, amount)` pair in `wdrls` is present in `rewards`, which means the withdrawal amount must equal the current reward balance.  CIP-159 changes this to allow **partial withdrawals**: the withdrawal amount must be between 0 and the current balance (inclusive).

1.  **Relax the withdrawal precondition in `PRE-CERT`**.

    ```agda
    -- Old: exact match
    ∙ mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ

    -- New: withdrawal amount ≤ current balance, and credential is registered
    ∙ ∀[ (addr , amt) ∈ wdrls ˢ ] (stake addr ∈ dom rewards × amt ≤ lookupOr 0 (stake addr) rewards)
    ```

2.  **Update the withdrawal effect in `PRE-CERT`**.  Instead of setting withdrawn credentials to 0, subtract the withdrawal amount.

    ```agda
    -- Old: zero out withdrawn credentials
    constMap wdrlCreds 0 ∪ˡ rewards

    -- New: subtract withdrawal amounts
    rewards ∪⁺ (mapKeys stake (mapValues negate wdrls))
    -- or equivalently, for each (c, amt) in wdrls: rewards[stake c] -= amt
    ```

   The precise formulation depends on available map operations. A helper like `subtractFromRewards` may be cleaner.

3.  **Partial withdrawal version restriction**.  CIP-159 states that partial withdrawals are only possible in transactions *without* Plutus v1–v3 scripts, or inside sub-transactions (CIP-118). This version gating is handled in Issue #1119, but the `CERTS` environment may need a flag or the `CertEnv` may need extension.

### Deregistration with Non-Zero Balance

Currently, `DELEG-dereg` requires `(c , 0) ∈ rwds` — the balance must be zero to deregister. With CIP-159, an account can have a non-zero balance from direct deposits. The deregistration rule should:

- Either require the account to have zero balance (forcing a withdrawal first), or
- Automatically include the remaining balance as a refund.

CIP-159 does not explicitly change deregistration rules, so the existing requirement (`(c, 0) ∈ rwds`) is likely preserved — the user must withdraw all funds before deregistering. Confirm this interpretation.

## Acceptance Criteria

- [ ] `PRE-CERT` allows partial withdrawals (amount ≤ balance).
- [ ] `PRE-CERT` correctly subtracts partial withdrawal amounts from rewards.
- [ ] Existing full-withdrawal semantics remain valid (as a special case).
- [ ] The module compiles with `--safe`.

## Design Notes

The partial withdrawal change is semantically significant: it moves from an "all-or-nothing" model to a "withdraw any prefix" model.  This affects the `CERTS-PoV` (preservation of value) proof since the withdrawn amount no longer equals the balance.

## References

- [CIP-159 Partial Withdrawals](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0159#partial-withdrawals-and-native-asset-withdrawals)
- [CIP-159 Whitelist Certificate](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0159#specification) (stake_credential_adjust_whitelist)
- Existing `PRE-CERT` and `DELEG` rules in `src/Ledger/Dijkstra/Specification/Certs.lagda.md`

williamdemeo:
## Description

This PR closes Issue #1116.

**Reviewer Note**.  This PR branches off of #1125 (Issue #1115), so it targets that branch to make reviewing the diffs easier.

This PR updates the `PRE-CERT` rule to support partial withdrawals as specified by
[CIP-159](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0159); specifically, handling of withdrawals changes from an "all-or-nothing" model (the withdrawn amount must equal the full account balance) to a "partial withdrawal" model (any amount up to the current balance may be withdrawn).

### Changes

`src/Ledger/Dijkstra/Specification/Certs.lagda.md`:

+  **New function**.  `applyWithdrawals : Withdrawals → Rewards → Rewards` — fold-based pointwise subtraction; For each `(addr, amt)` in the withdrawal map, looks up `stake addr` in rewards, computes `bal ∸ amt`, and replaces the entry via a singleton `∪ˡ` with complement restriction; credentials not in the withdrawal map are unaffected.

+  **Updated `PRE-CERT` preconditions**.  
   1.  `filter isKeyHash wdrlCreds ⊆ dom voteDelegs` (unchanged).
   2.  `wdrlCreds ⊆ dom rewards` — all withdrawal credentials are registered.
   3.  `∀[ (addr , amt) ∈ wdrls ˢ ] amt ≤ maybe id 0 (lookupᵐ? rewards (stake addr))` — each withdrawal amount ≤ current balance.

       In case `stake addr` is not in the domain of the `rewards` map, then `maybe id 0 (lookupᵐ? rewards (stake addr))` is 0, in which case the predicate fails if `amt` is nonzero and succeeds otherwise.

       Note this means a withdraw of 0 will not cause the predicate to fail, even if the staking credential of `addr` is not in the domain of `rewards`.

+  **Updated `PRE-CERT` effect**.
   +  Old: `constMap wdrlCreds 0 ∪ˡ rewards` — zeroes out withdrawn credentials.
   +  New: `applyWithdrawals wdrls rewards` — subtracts withdrawal amounts pointwise.

+  **New documentation**.  Explains the partial withdrawal semantics, documents that full withdrawals remain valid as a special case (`amt ≡ bal` gives `bal ∸ amt = 0`), and notes the deferred version restriction.

`src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md`:

+  **Updated `Computational-PRE-CERT`**.  Both `computeProof` and `completeness` use the new three-premise `¿ ... ¿` expression.

---

### Design Considerations

+  **Fold-based `applyWithdrawals`**.  Iterates over withdrawal entries, using `lookupᵐ?` + `∸` + singleton `∪ˡ` with complement restriction for each entry; avoids needing a `mapKeys stake wdrls` conversion (which fails because `stake : RewardAddress → Credential` drops the `net` field and isn't injective).

+  **`DELEG-dereg` unchanged**.  The `(c , 0) ∈ rwds` precondition is preserved; CIP-159 does not change deregistration rules; users must withdraw all funds before deregistering, which is now possible via one or more partial withdrawals.

+  **`legacyMode` version restriction deferred**.  CIP-159 specifies partial withdrawals are only permitted without V1–V3 scripts; eEnforcing this requires adding `legacyMode : Bool` to `CertEnv` and threading it through `SUBLEDGER`/`LEDGER`, which requires UTXOW interface changes; deferred to a follow-up issue; #1119 already gates CIP-159 *fields* in legacy mode; the remaining gap (partial withdrawals in legacy mode without CIP-159 fields) is a refinement.

## Acceptance criteria checklist

- [x] `PRE-CERT` allows partial withdrawals (amount ≤ balance)
- [x] `PRE-CERT` correctly subtracts partial withdrawal amounts
- [x] Existing full-withdrawal semantics remain valid (special case)
- [x] Module compiles with `--safe`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### WIP
 
+- Allow partial withdrawals in `PRE-CERT` rule; define `applyWithdrawals` and `_≤ᵐ_` (CIP-159).
 - Extend `TxInfo` with `txDirectDeposits` and `txBalanceIntervals` fields (CIP-159).
 - Remove `allBalanceIntervals` batch-level aggregation helper; balance interval
   assertions will be checked per-(sub)transaction in `UTXO`/`SUBUTXO` rules (see #1117).
--- a/src/Ledger/Dijkstra/Specification/Certs.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Certs.lagda.md
@@ -466,12 +466,9 @@ data _⊢_⇀⦇_,PRE-CERT⦈_ : CertEnv → CertState → ⊤ → CertState →
     in
     ∙ filter isKeyHash wdrlCreds ⊆ dom voteDelegs
     ∙ wdrlCreds ⊆ dom rewards
-    ∙ ∀[ (addr , amt) ∈ wdrls ˢ ] amt ≤ᵐ lookupᵐ? rewards (stake addr)
+    ∙ ∀[ (addr , amt) ∈ wdrls ˢ ] amt ≤ maybe id 0 (lookupᵐ? rewards (stake addr))
       ────────────────────────────────
-      ⟦ e , pp , vs , wdrls , cc ⟧ ⊢
-        ⟦ ⟦ voteDelegs , stakeDelegs , rewards , deposits ⟧ , stᵖ , ⟦ dReps , ccHotKeys , deposits' ⟧ ⟧
-          ⇀⦇ _ ,PRE-CERT⦈
-        ⟦ ⟦ voteDelegs , stakeDelegs , applyWithdrawals wdrls rewards , deposits ⟧ , stᵖ , ⟦ refreshedDReps , ccHotKeys , deposits' ⟧ ⟧
+      ⟦ e , pp , vs , wdrls , cc ⟧ ⊢ ⟦ ⟦ voteDelegs , stakeDelegs , rewards , deposits ⟧ , stᵖ , ⟦ dReps , ccHotKeys , deposits' ⟧ ⟧ ⇀⦇ _ ,PRE-CERT⦈ ⟦ ⟦ voteDelegs , stakeDelegs , applyWithdrawals wdrls rewards , deposits ⟧ , stᵖ , ⟦ refreshedDReps , ccHotKeys , deposits' ⟧ ⟧
 ```
 
 **TODO**: Version restriction (deferred).  CIP-159 specifies that partial withdrawals are
--- a/src/Ledger/Dijkstra/Specification/Certs.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Certs.lagda.md
@@ -293,6 +293,32 @@ applyDirectDeposits dd ds = record ds { rewards = RewardsOf ds ∪⁺ dd }
 The `LEDGER`{.AgdaDatatype} rule will call `applyDirectDeposits`{.AgdaFunction} to
 credit direct deposits to account balances as part of processing a transaction batch.
 
+```agda
+applyWithdrawals : Withdrawals → Rewards → Rewards
+applyWithdrawals wdrls rwds =
+  foldl applyOne rwds (setToList (wdrls ˢ))
+  where
+    -- For each withdrawal entry `(addr, amt)`, look up `stake addr` in the acc,
+    -- compute `bal ∸ amt`, create a singleton map with the new balance, and
+    -- merge it with the rest (complement-restricted, to remove the old entry).
+    applyOne : Rewards → RewardAddress × Coin → Rewards
+    applyOne acc (addr , amt) =
+      case lookupᵐ? acc (stake addr) of λ where
+        (just bal) → ❴ stake addr , bal ∸ amt ❵ ∪ˡ (acc ∣ ❴ stake addr ❵ ᶜ)
+        nothing    → acc
+        -- `nothing` case is defensive; the PRE-CERT precondition guarantees the
+        -- credential is registered, but handling it makes the function total.
+```
+
+In the Dijkstra era, CIP-159 allows **partial withdrawals**: a transaction may
+withdraw any amount up to the current account balance.
+`applyWithdrawals`{.AgdaFunction} subtracts each withdrawal amount from the
+corresponding account balance.  Full withdrawals remain valid as a special case
+(where the withdrawn amount equals the balance, leaving a zero balance).
+
+The `PRE-CERT`{.AgdaDatatype} rule calls `applyWithdrawals`{.AgdaFunction}
+to process withdrawals as part of certificate processing.
+
 <!--
 ```agda
 instance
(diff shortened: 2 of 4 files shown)


src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -421,13 +474,24 @@ data _⊢_⇀⦇_,PRE-CERT⦈_ : CertEnv → CertState → ⊤ → CertState →
         wdrlCreds       = mapˢ stake (dom wdrls)
     in
     ∙ filter isKeyHash wdrlCreds ⊆ dom voteDelegs
-    ∙ mapˢ (map₁ stake) (wdrls ˢ) ⊆ rewards ˢ
+    ∙ wdrlCreds ⊆ dom rewards
+    ∙ ∀[ (addr , amt) ∈ wdrls ˢ ] amt ≤ᵐ lookupᵐ? rewards (stake addr)
```
carlostome: It seems to me there is certain dependency here that is not being exploited, and thus, you need to introduce a new ad-hoc relation `≤ᵐ`.

I think a simpler condition will check both conditions at the "same time", using the information about the presence of a key in the map to directly lookup into a value.

Something along the lines of:

```agda
∀[ (addr , amt) ∈ wdrls ˢ ] (if lookupᵐ? rewards (stake addr) then (\{reward} -> amt ≤ reward) else ⊥)
```

williamdemeo: To me the "dependent relation" `≤ᵐ` is easier to read than the `if-then-else`.  Also, I'll need the same relation when I define the `NoPhantomWithdrawals` function in the PR I'm working on now.  (I suppose you would have the same objection there, ...although, since I'd be using it twice, it's less "ad hoc" :)

I'll leave it in both PRs for now and we can discuss it when we meet again.

williamdemeo: Btw, since I want to use it in two different modules, I'll move it to `Ledger.Prelude`.

williamdemeo: Actually, I thought of a better way to handle this... I'll change this PR to draft while I sort it out.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### WIP
 
+- Allow partial withdrawals in `PRE-CERT` rule; define `applyWithdrawals` and `_≤ᵐ_` (CIP-159).
```
carlostome, line 7: This is outdated

src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -293,6 +293,32 @@ applyDirectDeposits dd ds = record ds { rewards = RewardsOf ds ∪⁺ dd }
 The `LEDGER`{.AgdaDatatype} rule will call `applyDirectDeposits`{.AgdaFunction} to
 credit direct deposits to account balances as part of processing a transaction batch.
```
carlostome, line 294: Why the LEDGER rule?

carlostome (approved):
Nice work, LGTM!

PR #1201, opened 2026-05-18
Add ENTITIES rule

carlostome:
# Description

This PR adds the `ENTITIES` rule which replaces `CERTS` in Dijkstra by inlining PRE/POST-CERT.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Specification/Entities.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Entities.lagda.md
@@ -3,7 +3,7 @@ source_branch: master
 source_path: src/Ledger/Dijkstra/Specification/Entities.lagda.md
 ---
 
-# Certificates {#sec:certificates}
+# Entities {#sec:entities}
 
 <!--
 ```agda
@@ -75,6 +75,12 @@ Premise (1) ensures that each withdrawal targets a registered
 account and that the withdrawal amount does not exceed the account's
 current balance.
 
+The phantom-asset prohibition of CIP-159 — that withdrawals in one
+sub-transaction may not draw from deposits made by an earlier
+sub-transaction in the same batch — is enforced separately in the
+`Utxo`{.AgdaModule} (see [Phantom Asset
+Prevention](Ledger.Dijkstra.Specification.Utxo.md#subsubsec:phantom-asset-prevention)).
+
 ## Direct Deposit Application (CIP-159)
 
 The `ENTITIES`{.AgdaDatatype} rule applies CIP-159 direct deposits to the
@@ -109,8 +115,7 @@ private variable
   voteDelegs₀ voteDelegs₁ : VoteDelegs
   wdrls : Withdrawals
   dd : DirectDeposits
-  A : Type
-  depositsᵍ depositsᵈ₀ depositsᵈ₁ : A ⇀ Coin
+  depositsᵍ depositsᵈ₀ depositsᵈ₁ : Credential ⇀ Coin
 
   dCerts : List DCert
   e : Epoch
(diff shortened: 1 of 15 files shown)


src/Ledger/Dijkstra/Specification/Ledger.lagda.md
```diff
@@ -27,7 +27,8 @@ open import Ledger.Dijkstra.Specification.Enact govStructure
 open import Ledger.Dijkstra.Specification.Gov govStructure
 open import Ledger.Dijkstra.Specification.Utxo txs abs
 open import Ledger.Dijkstra.Specification.Utxow txs abs
-open import Ledger.Dijkstra.Specification.Certs govStructure
+open import Ledger.Dijkstra.Specification.Cert govStructure
```
williamdemeo:
```suggestion
open import Ledger.Dijkstra.Specification.Certs govStructure
```

williamdemeo (changes requested):
Nice work, Carlos!  Overall this is a clean refactoring and I'm happy with the direction.  Folding the pre-/post-CERT effects into a single wrapping rule that calls `CERTS` (now restored to a plain reflexive-transitive closure of `CERT`) is more readable than the previous `RunTraceAfterAndThen` pipeline, and it matches the Haskell-team compromise we agreed on.  I have a few other comments that I'll flag in separate threads.

williamdemeo:
## Naming

I still think `ENTITIES` is uninformative — it doesn't tell the reader what the rule does, and "entities" doesn't have an established meaning in the ledger spec.  A few alternatives, roughly in order of my preference, in case any of them appeal:

+  **`ACCOUNTING`**.  The rule accounts for the reward-account flows (withdrawals out, direct deposits in) and the certificate-driven state updates.  Captures the bookkeeping nature without being generic.
+  **`DELEGS`**.  Read broadly as "the delegation subsystem", which in Cardano traditionally encompasses stake, pool, and (post-Conway) DRep bookkeeping plus reward-account effects.
+  **`STAKE`**.  Short, Cardano-conventional, and accurate at the level of "everything that touches the staking and reward subsystem in a tx".

I don't have strong feelings between these, but I'd push back on `ENTITIES` if there's still room to negotiate with the Haskell team.  Even `WITHDRAWALS-CERTS-DEPOSITS` (clunky but accurate) seems preferable to a name that could equally well describe almost any rule in the spec.

williamdemeo:
## `Entities.lagda.md` documentation

+  **File title and section anchor**.

   The new file opens with

   ```markdown
   # Certificates {#sec:certificates}
   ```

   which collides with the existing `Certs.lagda.md` heading and anchor, and misnames the section.  Suggest e.g.

   ```markdown
   # Entities {#sec:entities}
   ```

   (or whatever the final rule name is).

+  **Heading levels in `Certs.lagda.md` are now inconsistent**.

   `DELEG`, `POOL`, and `GOVCERT` are introduced as `##` subsections, but `CERT` and `CERTS` are `#` top-level headings.  I'd promote all five to the same level (probably `###`) under a single `## Transition Systems` header.

+  **Lost prose from the old `Certs.lagda.md`**.

   A few useful paragraphs disappeared with the deletion of the `PRE-CERT` and `POST-CERT` sections and weren't carried over to `Entities.lagda.md`.  (Caveat: I did a rebase onto master before reviewing, so I might have accidentally deleted some of these allegedly missing paragraphs.)  If you prefer more concise, trimmed down versions of any of these, or if you have specific reasons for deleting them, that's perfectly fine.

   +  The `TODO` note about CIP-159's version restriction (`legacyMode ≡ false` ⇒ partial withdrawals only, threading required through `CertEnv`, `SUBLEDGER`, and `LEDGER`).  That follow-up is still planned, so the note probably belongs somewhere in the new file.
   +  The detailed paragraph explaining the relationship between `applyDirectDeposits`, the `dom dd ⊆ dom rewards` premise, and CIP-159's phantom-asset prohibition (which is enforced separately in `Utxo` via `NoPhantomWithdrawals`).  This is genuinely important context for the rule and I think it should be preserved.
   +  The paragraph clarifying that the deposit-domain check is performed against the *post-`CERT` of the same transaction*, so earlier-sub-transaction deregistrations are handled correctly.

+  **Premise numbering in the prose**.

   The prose refers to "Premise (1)" and "Premise (2)" but the rule has four numbered premises (or three withdrawal premises plus one direct-deposit premise, depending on how you count). Probably we should update the references and/or the premises explicitly numbered with `-- (1)` … `-- (4)` comments next to each.

src/Ledger/Dijkstra/Specification/Entities.lagda.md
```diff
@@ -0,0 +1,144 @@
...
+        refreshedDReps  = mapValueRestricted (const (e + pp .drepActivity)) dReps refresh
+        wdrlCreds       = mapˢ stake (dom wdrls)
+        ddCreds         = mapˢ stake (dom dd)
+        activeVDelegs   = mapˢ vDelegCredential (dom (DRepsOf gState₁))
+                               ∪ fromList (vDelegNoConfidence ∷ vDelegAbstain ∷ [])
+    in
+    ∙ filter isKeyHash wdrlCreds ⊆ dom voteDelegs₀
+    ∙ wdrlCreds ⊆ dom rewards₀
+    ∙ ∀[ (addr , amt) ∈ wdrls ˢ ] amt ≤ maybe id 0 (lookupᵐ? rewards₀ (stake addr)) -- (1)
+    ∙ ddCreds ⊆ dom rewards₁ -- (2)
```
williamdemeo: (We already discussed this, and you mentioned you were planning to reorder... but I'll make a note anyway, just for the record.)  The fourth premise `ddCreds ⊆ dom rewards₁` is stated before the `CERTS` premise that introduces `rewards₁`.  That makes computational instances easier to prove and is fine logically (premises are conjuncts which commute), but (as you mentioned) it might read more naturally if the `CERTS` step comes first and the deposit-domain check follows, since the latter constrains a variable bound by the former.

src/Ledger/Dijkstra/Specification/Entities.lagda.md
```diff
@@ -0,0 +1,144 @@
...
+private variable
+  rewards₀ rewards₁ : Rewards
+  dReps : DReps
+  stakeDelegs stakeDelegs₀ stakeDelegs₁ : StakeDelegs
+  ccHotKeys : CCHotKeys
+  voteDelegs₀ voteDelegs₁ : VoteDelegs
+  wdrls : Withdrawals
+  dd : DirectDeposits
+  A : Type
+  depositsᵍ depositsᵈ₀ depositsᵈ₁ : A ⇀ Coin
```
williamdemeo: Both `DState.deposits` and `GState.deposits` are `Credential ⇀ Coin`, so it seems the abstraction over `A` doesn't buy us anything and slightly obscures the rule.  Maybe clearer to just do

```agda
depositsᵍ depositsᵈ₀ depositsᵈ₁ : Credential ⇀ Coin
```

williamdemeo:
Finally, `Foreign/Cert.lagda.md` still exports `cert-step` and `certs-step` (which continue to make sense) but there's nothing exporting an `entities-step`.  If we need to call the new rule from the Haskell side via the FFI, I suppose we need a new binding.  Is it worth creating a ticket for this as a follow-up?

williamdemeo (approved):
Great work, Carlos!  LGTM!

PR #1217, opened 2026-06-09
Gov property #414: state that GA deposits are eventually refunded

### LINKED ISSUE #414: Property: GA deposits are eventually refunded
For all proposal deposits `d` currently in a block, there exists an epoch `e` where if it contains a block, the deposit for `d` is not in the deposit pot anymore.

The proof goes something like this:  Every proposal deposit has a corresponding `GovActionState`, which contains an `expiresIn` field. We pick that value for `e`. Since there was a block in `e`, we did a step with `EPOCH` into `e`, and if the GA was still in the state before `EPOCH`, it will be removed by it.

williamdemeo:
## Description

Closes #414. States and proves the governance property "GA deposits are eventually refunded."

+ `Ledger.Conway.Specification.Chain` gains **`CHAINS`**, the reflexive-transitive closure of `CHAIN` along a list of blocks, via the existing `ReflexiveTransitiveClosure` combinator (mirroring `LEDGERS`).

+ `Interface.STS` gains a generic trace-invariant combinator: **`RTC-All P tr`** asserts `P` at the target state of every step of a trace; **`RTC-All-last`** reads `P` off at the final state of a nonempty trace.

+ New module `Ledger.Conway.Specification.Chain.Properties.EventuallyRefunded`: `gaDepositInPot`, the per-state refund-deadline invariant `InvariantAt`, the statement `GADepositsEventuallyRefunded`, and its proof.

---

### Theorem

Given

1. `govDepsMatch (LStateOf cs)` (a known CHAIN invariant), 
2. a `GovState` membership witness for `(gaid , gaSt)`,
3. `LastEpochOf cs ≤ expiresIn gaSt` (not yet expired),
4. a chain extension `chain : _ ⊢ cs ⇀⦇ bs ,CHAINS⦈ cs'` satisfying `RTC-All (InvariantAt gaid gaSt) chain`, and
5. the cutoff `sucᵉ (sucᵉ (expiresIn gaSt)) ≤ LastEpochOf cs'`,

the deposit is gone: `¬ gaDepositInPot cs' gaid`.

The cutoff is two epochs past expiry because of the `RATIFY`/`GovernanceUpdate` pipeline: `RATIFIES` marks the expired action for removal at `sucᵉ (expiresIn)`, and `GovernanceUpdate` filters it out and refunds the deposit at the following boundary.

Precondition 4 holds at every reachable chain state but is presently an assumption; deriving it from the CHAIN rule requires infrastructure not yet in the specification (the `rrm` premise of `CHAIN-govDepsMatch`, tracking the action through `LEDGERS`/`RATIFIES`, abstract `TxId` freshness, and a `sucᵉ`-is-least `EpochStructure` axiom).  (Tracked in Issue #1260, which catalogs the partial machinery in this PR's history.)

---

### Proof structure

The trace invariant constrains every state reached along the extension; in particular, the final one (`RTC-All-last`).  The proof reads the invariant off at the final state and compares epochs: for the empty trace, `cs' = cs` and precondition 3 contradicts the cutoff; otherwise the invariant at `cs'` either yields the refund outright or bounds `LastEpochOf cs'` below the cutoff, a contradiction.

---

### Extra Notes for Reviewer

`CHAINS` now appears in the rendered spec document (Chain chapter) even though the spec itself doesn't use it yet.  (That's what Carlos asked for and it matches the `LEDGERS` precedent; but if we change our mind, the fallback is defining it in the Properties module.)

---

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
+++ b/src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
@@ -17,28 +17,16 @@ module Ledger.Conway.Specification.Chain.Properties.EventuallyRefunded
   (abs : AbstractFunctions txs)
   where
 
-open import Data.List.Relation.Unary.All using (All; []; _∷_)
-open import Data.List.Membership.Propositional.Properties using (∈-filter⁻)
-open import Data.List.Relation.Unary.Any using (here; there)
-
-open import Ledger.Prelude hiding (All)
+open import Ledger.Prelude
 
 open import Ledger.Conway.Specification.Certs govStructure
 open import Ledger.Conway.Specification.Chain txs abs
-open import Ledger.Conway.Specification.Enact govStructure
 open import Ledger.Conway.Specification.Epoch txs abs
-open import Ledger.Conway.Specification.Gov govStructure using (GovState; GovStateOf)
+open import Ledger.Conway.Specification.Gov govStructure using (GovStateOf)
 open import Ledger.Conway.Specification.Ledger txs abs
 open import Ledger.Conway.Specification.Ledger.Properties.Base txs abs
-open import Ledger.Conway.Specification.PoolReap txs abs using (POOLREAP)
-open import Ledger.Conway.Specification.Ratify govStructure
-open import Ledger.Conway.Specification.RewardUpdate txs abs using (TICK)
-open import Ledger.Conway.Specification.Utxo txs abs
 
-open GovActionState  using (votes; expiresIn)
-open Block           using (ts)
-open RatifyEnv       using (currentEpoch)
-open RatifyState     using (removed)
+open GovActionState using (expiresIn)
 ```
 -->
 
@@ -67,9 +55,12 @@ pipeline: at epoch `sucᵉ`{.AgdaFunction} (`expiresIn`{.AgdaField}
 `gaSt`{.AgdaBound}), `RATIFIES`{.AgdaDatatype} marks the expired action for removal
 (adding it to `removed`{.AgdaField}); at the *following* epoch, the
 `GovernanceUpdate`{.AgdaModule} actually filters the action out of the governance state
-and strips its deposit from the deposit pot.
+and strips its deposit from the deposit pot.  The deposit may therefore still be held
+at any state whose last epoch is at most `sucᵉ`{.AgdaFunction}
+(`expiresIn`{.AgdaField} `gaSt`{.AgdaBound}) — the *refund deadline* — and is gone at
+every reachable state beyond it.
 
-Six preconditions are needed.
+Four preconditions are needed.
 
 +  `govDepsMatch`{.AgdaFunction} ensures that every
    `GovActionDeposit`{.AgdaInductiveConstructor} entry has a corresponding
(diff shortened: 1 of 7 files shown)


williamdemeo:
Machine-checked this branch with Agda 2.8.0 (via the project's Nix flake): `EventuallyRefunded.lagda.md` and the `Ledger.Conway.Specification.Properties` index both typecheck under `--safe` with no errors or warnings. The statement stands as a `Claim` (`CHAINStar`, `Σ[ e ∈ Epoch ]`, and `e ≤ LastEpochOf cs'` all elaborate as written) — no proof yet.

Pushed two small refinements:

- Switched the pending marker from `*Proof*. (TODO)` to `*Proof*. (coming soon)`. That string is the machine-readable "proof pending" signal the property-tracking scanner keys on; with `(TODO)` the claim would later be misread as *proved*.
- Fixed the roadmap cross-link: added the `[CHAIN-EventuallyRefunded]` / `[Chain.Properties.EventuallyRefunded]` link definitions in `includes/links.md` and rendered the claim as a list item, matching its siblings.

Statement only; proof to follow. Part of #414.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+gaDepositInPot : ChainState → GovActionID → Type
+gaDepositInPot cs gaid = GovActionDeposit gaid ∈ dom (DepositsOf cs)
+
+-- Reflexive-transitive closure of CHAIN along a list of blocks.
+data CHAINStar : ChainState → List Block → ChainState → Type where
+  []*  : {cs : ChainState} → CHAINStar cs [] cs
+  _∷*_ : {cs cs' cs'' : ChainState} {b : Block} {bs : List Block}
+         → _ ⊢ cs ⇀⦇ b ,CHAIN⦈ cs'
+         → CHAINStar cs' bs cs''
+         → CHAINStar cs (b ∷ bs) cs''
```
carlostome: Why not use one of the `ReflexiveTransitiveClosure` combinators we have?
In addition, for consistency this should be called `CHAINS` and live in the `Chain` Agda module.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+-- The invariant, imposed at every state reached along a CHAINStar derivation.
+CHAINStar-Invariant :
+  {cs   : ChainState}
+  {bs   : List Block}
+  {cs'  : ChainState}
+  → GovActionID → GovActionState → CHAINStar cs bs cs' → Type
+
+CHAINStar-Invariant _ _ []* = ⊤
+CHAINStar-Invariant gaid gaSt (_∷*_ {cs' = cs₁} _ rest) =
+  InvariantAt cs₁ gaid gaSt × CHAINStar-Invariant gaid gaSt rest
```
carlostome: It feels this should be abstracted into a combinator that basically asserts a predicate over the reflexive transitive closure of a relation.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+  {gaSt  : GovActionState}
+  → govDepsMatch (LStateOf cs)
+  → (gaid , gaSt) ∈ fromList (GovStateOf (LStateOf cs))
+  → LastEpochOf cs ≤ expiresIn gaSt
+  →  {bs : List Block}
+     {cs' : ChainState}
+     → (chain : CHAINStar cs bs cs')
+     → CHAINStar-Invariant gaid gaSt chain
+     → sucᵉ (sucᵉ (expiresIn gaSt)) ≤ LastEpochOf cs'
+     → ¬ gaDepositInPot cs' gaid
```
carlostome: I don't understand the purpose of `REFUNDED` relation other than to restate the property being proved by `GADepositsEventuallyRefunded'`. Is also a bit confusing that it uses the STS notation that we use for ledger rules.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+  []*  : {cs : ChainState} → CHAINStar cs [] cs
+  _∷*_ : {cs cs' cs'' : ChainState} {b : Block} {bs : List Block}
+         → _ ⊢ cs ⇀⦇ b ,CHAIN⦈ cs'
+         → CHAINStar cs' bs cs''
+         → CHAINStar cs (b ∷ bs) cs''
+
+-- Either the deposit has already been refunded, or the refund deadline —
+-- one epoch past expiry, owing to the ratification pipeline — has not yet
+-- passed.  This holds at every reachable chain state (see prose above);
+-- deriving it formally from the CHAIN rule is future work.
```
carlostome: I'd suggest opening an issue or this will be forgotten.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
```diff
@@ -0,0 +1,279 @@
...
+Preconditions 1–3 set the scene: they say that `gaSt`{.AgdaBound} is the state of a
+governance action that is live and correctly accounted for at `cs`{.AgdaBound}.  Of
+these, only precondition 3 is consumed by the present proof (in the empty-chain
+case); together they are the intended starting point for the future derivation of
+precondition 4, which is where the remaining "engineering debt" lives: deriving the
+invariant from the `CHAIN`{.AgdaDatatype} rule requires infrastructure not yet
+present in the specification (the `rrm` premise of
+`CHAIN-govDepsMatch`{.AgdaFunction}, preservation of the action through
+`LEDGERS`{.AgdaDatatype} and the `RATIFIES`{.AgdaDatatype} pipeline, and abstract
+`TxId`{.AgdaDatatype} freshness).
```
carlostome: I don't think this paragraph adds much here. The proof is straightforward enough that 1-3 can be understood from the code. The stuff about 4 is already said in the comment above.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
williamdemeo: Good call.  There was no reason for the new datatype. `CHAINS` is now defined in `Ledger.Conway.Specification.Chain` as `ReflexiveTransitiveClosure {sts = _⊢_⇀⦇_,CHAIN⦈_}` and the property is stated over `_ ⊢ cs ⇀⦇ bs ,CHAINS⦈ cs'`.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
williamdemeo: Excellent idea.  I added a generic combinator to `Interface.STS`, next to the existing `RTC-preserves-inv`.

The new `RTC-All P tr` asserts `P` at the target state of every step of a trace, and `RTC-All-last` reads `P` off at the final state of a nonempty trace. `CHAINStar-Invariant` and the ad-hoc `lastInvariantAt` lemma are gone; the premise is now `RTC-All (InvariantAt gaid gaSt) chain`.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
williamdemeo: Fair point.  The intent was to bundle the premises in a familiar style, so a reader who's used to the ledger rules could quickly see/understand what's assumed and asserted... but you're right that it just restated the property one level removed, and borrowing the STS notation for something that isn't a transition rule could be a bit confusing.

I removed `_⊢_⇀⦇_,REFUNDED⦈_` (and the dependent `GADepositsEventuallyRefunded h` form); the claim is now a single top-level `GADepositsEventuallyRefunded : Type` with proof `gaDepositsEventuallyRefunded`.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
williamdemeo: Done.  Opened Issue #1260 to track deriving the invariant from the `CHAIN` rule.  I'll also add a reminder that some of the machinery needed already exists but was removed in commit 7671dafbf (`RATIFY` lemmas, deposit-absence preservation, `TxId` freshness, the `sucᵉ-is-least` axiom), with commit pointers, so none of it has to be rediscovered. The module comment and precondition table now reference the issue.

src/Ledger/Conway/Specification/Chain/Properties/EventuallyRefunded.lagda.md
williamdemeo: Agreed and removed.  I kept the summary table itself since it records each precondition's status at a glance (and now carries the follow-up issue reference for precondition 4), but I don't mind dropping that too if you think it's redundant.

carlostome (approved):
LGTM!

PR #1250, opened 2026-07-06
[Dijkstra] Explicit deposits in certificate deregistration

carlostome:
# Description

This PR makes the optional field in the deregistration certificate mandatory.

In addition, it changes the calculation of new and refunded deposits from certificates.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Specification/Certs.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Certs.lagda.md
@@ -63,7 +63,7 @@ StakeDelegs = Credential ⇀ KeyHash
 
 data DCert : Type where
   delegate    : Credential → Maybe VDeleg → Maybe KeyHash → Coin → DCert
-  dereg       : Credential → Maybe Coin → DCert
+  dereg       : Credential → Coin → DCert
   regpool     : KeyHash → StakePoolParams → DCert
   retirepool  : KeyHash → Epoch → DCert
   regdrep     : Credential → Coin → Anchor → DCert
@@ -401,9 +401,8 @@ data _⊢_⇀⦇_,DELEG⦈_ : DelegEnv → DState → DCert → DState → Type 
   DELEG-dereg :
     ∙ (c , 0) ∈ rwds
     ∙ (c , d) ∈ deposits
-    ∙ md ≡ nothing ⊎ md ≡ just d
       ────────────────────────────────
-      ⟦ pp , pools , delegatees ⟧ ⊢ ⟦ vDelegs , sDelegs , rwds , deposits ⟧ ⇀⦇ dereg c md ,DELEG⦈ ⟦ vDelegs ∣ ❴ c ❵ ᶜ , sDelegs ∣ ❴ c ❵ ᶜ , rwds ∣ ❴ c ❵ ᶜ , deposits ∣ ❴ c ❵ ᶜ ⟧
+      ⟦ pp , pools , delegatees ⟧ ⊢ ⟦ vDelegs , sDelegs , rwds , deposits ⟧ ⇀⦇ dereg c d ,DELEG⦈ ⟦ vDelegs ∣ ❴ c ❵ ᶜ , sDelegs ∣ ❴ c ❵ ᶜ , rwds ∣ ❴ c ❵ ᶜ , deposits ∣ ❴ c ❵ ᶜ ⟧
 
 
 isPoolRegistered : Pools -> KeyHash -> Maybe StakePoolParams
--- a/src/Ledger/Dijkstra/Specification/Certs.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Certs.lagda.md
@@ -303,70 +303,32 @@ rewardsBalance ds = ∑[ x ← RewardsOf ds ] x
 
 Functions in this section compute the effect that a `DCert`{.AgdaRecord} list has on
 the three deposit fields (`DState.deposits`{.AgdaField}, `PState.deposits`{.AgdaField},
-`GState.deposits`{.AgdaField}) carried by a `CertState`{.AgdaRecord}.  They describe
-the deposit evolution for a single `DCert` and mirror that of the corresponding
-`CERT` sub-rule.
+`GState.deposits`{.AgdaField}) carried by a `CertState`{.AgdaRecord}.
 
-### Helper Functions
-
-The three deposit fields carried by a `CertState`{.AgdaRecord} have types
-`Credential ⇀ Coin`, `KeyHash ⇀ Coin`, and `Credential ⇀ Coin`, respectively,
-and we package them up as a triple of values which are computed via
-`coinFromDepositTriple`{.AgdaFunction}.  By definition,
-
-    coinFromDeposits cs ≡ coinFromDepositTriple (depositTripleOf cs).
+In Dijkstra, both registration and deregistration certificates carry
+the explicit deposit. Hence, it is sufficient to use the list of
+certificates in a transaction to compute the amount from new and
+refunded deposits.
 
 ```agda
-DepositTriple : Type
-DepositTriple = (Credential ⇀ Coin) × (KeyHash ⇀ Coin) × (Credential ⇀ Coin)
-
-depositTripleOf : CertState → DepositTriple
-depositTripleOf cs = DepositsOf (DStateOf cs) , DepositsOf (PStateOf cs) , DepositsOf (GStateOf cs)
-
-coinFromDepositTriple : DepositTriple → Coin
-coinFromDepositTriple (dd , dp , dg) = getCoin dd + getCoin dp + getCoin dg
-
 module _ (pp : PParams) where
 
-  updateCertDeposit : DCert → DepositTriple → DepositTriple
-  updateCertDeposit cert (dd , dp , dg) = case cert of λ where
-    (delegate c _ _ d)  → (dd ∪⁺ ❴ c , d ❵  , dp                              , dg)
-    (dereg c _)         → (dd ∣ ❴ c ❵ ᶜ     , dp                              , dg)
-    (regpool kh _)      → (dd               , dp ∪ˡ ❴ kh , pp .poolDeposit ❵  , dg)
-    (regdrep c d _)     → (dd               , dp                              , dg ∪⁺ ❴ c , d ❵)
-    (deregdrep c _)     → (dd               , dp                              , dg ∣ ❴ c ❵ ᶜ)
-    _                   → (dd               , dp                              , dg)
-
-  updateCertDepositsStep : CertState → DCert → CertState
-  updateCertDepositsStep cs c =
-    let (dd , dp , dg) = updateCertDeposit c (depositTripleOf cs) in
-    ⟦ record dState { deposits = dd } , record pState { deposits = dp } , record gState { deposits = dg } ⟧
-    where open CertState cs
-```
-
-Note that any drift between the `updateCertDepositsStep`{.AgdaFunction} and the
-`CERT` sub-rule semantics is a soundness problem: it would allow the UTXO
-batch-balance equation to accept transactions whose actual `CertState` evolution
-doesn't balance.
-
-```agda
-coinFromDeposits : CertState → Coin
-coinFromDeposits cs = coinFromDepositTriple (depositTripleOf cs)
-
-module _ (pp : PParams) (certState : CertState) where
-
-  updateCertDeposits : List DCert → CertState
-  updateCertDeposits = foldl (updateCertDepositsStep pp) certState
-
-  depositsChange : List DCert → ℤ
-  depositsChange certs =
-    coinFromDeposits (updateCertDeposits certs) - coinFromDeposits certState
-
   newCertDeposits : List DCert → Coin
-  newCertDeposits certs = posPart (depositsChange certs)
+  newCertDeposits = foldl addNewCertDeposit 0
+    where
+      addNewCertDeposit : DCert → Coin → Coin
+      addNewCertDeposit (delegate _ _ _ d) acc = acc + d
+      addNewCertDeposit (regpool _ _)      acc = acc + pp .poolDeposit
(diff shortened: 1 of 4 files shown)


src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -303,70 +303,32 @@ rewardsBalance ds = ∑[ x ← RewardsOf ds ] x
...
-  depositsChange certs =
-    coinFromDeposits (updateCertDeposits certs) - coinFromDeposits certState
-
   newCertDeposits : List DCert → Coin
-  newCertDeposits certs = posPart (depositsChange certs)
+  newCertDeposits = foldr addNewCertDeposit 0
+    where
+      addNewCertDeposit : DCert → Coin → Coin
+      addNewCertDeposit (delegate _ _ _ d) acc = acc + d
+      addNewCertDeposit (regpool _ _)      acc = acc + pp .poolDeposit
```
williamdemeo: This charges `poolDeposit` for *every* `regpool` cert, but `POOL-rereg` leaves `deposits` unchanged when the pool is already registered — only `POOL-reg` takes a deposit. The old simulation-based accounting used `∪ˡ` starting from the actual `CertState`, so re-registration contributed 0 to `depositsChange`; this fold has no state to consult and can't distinguish the two cases.

Concretely: a transaction that re-registers an existing pool (e.g. to update pool params) must consume an extra `poolDeposit` to satisfy `consumedBatch ≡ producedBatch`, but no ledger bucket receives that value — deposits, rewards, fees, and outputs are all unchanged by it. Value vanishes, so preservation of value breaks. The same drift occurs within a single transaction containing two `regpool` certs for the same key: the first is a reg, the second a rereg, but the fold charges twice.

A fix consistent with this PR's own design: give `regpool` an explicit deposit field like `delegate`/`regdrep` carry — `regpool : KeyHash → StakePoolParams → Coin → DCert` — with `POOL-reg` requiring `d ≡ pp .poolDeposit`, `POOL-rereg` requiring `d ≡ 0`, and this fold summing the carried `d`. Then the rules police the claimed amount and the accounting is correct by construction, as it already is for the other four deposit-bearing certs (`delegate`/`regdrep` grow the pot by exactly the cert's `d` via `∪⁺` plus the `d ≡ …` premises; `dereg`/`deregdrep` shrink it by exactly `d` via `(c , d) ∈ deposits`). `regpool` is the only drift.

src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -303,70 +303,32 @@ rewardsBalance ds = ∑[ x ← RewardsOf ds ] x
...
-The three deposit fields carried by a `CertState`{.AgdaRecord} have types
-`Credential ⇀ Coin`, `KeyHash ⇀ Coin`, and `Credential ⇀ Coin`, respectively,
-and we package them up as a triple of values which are computed via
-`coinFromDepositTriple`{.AgdaFunction}.  By definition,
-
-    coinFromDeposits cs ≡ coinFromDepositTriple (depositTripleOf cs).
+In Dijkstra, both registration and deregistration certificates carry
+the explicit deposit. Hence, it is sufficient to use the list of
+certificates in a transaction to compute the amount from new and
+refunded deposits.
```
williamdemeo: This justification doesn't hold for `regpool`, which carries no explicit deposit — the fold pulls it from `PParams`, and that's exactly where the accounting drifts from the rules (see the comment on `addNewCertDeposit`). Worth rewording once the `regpool` question is settled.

src/Ledger/Dijkstra/Specification/Ledger.lagda.md
```diff
@@ -188,7 +188,7 @@ instance
   HasCoin-LedgerState : HasCoin LedgerState
   HasCoin-LedgerState .getCoin s =  getCoin (UTxOStateOf s)
                                     + rewardsBalance (DStateOf (CertStateOf s))
-                                    + coinFromDeposits (CertStateOf s)
+                                    + getCoin (DepositsOf (DStateOf s)) + getCoin (DepositsOf (PStateOf s)) + getCoin (DepositsOf (GStateOf s))
```
williamdemeo, line 191: Nit: the prose at line 179 above still says the deposit fields are summed "via coinFromDeposits", which no longer exists. Also, consider keeping a small helper (e.g. `depositsBalance : CertState → Coin`) instead of the inline triple sum — that would also make the access style uniform (the previous line goes through `CertStateOf s`, this one uses `DStateOf s` directly).

src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md
```diff
@@ -63,23 +60,10 @@ instance
...
-      | (no ¬s) = ⊥-elim (¬s (_ , q , refl))
-  Computational-DELEG .completeness _ ds (dereg c (just d)) _ (DELEG-dereg h@(p , q , inj₂ refl))
-    with lookupᵐ?? (DepositsOf ds) c
-  ... | (yes ((_ , d') , q' , refl)) rewrite dec-yes
-          (¿ (c , 0) ∈ (RewardsOf ds)
-           × (c , d') ∈ (DepositsOf ds)
-           × (just d ≡ nothing {A = ℕ} ⊎ just d ≡ just d')
-           ¿) (p , q' , inj₂ (cong just (proj₂ (DepositsOf ds) q q'))) .proj₂ = refl
-  ... | (no ¬s) = ⊥-elim (¬s (_ , q , refl))
+  Computational-DELEG .completeness _ ds (dereg c d) _ (DELEG-dereg h@(p , q))
```
williamdemeo: Nits: the `h@` as-pattern is now unused; `yes p` shadows the `p` from the pattern (could be `yes _`); and the `dereg` case in `computeProof` above kept the deep indentation from the removed `lookupᵐ??` nesting — could be reflowed to match the `delegate` case.

williamdemeo (changes requested):
Nice simplification overall — dropping the `updateCertDeposits` simulation in favour of rule-checked explicit amounts is the right direction, and the new `Computational-DELEG` proofs are much cleaner. One real issue, though: `newCertDeposits` drifts from `POOL-rereg`, which breaks preservation of value for pool re-registration (details inline).

A few points that don't attach to diff lines:

- **Certs.lagda.md line 277**: the private variable `md : Maybe Coin` is now unused. (Slightly related: I had Claude write a Python script that detects unused *imports*; it works really well on `agda-algebras` because in that library I almost always specify exactly what is being imported using `using` and `renaming`, but it might work pretty well on the fls too... I'll give it a try.  If it works, maybe we could add an automatic "clean up imports" step to our workflow.)
- Follow-up thought (fine as a separate PR): a lemma of the shape `Γ ⊢ cs ⇀⦇ certs ,CERTS⦈ cs' → depositsBalance cs' + refundCertDeposits pp certs ≡ depositsBalance cs + newCertDeposits pp certs` would tie the accounting to the rules mechanically — it's exactly what would have caught the `regpool` drift, and would replace the warning comment the old code carried.

williamdemeo:
Claude generated most of the comments above, but I've read/revised them and they seem reasonable/correct to me... but let me know if you think I should dial back the AI when reviewing your PRs.

(tbh, I'm not sure I would have caught the accounting error by myself unless/until I tried to prove a property about it.)

src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -303,70 +305,35 @@ rewardsBalance ds = ∑[ x ← RewardsOf ds ] x
...
```
williamdemeo: Perhaps you plan to change whether `regpool` carries an explicit deposit...?  As it stands, this paragraph seems misleading at best.  `regpool` is the certificate that does not carry a deposit, hence why `newCertDeposits` needs the extra `ℙ KeyHash` argument.

```suggestion
In Dijkstra, delegation and `DRep` (de)registration certificates carry their deposit explicitly, so new and refunded deposits can be computed from the certificate list alone.  The exception is pool registration: `regpool`{.AgdaInductiveConstructor} carries no deposit, and whether it charges `poolDeposit`{.AgdaField} depends on whether the pool is already registered. `newCertDeposits` therefore additionally takes the set of registered pool keys (`pools : ℙ KeyHash`), and threads it through the certificate list, charging each newly registered pool exactly once.
```

src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -303,70 +305,35 @@ rewardsBalance ds = ∑[ x ← RewardsOf ds ] x
...
-  updateCertDeposits = foldl (updateCertDepositsStep pp) certState
-
-  depositsChange : List DCert → ℤ
-  depositsChange certs =
-    coinFromDeposits (updateCertDeposits certs) - coinFromDeposits certState
-
-  newCertDeposits : List DCert → Coin
-  newCertDeposits certs = posPart (depositsChange certs)
+  newCertDeposits : ℙ KeyHash → List DCert → Coin
+  newCertDeposits pools = proj₁ ∘ foldr addNewCertDeposit (0 , pools)
```
williamdemeo: Why `foldr`?  `CERTS` processes certificates head-first, so `foldr` threads the pool set through the list in reverse order relative to the `CERTS` rule.

As of now, `foldr` and `foldl` probably yield the same result, because nothing in a certificate list removes a pool from the registered set... but maybe a certificate that immediately unregisters a pool could break this.

In `CHANGELOG` there's a WIP entry "Fix updateCertDeposits: use foldl (CERTS is head-first)."  I think the same logic applies here.  A `foldl` will also make the PoV proof line up directly with the `CERTS` induction, whereas `foldr` would probably need a fold-reversal lemma.

src/Ledger/Dijkstra/Specification/Certs.lagda.md
```diff
@@ -79,6 +79,9 @@ cwitness (regdrep c _ _)     = just c
 cwitness (deregdrep c _)     = just c
 cwitness (ccreghot c _)      = just c
 
+IsPoolRegistered : Pools -> KeyHash -> Type
+IsPoolRegistered ps kh = Is-just (lookupᵐ? ps kh)
```
williamdemeo:
```suggestion
IsPoolRegistered : Pools → KeyHash → Type
IsPoolRegistered ps kh = kh ∈ dom ps
```
It's equivalent and still decidable, but it matches the membership test used in `newCertDeposits` (`kh ∈ pools` over `dom pools`), so a future lemma relating the fold to the `POOL` rules won't need an `Is-just`/`∈ dom` conversion.

src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md
williamdemeo, line 72: Why not use

```agda
    with ¿ IsPoolRegistered (PoolsOf stᵖ) c ¿
```

in place of

```agda
    with ¿ Is-just (lookupᵐ? (PoolsOf stᵖ) c) ¿
```

...isn't that what it's there for? :)  (applies to three lines)

williamdemeo:
```suggestion

```
(unused)

williamdemeo:
```suggestion
```
(unused)

williamdemeo:
```suggestion

```
(unused)

williamdemeo:
```suggestion
```
(unused)

src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md
```diff
@@ -63,34 +60,21 @@ instance
...
-    with lookupᵐ?? (DepositsOf ds) c
-  ... | (yes ((_ , d') , q' , refl)) rewrite dec-yes
-          (¿ (c , 0) ∈ (RewardsOf ds)
-           × (c , d') ∈ (DepositsOf ds)
-           × (just d ≡ nothing {A = ℕ} ⊎ just d ≡ just d')
-           ¿) (p , q' , inj₂ (cong just (proj₂ (DepositsOf ds) q q'))) .proj₂ = refl
-  ... | (no ¬s) = ⊥-elim (¬s (_ , q , refl))
+  Computational-DELEG .completeness _ ds (dereg c d) _ (DELEG-dereg (p , q))
+    with ¿ (c , 0) ∈ (RewardsOf ds) × (c , d) ∈ (DepositsOf ds) ¿
+  ... | yes p = refl
```
williamdemeo, line 54:
```suggestion
  ... | yes _ = refl
```

Minor nit (avoids appearance of shadowing the other `p`)

williamdemeo (approved):
Nice work, Carlos!  Just a few minor comments/suggestions.  Otherwise, looks great!

### Clear wins

+  all issues from previous review are addressed.
+  the stale `coinFromDeposits` prose is reworded;
+  the unused `md` variable is gone;
+  the `dereg` Computational proof is much simpler (the `lookupᵐ??` case analysis is no longer needed);

### Remaining comments/nits

(specifics are inlined)

+ one inaccurate prose paragraph;
+ a foldr-vs-foldl consistency question;
+ some dead code.
+ In `CHANGELOG.md`: WIP bullets referencing `updateCertDeposits` ("Move cert-deposit helpers from Utxo to Certs", "Fix updateCertDeposits: use foldl...") describe stuff this PR removes.  Those entries should be removed or updated.

src/Ledger/Dijkstra/Specification/Certs.lagda.md
carlostome: I agree that it makes more sense to use `foldl` since we think of the list of certificates to be applied left to right (even if technically doesn't matter).

Issue #1254, opened 2026-07-08
flake: use git+https for inputs so `nix develop` works behind the GitHub tarball proxy


williamdemeo:
## Summary

`nix develop` fails in the Claude Code web (cloud sandbox) environment because our flake inputs are declared with the `github:` shorthand, which fetches source as tarballs from `api.github.com`/`codeload.github.com`. Those endpoints are repo-scoped by the sandbox's GitHub proxy and return `403` for any repo outside the current session's scope. Switching the inputs to `git+https://github.com/...` resolves this with no change to the pinned revisions. A companion setup-script prefetch then removes the resolution cost entirely for subsequent sessions.

## Symptom

```
$ nix develop --command agda --version
error: unable to download
  'https://github.com/hercules-ci/flake-parts/archive/57928607….tar.gz':
  HTTP error 403
  {"message":"GitHub access to this repository is not enabled for this session…"}
```

Resolution dies on the first third-party input; the same would happen for `nixpkgs`, `agda.nix`, and the transitive Agda libraries.

## Root cause

The sandbox routes outbound traffic through two independent gates:

- **Security proxy** — a host allowlist for plain HTTPS. GitHub *tarball* endpoints go through here and are scoped to the session's own repositories (→ `403` for third-party repos, even public ones).
- **GitHub proxy** — handles *git operations* with a scoped credential that reaches any public repo.

Nix's `github:` fetcher uses the tarball path (the blocked one). Plain `git`-over-https uses the working path. Confirmed empirically:

| Fetch of `hercules-ci/flake-parts` (out of session scope) | Result |
|---|---|
| `curl` archive tarball / `nix flake metadata github:…` | `403` |
| `git ls-remote …` / `nix flake metadata git+https://…` | success |

## Proposed change

Two parts:

1. **Route inputs through git.** Declare each input as `git+https://github.com/<owner>/<repo>` instead of `github:<owner>/<repo>`, and pin the `flake.lock` nodes to the `git` fetcher. The `narHash` for any given revision is **identical** between the two fetchers, so every pin is preserved — this is a no-op on an unrestricted network and does not bump any dependency. Touches only `flake.nix` (the three direct inputs) and `flake.lock` (all nodes, including transitive ones pulled in via `agda.nix` and `flake-parts`).

[...]

PR #1255, opened 2026-07-08
flake: fetch inputs via git+https so nix develop works behind the Claude web proxy (#1254)

williamdemeo:
# Description

Addresses #1254. In Claude Code web sandboxes, `nix develop` fails immediately: the sandbox's GitHub proxy returns 403 for the `github:` flake-ref shorthand (the tarball endpoints are scoped to the session's own repo), but serves plain git-over-https for any public repo. This PR re-declares the flake inputs as `git+https://github.com/...` accordingly. The `narHash` of every pinned rev is identical between the two fetchers, so **no dependency changes — same revs, same hashes, a no-op on unrestricted networks** (empirical before/after in the comment below).

**Commit 1** converts the three direct inputs and pins every `flake.lock` node (direct + transitive) to the `git` fetcher; `nixpkgs` also gets `shallow=1`, so a binary-cache miss fetches one rev rather than full history. **Commit 2** future-proofs updates: a plain `nix flake update` would regress transitive nodes to `github:` (their upstream flakes still declare them that way), so `flake.nix` documents a tested `--override-input` update invocation plus a one-line check for inputs upstreams add later.

Verified in the web sandbox (comment below) and locally: `nix develop --command agda --version` → `Agda version 2.8.0`, everything substituted from `cache.nixos.org` / `cache.iog.io`.

# Checklist

- [ ] Any semantic changes to the specifications are documented in `CHANGELOG.md` — n/a, infrastructure-only change with no semantic changes to the specifications
- [x] Self-reviewed the diff

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -589,25 +589,53 @@ See [the `conformance-example` directory][conformance-example].
 ## 🗃️ Miscellanea
 
 
-### Updating Agda dependencies in the Nix configuration
+### Updating Nix flake inputs
+
+The flake inputs are declared as `git+https://...` URLs rather than the
+`github:` shorthand, because some sandboxed environments block the GitHub
+tarball endpoints that `github:` uses (see the note in `flake.nix`). A plain
+`nix flake update` would re-introduce `github:` entries for the transitive
+inputs, so update all inputs with
+
+```
+./build-tools/nix/update-flake-inputs.sh
+```
+
+The script (bash + jq) runs `nix flake update`, then converts any `github:`-typed
+entries in `flake.lock` back to `git+https`, including entries for inputs that
+upstream flakes acquired since the last update.  Run it on a network with
+unrestricted GitHub access.
+
+
+#### Updating Agda dependencies in the Nix configuration
 
 The following example ilustrates the procedure
 
 ```
 nix flake update agda-nix/abstract-set-theory \
   --override-input agda-nix/abstract-set-theory \
-  github:input-output-hk/agda-sets/bbaa00abc4aef061896ae5d3cdec148bfbf5029f
+  'git+https://github.com/input-output-hk/agda-sets?rev=bbaa00abc4aef061896ae5d3cdec148bfbf5029f'
 nix build ./#fls-agdaWithPackages -o ~/ledger-agda
 ```
 
-The first line updates the commit hash to use for a dependency. In the example,
-it updates the dependency `agda-nix/abstract-set-theory` to point at the commit
-`bbaa00abc4aef061896ae5d3cdec148bfbf5029f` or the repository
-`github:input-output-hk/agda-sets`.
+The first line updates the commit hash to use for a dependency.  In the example, it
+updates the dependency `agda-nix/abstract-set-theory` to point at the commit
+`bbaa00abc4aef061896ae5d3cdec148bfbf5029f` of the repository `input-output-hk/agda-sets`.
+Use the `git+https` form rather than `github:` here, so the lock entry stays on the
+git fetcher.
 
-The second line rebuilds the Agda mode to use with emacs. This step is necessary
-for emacs to use the new version of the dependency when loading Agda code. This
-assumes that `~/ledger-agda/bin` is in your `PATH`.
+Note that re-locking one input this way can also refresh its sibling inputs and
+re-introduce `github:` entries for them.  Afterwards, convert those without updating
+anything further, as follows:
+
+```
+./build-tools/nix/update-flake-inputs.sh --convert-only
+```
+
+The second command in the example rebuilds the project's `agda` executable (with the
+updated libraries) and links the result at `~/ledger-agda`.  If `~/ledger-agda/bin`
+is on your `PATH`, emacs (via agda-mode) picks up the new dependency version the next
+time it loads Agda code.
 
 
 ### Plotting typechecking times
--- a/build-tools/nix/update-flake-inputs.sh
+++ b/build-tools/nix/update-flake-inputs.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+#
+# Update the Nix flake inputs, keeping every flake.lock entry on the git
+# fetcher. The inputs use git+https URLs because some sandboxed environments
+# block the GitHub tarball endpoints behind the github: shorthand (see the
+# note in flake.nix). A plain `nix flake update` re-reads transitive inputs
+# from their upstream flakes, which still declare github:, so this script
+# updates and then converts any github:-typed lock entries back to git+https,
+# repeating until the lock is clean.
+#
+# Usage: ./build-tools/nix/update-flake-inputs.sh [--convert-only]
+#   (default)        update all inputs, then convert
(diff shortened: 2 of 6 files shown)


williamdemeo:
## Empirical test in the environment this PR targets

*(Edited 2026-08-10: trimmed to the flake evidence after the PR was reduced to its flake-only commits; the dev-shell prefetch script this comment also covered was dropped from the PR.)*

Ran 2026-07-28 in a **fresh Claude Code web sandbox** for this repo: recycled container, completely empty `/nix`; Nix 2.24.9 bootstrapped by hand from `releases.nixos.org` (allowed by the sandbox proxy, as are `cache.nixos.org` and `cache.iog.io`).

| # | Scenario | Command | Result |
|---|---|---|---|
| A | **master flake** (`github:` inputs) | `nix develop --command true` | **Fails in <1 s** — the first input fetch (`flake-parts`) gets HTTP 403 from the session proxy: *"GitHub access to this repository is not enabled for this session"* |
| B | **this PR's flake** (`git+https` inputs), stone-cold empty store | `nix develop --command agda --version` | **≈90 s**, success: `Agda version 2.8.0` — all inputs resolved over `git+https`, ~5.6 GB / 110+ store paths substituted from `cache.nixos.org` + `cache.iog.io`, nothing compiled locally. Repeat runs on the warm store: **~2 s** |

Direct endpoint probes confirm the mechanism: `codeload.github.com` / `api.github.com` → **403** for any repo not attached to the session, while `git ls-remote https://github.com/input-output-hk/agda.nix` succeeds through the same proxy.

Two readings worth keeping: the flake change is the **on/off switch**, not an optimization (A vs B) — and with identical narHashes it is a no-op elsewhere. Also, nix substituted the **nixpkgs source itself by narHash** from the binary cache, so no nixpkgs clone happened at all; the `shallow=1` on the nixpkgs input is the fallback for when source substitution misses.

🤖 Posted from a [Claude Code](https://claude.com/claude-code) session that ran the experiments above.

flake.nix
carlostome, line 1: If it is this complicated to update the flake I suggest adding a script (preferably bash + jq) to be able to update the inputs.

Also add a reference to this script in the CONTRIBUTING in how to update the nix flake inputs,

flake.nix
```diff
@@ -2,15 +2,44 @@
...
+  #     --override-input agda-nix/standard-library-meta git+https://github.com/agda/agda-stdlib-meta \
+  #     --override-input flake-parts/nixpkgs-lib git+https://github.com/nix-community/nixpkgs.lib
+  #
+  # then verify nothing github-locked slipped in (upstreams gain inputs over time;
+  # `nix flake metadata` shows a new input's path; append an override):
+  #
+  #   python3 -c "import json; L=json.load(open('flake.lock'))['nodes']; \
+  #     print(*[k for k,v in L.items() if v.get('locked',{}).get('type')=='github'] or ['OK'])"
+  #
+  # (Transitive *original* fields read github after an update; harmless since fetches go by *locked* entries.)
```
carlostome: I don't think this documentation belongs here. Maybe only a pointer to CONTRIBUTING and how to update the flakes

flake.nix
williamdemeo: Done: `build-tools/nix/update-flake-inputs.sh` (bash + jq). 

It derives the github-to-git+https conversions from `flake.lock` itself rather than a fixed list, so it stays valid when upstream flakes gain inputs.

`CONTRIBUTING.md` has a new "Updating Nix flake inputs" section, and the existing single-dependency example there now uses the git+https form as well.

flake.nix
williamdemeo: Agreed. The update documentation moved to `CONTRIBUTING.md`; the `flake.nix` note is now the short "why" plus a pointer.

build-tools/nix/update-flake-inputs.sh
```diff
@@ -0,0 +1,84 @@
...
+# For every github:-typed node, emit "--override-input <path> <git+https url>"
+# as tab-separated fields, preserving a declared branch or subdirectory.
+overrides() {
+  local path key
+  while IFS=$'\t' read -r path key; do
+    jq -r --arg k "$key" --arg p "$path" '
+      .nodes[$k]
+      | select(.locked.type == "github")
+      | (if .original.type == "github" then .original else .locked end) as $o
+      | ([$o.ref // empty | "ref=\(.)"] + [$o.dir // empty | "dir=\(.)"]) as $q
```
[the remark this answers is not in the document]
williamdemeo: Confirmed, and reproduced before fixing: pinning `agda-nix/abstract-set-theory` to an old rev via `github:input-output-hk/agda-sets/bbaa00ab…` and running `--convert-only` converted the fetcher but re-resolved the revision (`bbaa00a` became `2253231`). The overrides now pin `rev=<locked.rev>` (plus `ref`/`dir` when declared), so conversion changes the fetcher and nothing else. The same fixture now preserves the old revision, and the full-update mode was re-verified end to end (one conversion pass, no github-typed entries remain). Fixed in a7dd792a6.

flake.nix
```diff
@@ -2,15 +2,27 @@
...
 
+  # NOTE: inputs use git+https://... instead of the github: shorthand.
+  # The github: fetcher pulls tarballs from api.github.com / codeload.github.com,
+  # which the Claude Code web sandbox proxy 403s for third-party repos; plain
+  # git-over-https works there for any public repo.
+  #
+  # narHashes are identical between the two fetchers, so the pins are unchanged.
+  #
+  # nixpkgs adds `shallow=1` so a binary-cache miss fetches only the locked rev.
+  #
```
carlostome, line 13: I'd say to remove this since the motivation is already clear enough in CONTRIBUTING

CONTRIBUTING.md
```diff
@@ -589,25 +589,53 @@ See [the `conformance-example` directory][conformance-example].
 ## 🗃️ Miscellanea
 
 
-### Updating Agda dependencies in the Nix configuration
+### Updating Nix flake inputs
+
+The flake inputs are declared as `git+https://...` URLs rather than the
+`github:` shorthand, because some sandboxed environments block the GitHub
+tarball endpoints that `github:` uses (see the note in `flake.nix`). A plain
```
carlostome, line 596:
```suggestion
tarball endpoints that `github:` uses. A plain
```

flake.nix
```diff
@@ -2,15 +2,27 @@
...
+  # To update the inputs, run build-tools/nix/update-flake-inputs.sh (see
+  # "Updating Nix flake inputs" in CONTRIBUTING.md); a plain `nix flake
+  # update` would regress transitive inputs to github:.
```
carlostome, line 16: I'd change this to a more noticeable comment:

```suggestion
  # !WARNING!:
  # To update the flake's inputs, DO NOT USE `nix flake update`. Instead run build-tools/nix/update-flake-inputs.sh (see
  # "Updating Nix flake inputs" in CONTRIBUTING.md for more info).
```

williamdemeo:
We've decided to put off supporting web-ui.  We can resurrect this PR if/when that changes.

Issue #1261, opened 2026-07-15
[Dijkstra] Legacy-mode top-level withdrawals can drain mid-batch direct deposits


williamdemeo:
## Description

In the `ENTITIES` legacy branch, the top-level withdrawal must equal the current balance — which, at the `LEDGER-V` call site, is the post-SUBLEDGERS balance including sub-transaction direct deposits — and is not bounded by `rewards₀`, unlike every other withdrawal in the batch.

Since `UsesV4Features` inspects only the top-level transaction's own fields, a legacy-mode batch can still contain sub-transactions with direct deposits. Consequently a legacy top-level tx can withdraw funds deposited mid-batch — funds that the pre-V4 script triggering legacy mode cannot even see, since its script context cannot represent sub-transactions. 

### Candidate fixes

+ Bound legacy top-level withdrawals by `rewards₀`, 
+ extend `UsesV4Features` to sub-transaction fields, OR
+ forbid sub-transactions in legacy mode entirely.

carlostome:
Closing as this is a feature

PR #1262, opened 2026-07-15
[Dijkstra] Move normal/legacy mode decision to LEDGER

### LINKED ISSUE #1258: [Dijkstra] Move logic for deciding whether to process in normal or legacy mode to Ledger
Thread the result via rules' environment.

williamdemeo:
## Description

Closes #1258.

Previously the normal-vs-legacy decision lived in two places that could disagree: `UTXOW` chose a mode nondeterministically via its two rules (feeding `false`/`true` into `UTXO`'s self-balance premise), while `LEDGER-V` passed a hardcoded `true` to `ENTITIES` — making the `legacyMode ≡ false` branch of `ENTITIES` (partial top-level withdrawals) unreachable and letting an all-V4 transaction get normal-mode UTxO accounting but legacy-mode withdrawal treatment.

This PR makes the decision once, in `LEDGER`, and threads the result through the rules' environments.

### Changes

+  **`Utxow`**: new `neededP2Scripts` and `isLegacyMode : UTxO → ℙ Script → TopLevelTx → Bool` — legacy exactly when some needed phase-2 script uses a pre-V4 Plutus language (the same criterion `Computational-UTXOW` already used to dispatch).  `UTXOW-normal`/`UTXOW-legacy` gain a first premise `LegacyModeOf Γ ≡ false`/`≡ true` and pass `Γ` to `UTXO` unchanged.  Their language premises remain as checks, so the environment flag is validated rather than trusted, and each rule keeps its standalone meaning.

+  **`Utxo`**: `UTxOEnv` gains a named `legacyMode : Bool` field (mirroring how `isTopLevelValid` is threaded via `SubUTxOEnv`), with a `HasLegacyMode`/`LegacyModeOf` accessor. `UTXO`'s environment is plain `UTxOEnv` again instead of `UTxOEnv × Bool`.

+  **`Ledger`**: `LEDGER-V`/`LEDGER-I` compute `legacyMode = isLegacyMode utxo₀ allScripts tx` and thread it to `UTXOW`/`UTXO` via `UTxOEnv` and to `ENTITIES` via `EntitiesEnv`, replacing the hardcoded `true`. All three rules now see one consistent flag.

+  **Computational**: `Computational-UTXOW` dispatches on the flag premises, which removes the `DecV3` decision and the `V1,V2,V3∩V4⊆∅` disjointness lemma; `Computational-UTXO` loses the product-env plumbing; `Computational-LEDGER` mirrors the new rule environments.

+  **`Foreign/Utxo`**: `utxo-step` is now `UTxOEnv → UTxOState → Tx TxLevelTop → …` (the flag moved into the auto-derived `MkUTxOEnv`; no Haskell consumers exist yet).

### Verification

The full `Ledger.Dijkstra` aggregate typechecks (Agda 2.8.0 via the Nix flake), as does `Specification/Computational.lagda.md`, which sits outside the aggregate's import graph. No modules outside `src/Ledger/Dijkstra` import the changed ones.

### Out of scope

The legacy-mode withdrawal gap is *not* fixed here: the `ENTITIES` legacy branch still requires the top-level withdrawal to equal the current (post-`SUBLEDGERS`) balance with no `rewards₀` bound, and `UsesV4Features` only inspects top-level fields, so a legacy batch can still carry sub-transactions with direct deposits and drain them. This is now tracked in Issue #1261.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

**Checked/revised by @williamdemeo**

## Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Specification/Utxow.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Utxow.lagda.md
@@ -371,7 +371,7 @@ attempting both.
     in
 
     ∙ LegacyModeOf Γ ≡ false
-    ∙ ∀[ s ∈ p2ScriptsNeeded ] language s ∈ fromList (PlutusV4 ∷ []) -- (1)
+    ∙ isLegacyMode utxo₀ scriptsProvided txTop ≡ false -- (1)
     ∙ concatMapˡ (λ txSub → mapˢ proj₁ (TopLevelGuardsOf txSub)) (SubTransactionsOf txTop) ⊆ GuardsOf txTop -- (2)
     ∙ ∀[ (vk , σ) ∈ TxWitnesses.vKeySigs (Tx.txWitnesses txTop) ] isSigned vk (txidBytes (TxIdOf txTop)) σ
     ∙ ∀[ s ∈ p1ScriptsNeeded ] validP1Script vKeyHashesProvided (GuardsOf txTop) txVldt s
@@ -464,7 +464,7 @@ attempting both.
     in
 
     ∙ LegacyModeOf Γ ≡ true
-    ∙ ∃[ s ∈ p2ScriptsNeeded ] language s ∈ fromList (PlutusV1 ∷ PlutusV2 ∷ PlutusV3 ∷ []) -- (1)
+    ∙ isLegacyMode utxo₀ scriptsProvided txTop ≡ true -- (1)
     ∙ concatMapˡ (λ txSub → mapˢ proj₁ (TopLevelGuardsOf txSub)) (SubTransactionsOf txTop) ⊆ GuardsOf txTop -- (2)
     ∙ ∀[ (vk , σ) ∈ vKeySigs ] isSigned vk (txidBytes (TxIdOf txTop)) σ
     ∙ ∀[ s ∈ p1ScriptsNeeded ] validP1Script vKeyHashesProvided (GuardsOf txTop) txVldt s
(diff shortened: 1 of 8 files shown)


src/Ledger/Dijkstra/Specification/Utxow.lagda.md
```diff
@@ -338,6 +370,7 @@ mode up front rather than deciding both.
                                       (credsNeeded utxo₀ txTop)
     in
 
+    ∙ LegacyModeOf Γ ≡ false
     ∙ ∀[ s ∈ p2ScriptsNeeded ] language s ∈ fromList (PlutusV4 ∷ []) -- (1)
```
williamdemeo: replace with isLegacymode = false check.

src/Ledger/Dijkstra/Specification/Utxow.lagda.md
```diff
@@ -430,6 +463,7 @@ mode up front rather than deciding both.
                                       (credsNeeded utxo₀ txTop)
     in
 
+    ∙ LegacyModeOf Γ ≡ true
     ∙ ∃[ s ∈ p2ScriptsNeeded ] language s ∈ fromList (PlutusV1 ∷ PlutusV2 ∷ PlutusV3 ∷ []) -- (1)
```
williamdemeo: replace with isLegacymode = true

carlostome (approved):
LGTM! (modulo the comments we discussed)

src/Ledger/Dijkstra/Specification/Utxow.lagda.md
williamdemeo: done

src/Ledger/Dijkstra/Specification/Utxow.lagda.md
williamdemeo: done

PR #1263, opened 2026-07-20
[Dijkstra] Add requiredTopLevel guards at the top level

carlostome:
# Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages
- [ ] Any semantic changes to the specifications are documented in `CHANGELOG.md`
- [ ] Code is formatted according to [CONTRIBUTING.md](https://github.com/input-output-hk/formal-ledger-specifications/blob/master/CONTRIBUTING.md)
- [x] Self-reviewed the diff

PR #1265, opened 2026-07-21
[Dijkstra] Add missing script checks to UTXOW

carlostome:
# Description

- Check that txOuts locked by Plutus V1-V2 scripts have datahashes/inline datums
- Add check for Conway-era certificates to `UsesV3Features`

_Blocked by #1263_

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Specification/Certs.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Certs.lagda.md
@@ -139,6 +139,17 @@ record GovCertEnv : Type where
 
 <!--
 ```agda
+IsConwayCert? : IsConwayCert ⁇¹
+IsConwayCert? {x} .dec with x
+... | regdrep _ _ _ = yes tt
+... | deregdrep _ _ = yes tt
+... | ccreghot _ _  = yes tt
+... | delegate _ (just _) _ _ = yes tt
+... | delegate _ nothing  _ _ = no (λ ())
+... | dereg _ _ = no (λ ())
+... | regpool _ _ = no (λ ())
+... | retirepool _ _ = no (λ ())
+
 record HasDeposits (A : Type) {K : Type} : Type where
   field DepositsOf : A → K ⇀ Coin
 open HasDeposits ⦃...⦄ public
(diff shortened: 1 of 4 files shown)


williamdemeo (approved):
Nice work Carlos!  Both additions look semantically right.

Minor comments/observations:

**`hasConwayCerts` / `IsConwayCert`**.

+  The vote delegation classification matches the Conway spec's existing `IsConwayCert` (modulo Dijkstra's `DCert` having no `reg` constructor), and via `UsesV3Features` it excludes PlutusV1/V2 from `allowedLanguagesLegacy`.

+  One observation, not a blocker: in the implementation, deregistration/registration with an *explicit deposit* also fails V1/V2 translation, but this spec's `DCert` bakes deposits in unconditionally (`dereg : Credential → Coin → DCert`), so that distinction isn't representable here.  Is this worth a comment in the code?  (Same applies to the pre-existing Conway version, so this is really a general/upstream note.)

+  Organization nit: `IsConwayCert` lives in `Certs.lagda.md` but its decidability `IsConwayCert?` lives in `Utxow.lagda.md`. Should we move `IsConwayCert?` next to `IsConwayCert` ? 

**`TxOutSpendable-PlutusV1-V2`**.

+  The predicate correctly encodes Alonzo/Babbage unspendability: V1-locked outputs need a datum *hash* (an inline datum won't satisfy it; this is correct, since V1 can't use inline datums), V2-locked outputsneed hash or inline datum.  Quantifying over `range (utxo₀ ∣ SpendInputsOf txTop)` matches "outputs being spent," and adding the premise only to `UTXOW-legacy` is right, since normal mode is V4-only.

+  The `Maybe.All … (txOutToP2Script scriptsProvided txOut)` formulation is trivially satisfied when the locking script isn't in the script pool, which is fine because `scriptHashesNeeded ⊆ mapˢ hash scriptsProvided` forces needed scripts to be in the pool.

+  The new definition lands in the rendered spec between `allowedLanguagesLegacy` and `allowedLanguages` with no prose.  A sentence explaining it (analogous to the Alonzo "unspendable outputs" discussion) would help readers, plus a mention in the numbered conditions above `UTXOW-legacy`.

**Other**.

Because this PR is based on the #1263 branch, only `shellcheck`/`ci/eval` ran... fine since the full CI should run once this retargets to `master` after #1263 merges.

PR #1268, opened 2026-07-22
[Conway/Dijkstra] Move txIns ∩ refInputs ≡ ∅ to allowedLanguages

carlostome:
# Description

For backwards compatibility Plutus V1-V2 scripts are allowed in transactions where the inputs and reference inputs are not disjoint sets. In contrast, when this is the case, Plutus V3 scripts are not allowed.

For more info see https://github.com/IntersectMBO/cardano-ledger/pull/5011

In addition, this PR adds the disjointness check to Plutus V4 scripts.
  
~_Blocked by https://github.com/IntersectMBO/formal-ledger-specifications/pull/1265_~

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Specification/Utxow.lagda.md
+++ b/src/Ledger/Conway/Specification/Utxow.lagda.md
@@ -67,18 +67,17 @@ IsConwayCert (delegate _ (just _) _ _) = ⊤
 IsConwayCert _               = ⊥
 
 private
-  instance
-    IsConwayCert? : IsConwayCert ⁇¹
-    IsConwayCert? {x} .dec with x
-    ... | regdrep _ _ _ = yes tt
-    ... | deregdrep _ _ = yes tt
-    ... | ccreghot _ _  = yes tt
-    ... | delegate _ (just _) _ _ = yes tt
-    ... | delegate _ nothing  _ _ = no (λ ())
-    ... | dereg _ _ = no (λ ())
-    ... | regpool _ _ = no (λ ())
-    ... | retirepool _ _ = no (λ ())
-    ... | reg _ _ = no (λ ())
+  IsConwayCert? : IsConwayCert ⁇¹
+  IsConwayCert? {x} .dec with x
+  ... | regdrep _ _ _ = yes tt
+  ... | deregdrep _ _ = yes tt
+  ... | ccreghot _ _  = yes tt
+  ... | delegate _ (just _) _ _ = yes tt
+  ... | delegate _ nothing  _ _ = no (λ ())
+  ... | dereg _ _ = no (λ ())
+  ... | regpool _ _ = no (λ ())
+  ... | retirepool _ _ = no (λ ())
+  ... | reg _ _ = no (λ ())
 
 module _ (txb : TxBody) (let open TxBody txb) where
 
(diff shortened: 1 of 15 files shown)


williamdemeo (commented):
Great work, Carlos!

The core change looks right to me and it matches the intent of cardano-ledger#5011.

Here's a summary of what I verified, followed by two questions and a few minor points.

**Verified**.

+  Removing `txIns ∩ refInputs ≡ ∅` from the Conway UTXO rule (spec + conformance copies) and gating PlutusV3 on disjointness inside `allowedLanguages` gives exactly the intended behavior: a scriptless or V1/V2-only transaction with overlapping inputs/ref-inputs is now valid (Babbage-compatible), while any transaction *needing* a V3 script with overlap fails the `languages … ⊆ … ∩ allowedLanguages` premise.
+  The refactored union-of-guarded-singletons form is equivalent to the old if-chain for the V1 and V2 clauses in both Conway and Dijkstra.
+  All the arity-adjustment changes across the Equivalence/Properties/PoV/MinSpend files are mechanical and consistent, and the two-column → one-per-line premise formatting is a readability win.
+  Dijkstra's UTXO/SUBUTXO rules never had a disjointness premise, so no vacuous issue on that side.

**Questions**.

1.  **Should PlutusV4 also require disjointness?** In the implementation the restriction is "V3 *and newer*" (it lives in the V3+ context translation), but after this PR, Dijkstra allows V4 with overlapping spend/reference inputs everywhere: normal-mode `allowedLanguages` is `{V4}` gated only on bootstrap addresses, and the legacy V4 clause has no disjointness condition either. If overlap is deliberately re-allowed for V4 (e.g. the V4 context represents it fine), a comment in the spec would be valuable; if not, the V4 clauses need the same condition.
2. **The legacy V4 clause changed semantics slightly**.  The old code returned `∅` when `UsesV4Features tx` held, while the new code always includes `PlutusV4` (absent bootstrap addresses).  I convinced myself this is unobservable *inside* `UTXOW-legacy` (legacy mode guarantees a needed V1–V3 script, which already fails against `{V4}`), but it's a behavioral change of the function itself.  Was it intentional?  Worth documenting either way, in case `allowedLanguagesLegacy` is ever used elsewhere.

**Minor**.

+  In Dijkstra, `UsesV2Features` is now dead code — the new `allowedLanguagesLegacy` inlines its definition (and likewise inlines the bootstrap-address check instead of using `UsesBootstrapAddress`, which is still used by `allowedLanguages` at line 177).  Inlined near-copies invite drift; I'd reuse the named definitions or delete the orphan.

+  The prose above Conway's `allowedLanguages` wasn't updated.  It should mention that V3 additionally requires spend inputs and reference inputs to be disjoint.

+  Since this PR is based on the #1265 branch, the Agda typecheck hasn't run in CI on this head; the proof adjustments here are exactly the kind of thing that needs the full pipeline, so let's make sure CI runs after retargeting.

williamdemeo (commented):
Re-reviewed at head `99a3e61` (after the rebase over #1273).  Nice work getting this aligned with cardano-ledger#5011, but the rebase over Polina's fix introduced one small regression that's worth a quick fix before merging.

**Suggested fix**.  In `Conway/Specification/Utxo/Properties/Computational.lagda.md`, the `genErr` cascade still contains the branch for `"¬ txIns ∩ refInputs ≡ ∅"`, but this PR removes that premise from the UTXO rule, so from that branch onward every error message now describes the wrong premise (e.g. a failing `inInterval` check reports the disjointness message).  #1273 just realigned these messages, and this rebase partially undoes that.  Deleting the stale branch (shifting the cascade up one level) restores the alignment.  It typechecks either way, so CI won't flag it.

src/Ledger/Conway/Specification/Utxow.lagda.md
williamdemeo, line 130: The prose above `allowedLanguages` still describes the old behavior; it should mention that Plutus V3 (and V4) now additionally requires spend inputs and reference inputs to be disjoint.

williamdemeo (approved):
Nice work!  LGTM!!

PR #1273, opened 2026-07-28
conway collateral fix

polinavino:
# Description

The Conway spec was missing the check `collateralInputs ⊆ dom utxo`, which is now added. This is noted in the changelog, and reflected in the conformance testing. The `genErr` error messages adjusted to match the UTXO rule also.


carlostome (approved):
LGTM, thank you!

Issue #1233, opened 2026-06-23
[Conway] voteDelegs range is contained in VDelegs built from its domain


williamdemeo:
**Property:** voteDelegs range is contained in VDelegs built from its domain

- Catalog id: `conway-certs-votedelegs-vdeleg`
- Era: **conway**  ·  STS: **CERTS**
- Agda module: `Ledger.Conway.Specification.Certs.Properties.VoteDelegsVDeleg`  (anchor: `clm:VDelegsInRegDReps`)
- Key definitions: `voteDelegsVDeleg`
- Status (derived from the Agda): **stated**

---
Tracked in `docs/notes/properties.yaml`; see ADR `docs/notes/0001-ledger-property-tracking.md` and the roadmap `docs/notes/ledger-properties-roadmap.md`.
Part of the properties roadmap (umbrella #45).

PR #1288, opened 2026-08-05
Move vote delegation removal to GovCert

carlostome:
# Description

This PR changes the Conway (and Dijkstra) specification to match the implementation.
In particular, in the implementation a DRep deregistering certificate makes the DRep delegated stake be reset to empty, even if the same DRep is reregistered in the same transaction. 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Conway/Conformance/Equivalence/Certs.lagda.md
+++ b/src/Ledger/Conway/Conformance/Equivalence/Certs.lagda.md
@@ -160,12 +160,6 @@ instance
                      Γ C.⊢ (deposits ⊢conv s) ⇀⦇ _ ,PRE-CERT⦈ (deposits ⊢conv s')
   PRE-CERTToConf .convⁱ deposits (L.CERT-pre h) = C.CERT-pre h
 
-  POST-CERTToConf : ∀ {Γ s s'}
-                 → L.Deposits × L.Deposits
-                   ⊢ Γ L.⊢ s ⇀⦇ _ ,POST-CERT⦈ s' ⭆ⁱ λ deposits _ →
-                     Γ C.⊢ (deposits ⊢conv s) ⇀⦇ _ ,POST-CERT⦈ (deposits ⊢conv s')
-  POST-CERTToConf .convⁱ deposits L.CERT-post = C.CERT-post
-
   DELEGToConf : ∀ {Γ s dcert dcerts s'}
                   (open L.DelegEnv Γ renaming (pparams to pp))
               → CertDeps* pp (dcert ∷ dcerts) ⊢
@@ -183,7 +177,7 @@ instance
                   (open L.CertEnv Γ using (pp))
                 → CertDeps* pp (dcert ∷ dcerts) ⊢
                    Γ L.⊢ s ⇀⦇ dcert ,GOVCERT⦈ s' ⭆ⁱ λ deposits _ →
-                   Γ C.⊢ (deposits .depsᵍ ⊢conv s) ⇀⦇ dcert ,GOVCERT⦈ (updateCertDeps deposits .depsᵍ ⊢conv s')
+                   Γ C.⊢ (getCertDeps* deposits ⊢conv s) ⇀⦇ dcert ,GOVCERT⦈ (getCertDeps* (updateCertDeps deposits) ⊢conv s')
   GOVCERTToConf .convⁱ (regdrep* _ _)     (L.GOVCERT-regdrep h) = C.GOVCERT-regdrep h
   GOVCERTToConf .convⁱ (deregdrep* v _ _) (L.GOVCERT-deregdrep h) = C.GOVCERT-deregdrep (h , v)
   GOVCERTToConf .convⁱ (ccreghot* _ _)    (L.GOVCERT-ccreghot h)  = C.GOVCERT-ccreghot h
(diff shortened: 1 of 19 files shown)


src/Ledger/Dijkstra/Specification/Entities.lagda.md
williamdemeo: The activeVDelegs restriction was dropped from ENTITIES but is still bound and still applied in SUBENTITIES (voteDelegs' ∣^ activeVDelegs in the conclusion).  Is this intentional?

LEDGER runs SUBENTITIES for every sub-transaction and then ENTITIES for the top-level one, so the two levels now disagree, and a DRep that deregisters and reregisters inside a sub-transaction still gets its delegated stake restored at the end of SUBENTITIES.

williamdemeo (commented):
Here's something Claude found that I'm not sure about, but I'll post it in case it helps you.

Generated Haskell no longer compiles.

`CertState` moved modules

+  `src/Ledger/Conway/Foreign/Certs.agda:83-84`
+  `src/Ledger/Dijkstra/Foreign/Certs.agda:62-63`

Moving `HsTy-CertState/Conv-CertState` out of `Foreign/Cert.agda` moves the generated Haskell datatype with it.

In this PR's hs artifact, `data CertState = MkCertState…` appears only in `MAlonzo/Code/Ledger/{Conway,Dijkstra}/Foreign/Certs.hs` and is gone from `Foreign/Cert.hs`.  But the static re-export modules still import it from the old location:

+  `build-tools/static/hs-src/src/MAlonzo/Code/Ledger/Conway/Foreign/API.hs:14-15`
+  `build-tools/static/hs-src/src/MAlonzo/Code/Ledger/Dijkstra/Foreign/API.hs:19-20`

MAlonzo emits modules without export lists, so the Agda-level public on the import does not re-export the Haskell type and GHC will reject `import …Foreign.Cert (…, CertState(..))`.

CI will not flag it: the hs target only assembles the tree, and `build-tools/nix/hs-src.nix` checks nothing beyond the presence of the `.cabal` file.  Both `API.hs` files need `CertState(..)` imported from `Foreign.Certs`.

williamdemeo:
Another Claude find seems useful:

The GOVCERT step is now resolved by an ambiguous instance search

`src/Ledger/Conway/Specification/Certs/Properties/Computational.lagda.md:106`,
`src/Ledger/Conway/Conformance/Certs/Properties.lagda.md:121`,
`src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md:103`

Computational-GOVCERT and Computational-CERT now have the same index type `CertEnv → CertState → DCert → CertState → Type`, so both are candidates for the bare `computeProof ce cs dCert`, and it only lands on `GOVCERT` because the metavariable is pinned later by `CERT-vdel h` on the right-hand side.

Please make it explicit: `computeProof {STS = _⊢_⇀⦇_,GOVCERT⦈_} ce cs dCert`.

I suspect this is also why the `CERT-vdel` completeness clauses had to be rewritten to re-inline each `GOVCERT` decision procedure; with the annotation the old with `computeProof… | completeness _ _ _ _ h` form should work again, and the `¿ … ¿` predicates would not be duplicated across three sites.

src/Ledger/Conway/Specification/Certs.lagda.md
williamdemeo, line 431:
```suggestion
   `DRep`{.AgdaInductiveConstructor}, and every vote delegation to that `DRep`{.AgdaInductiveConstructor} is dropped.
```

williamdemeo: This was meant to suggest replacing the period with a comma and adding the phrase that begins, "and every vote..."

src/Ledger/Conway/Conformance/Equivalence/Certs.lagda.md
williamdemeo, line 230: This is a pre-existing flaw, not something this PR introduced, but we might as well fix it here, since it sits in the middle of what the PR touches.

In `GOVCERTFromConf` the right-hand sides are `C.GOVCERT-…` where I think `L.GOVCERT-…` is meant (cf. `DELEGFromConf` above).

src/Ledger/Conway/Specification/Certs/Properties/PoVLemmas.lagda.md
```diff
@@ -184,30 +186,30 @@ value of the withdrawals in `Γ`{.AgdaBound}.  In other terms,
...
+    base-pov : {Γ : CertEnv} {s s' : CertState}
+      → IdSTS {CertEnv} {CertState} Γ s tt s'
       → getCoin s ≡ getCoin s'
 ```
 
 *Proof*.
 
 ```agda
-    POST-CERT-pov CERT-post = refl
+    base-pov Id-nop = refl
```
williamdemeo: This lemma is trivial now.  We could delete it an just inline the `refl` proof at line 234.

williamdemeo:
Confirmed the two fixes: `SUBENTITIES` no longer applies the filter, and the `API.hs` re-exports resolve again.  (Claude cross-checked every re-exported name in the three `API.hs` files against the generated modules in the latest `hs` artifact; all 194 resolve, and the same check on the previous artifact flags exactly the two `CertState` misses.)

Remaining items, none of them blocking:

**Docs**. `src/Ledger/Conway/Specification/Certs.lagda.md:430-431` still says only that `GOVCERT-deregdrep` deregisters a DRep; it should also say that vote delegations to that DRep are dropped, and the section could use a sentence on why GOVCERT takes a whole `CertState` while DELEG and POOL stay on their own component.

**Small things**.

- `stᵍ'` is declared but unused in all three `Certs` modules (`Specification/Certs.lagda.md:362`, `Conformance/Certs.lagda.md:106`, `Dijkstra/Specification/Certs.lagda.md:322`).
- `src/Ledger/Dijkstra/Specification/Certs/Properties/Computational.lagda.md:74-97` still calls the GOVCERT state `gs` although it is a `CertState`; Conway's copy was renamed to `cs`.
- `src/Ledger/Dijkstra/Specification/Certs.lagda.md:463`: `deposits` for the `GState` field and `deposits'` for the `DState` field reads like a post-state; `depositsᵈ`/`depositsᵍ` as in `Entities` would be clearer.

**Maybe worth recording**. Pruning at deregistration makes `range voteDelegs ⊆ mapˢ vDelegCredential (dom dReps) ∪ {vDelegAbstain, vDelegNoConfidence}` an inductive invariant of `CERT` rather than something re-imposed at the end of `CERTS`: `DELEG-delegate` admits only registered delegatees, `dom dReps` shrinks only at `deregdrep`, and the epoch boundary only bumps the epoch values (`Epoch.lagda.md:688-691`). That is a better basis for #1233 than POST-CERT was, and it invalidates #1290 as written, which states the property in terms of POST-CERT's `activeVDelegs`.

williamdemeo (approved):
Nice work, Carlos!  Approving, but see the (non-blocking) comments I added in case you want to address any of them here before merging.

Issue #1259, opened 2026-07-14
Missing fields from `StakePoolParams`


WhatisRT:
We're missing a bunch of fields compared to the [implementation](https://github.com/IntersectMBO/cardano-ledger/blob/master/libs/cardano-ledger-core/src/Cardano/Ledger/State/StakePool.hs#L424). Since conformance tests didn't find any, this means these are either disabled or never generated?

I believe we are missing `id, vrf, relays` and `metadata`. I think `accountAddress` is our `rewardAccount` (matching with the fact that their `AccountAddress` type is just a newtype wrapped `Credential`.

When adding these, it would be good to check what kind of validation the implementation performs on these. Note that except for the `vrf`, all of these are also missing from the original Shelley spec.

WhatisRT:
Ah, I believe the `id` field is just a difference in bundling. We have `regpool : KeyHash → StakePoolParams → DCert`, compared to https://github.com/IntersectMBO/cardano-ledger/blob/master/libs/cardano-ledger-core/src/Cardano/Ledger/Core/TxCert.hs#L132

PR #1294, opened 2026-08-13
Implement missing parts from StakePoolParams specification

### LINKED ISSUE #1272: Add `minPoolCost` PParam

carlostome:
# Description

- Add `minPoolCost` protocol parameter and enforce cost constraint of pool registration
- Add VRF to `StakePoolParams`. Enforce VRF uniqueness wrt registered pools
- Change `StakePoolParams` reward account from `Credential` to `RewardAddress`, and enforce NetworkId coincides during pool registration
- Add pool retirement conditions wrt pp `Emax`

Closes #1272 
Addresses part of #1259 

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/build-tools/static/hs-src/src/MAlonzo/Code/Ledger/Dijkstra/Foreign/API.hs
+++ b/build-tools/static/hs-src/src/MAlonzo/Code/Ledger/Dijkstra/Foreign/API.hs
@@ -21,7 +21,7 @@ import MAlonzo.Code.Ledger.Dijkstra.Foreign.Cert              as X
 import MAlonzo.Code.Ledger.Dijkstra.Foreign.Chain             as X
   (ChainState(..), Block(..), chainStep)
 import MAlonzo.Code.Ledger.Dijkstra.Foreign.Certs             as X
-  ( StakePoolParams(..), PState(..), DelegEnv(..), GovCertEnv(..), CertEnv(..), DState(..), DCert(..), GState(..), CertState(..)
+  ( StakePoolParams(..), PState(..), DelegEnv(..), GovCertEnv(..), PoolEnv(..), CertEnv(..), DState(..), DCert(..), GState(..), CertState(..)
   , delegStep, govCertStep, poolStep)
 import MAlonzo.Code.Ledger.Dijkstra.Foreign.Entities          as X
   (EntitiesEnv(..), entitiesStep, SubEntitiesEnv(..), subentitiesStep)
--- a/formal-ledger-test/src/Test/LedgerImplementation.lagda.md
+++ b/formal-ledger-test/src/Test/LedgerImplementation.lagda.md
@@ -61,6 +61,7 @@ module Implementation where
 
   PlutusScript = ℕ × (List Data → Bool)
   ScriptHash = ℕ
+  VRF = ℕ
 
   ExUnits      = ℕ × ℕ
   ExUnit-CommutativeMonoid =
--- a/src/Ledger/Core/Foreign/Crypto/Structure.agda
+++ b/src/Ledger/Core/Foreign/Crypto/Structure.agda
@@ -32,6 +32,7 @@ HSCryptoStructure : CryptoStructure
 HSCryptoStructure = record {
     pkk = HSPKKScheme
   ; ScriptHash = ℕ
+  ; VRF = ℕ
   }
 
 open CryptoStructure HSCryptoStructure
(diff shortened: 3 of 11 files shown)


williamdemeo (commented):
Nice, focused PR, and the commit sequence is clean; one concern per commit.

Two things are blocking, and there is one conformance question that should be settled before merging.

1.  **formal-ledger-test no longer builds**.  This is what the two red Hydra checks are telling us.  The cause is the new VRF field of CryptoStructure.  Test.LedgerImplementation builds its crypto structure from a module,

    ```agda
      SVCrypto = record
        { Implementation
        ; pkk = SVPKKScheme
        }
    ```
    and Implementation has no VRF, so VRF and the DecEq-VRF instance are both left as metas.  (`nix build .#formal-ledger-test` reproduces it.)

    **The fix**.  One line beside `ScriptHash = ℕ` in `formal-ledger-test/src/Test/LedgerImplementation.lagda.md` (line 63) clears it, and the build then goes through:

    ```agda
       ScriptHash = ℕ
       VRF = ℕ
    ```

2.  PoolEnv is missing from the Dijkstra Haskell API

    `build-tools/static/hs-src/src/MAlonzo/Code/Ledger/Dijkstra/Foreign/API.hs:24-25`

    re-exports poolStep but not PoolEnv(..).  Now that pool-step takes a PoolEnv rather than a PParams, a downstream consumer cannot call it; neither the argument type nor the MkPoolEnv constructor is nameable.  Every other environment record in that export list is there, so this is probably an oversight.

    **The fix**.  Adding `PoolEnv(..)` on line 24.

**The conformance question**.  POOL-rereg's VRF premise is more permissive than the implementation.

`Certs.lagda.md:441` puts the condition as "the new VRF is not used by any other pool":

```agda
∙ ¬ (poolParams .vrf ∈ mapˢ vrf (range (pools ∣ ❴ kh ❵ ᶜ) ∪ range (fPools ∣ ❴ kh ❵ ᶜ)))
```

The implementation states it differently:

```haskell
-- re-register Pool
Just stakePoolState -> do
  when (hardforkConwayDisallowDuplicatedVRFKeys pv) $ do
    sppVrf == stakePoolState ^. spsVrfL
      || Map.notMember sppVrf psVRFKeyHashes
        ?! injectFailure (VRFKeyHashAlreadyRegistered sppId sppVrf)
```

stakePoolState comes from psStakePools, so the equality escape hatch covers only the pool's active VRF; psVRFKeyHashes, by contrast, does hold the VRF recorded by a pending re-registration.  The two agree everywhere but one case.  Pool kh is active with vrf₀; in epoch E it re-registers with a fresh vrf₁, then re-registers again in the same epoch with vrf₁, say to correct cost.  The implementation rejects that second certificate, since vrf₁ ≠ vrf₀ and vrf₁ is already in psVRFKeyHashes.  Our rule accepts it, since `fPools ∣ ❴ kh ❵ ᶜ` drops the pending entry.

I prefer your formulation, and deriving the in-use set from pools and fPools rather than maintaining it incrementally avoids the bookkeeping the implementation has to do by hand.  Since conformance is the point of the PR, though, I would rather see the divergence decided than inherited: either match the implementation here, or raise it upstream as an unintended bar on re-registering twice in one epoch after a VRF change.

###  Non-blocking

1.  Alignment messed up in `Certs.lagda.md` around lines 428, 445 and 461.
2.  The premise block is repeated a few times.  The conjunction of the VRF, network-id and minPoolCost conditions is spelled out twice in Certs.lagda.md and four more times in Computational.lagda.md; the retirement triple appears twice more.  A name for each, say

    ```agda
    ValidPoolParams : PParams → Pools → Pools → StakePoolParams → Type
    ```

    would shorten the rules and, more importantly, keep the rule and its decision procedure from drifting apart the next time a premise moves.

carlostome:
>     1. **formal-ledger-test no longer builds**.  This is what the two red Hydra checks are telling us.  The cause is the new VRF field of CryptoStructure.

Fixed.

>     2. PoolEnv is missing from the Dijkstra Haskell API

Fixed.

> **The conformance question**. POOL-rereg's VRF premise is more permissive than the implementation.

This has already been reported in `cardano-ledger`: https://github.com/IntersectMBO/cardano-ledger/issues/6003

> ### Non-blocking
> 
>     1. Alignment messed up in `Certs.lagda.md` around lines 428, 445 and 461.

Fixed.
 
>     2. The premise block is repeated a few times.  

The premise is repeated exactly twice, so I think it does not merit its own definition (would be less straightforward).

williamdemeo:
Nice work, Carlos!  LGTM!

williamdemeo (approved):

PR #1299, opened 2026-08-28
[Leios] PParams following CIP-164

ch1bo:
# Description

Introduces the Protocol Parameters as specified in [CIP-164](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0164/README.md), but already anticipates the updates from https://github.com/cardano-foundation/CIPs/pull/1250

Based on https://github.com/williamdemeo/formal-ledger-specifications/pull/20, but could not push onto that fork and neither onto upstream. Feel free to just pick whatever from this PR.

I would have **stacked** [this PR on top](https://github.com/ch1bo/formal-ledger-specifications/pull/1), but that does not work across forks seemingly.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Specification/PParams.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/PParams.lagda.md
@@ -26,7 +26,7 @@ module Ledger.Dijkstra.Specification.PParams
 
 open import Data.Product.Properties
 open import Data.Nat.Properties using (m+1+n≢m)
-open import Data.Rational as ℚ using (ℚ)
+open import Data.Rational using (ℚ)
 open import Relation.Nullary.Decidable
 open import Data.List.Relation.Unary.Any using (Any; here; there)
 
@@ -36,7 +36,7 @@ open import Ledger.Prelude
 open import Ledger.Core.Specification.Crypto
 open import Ledger.Core.Specification.Epoch
 -- open import Ledger.Dijkstra.Specification.Script.Base
-open import Ledger.Prelude.Numeric using (UnitInterval; fromUnitInterval; ℕ⁺)
+open import Ledger.Prelude.Numeric using (UnitInterval; ℕ⁺)
 
 
 private variable
@@ -114,12 +114,12 @@ record PParams : Type where
     pv                            : ProtVer -- retired, keep for now
 
     -- Network group (Leios)
-    leiosHeaderDiffusionPeriod    : ℕ
+    leiosHeaderPeriod             : ℕ
     leiosVotingPeriod             : ℕ
     leiosDiffusionPeriod          : ℕ
     leiosMaxEBSize                : ℕ
     leiosMaxEBTxsSize             : ℕ
-    leiosCommitteeStakeCoverage   : UnitInterval
+    leiosCommitteeSize            : ℕ
     leiosQuorumStakeThreshold     : UnitInterval
     leiosMaxEBExUnits             : ExUnits
     leiosMaxRefScriptSizePerEB    : ℕ
@@ -169,14 +169,16 @@ Leios adds the *endorser block* (EB), an ordered list of transaction references
 that a block producer announces alongside its ranking block; a committee of
 stake pools votes on the EB, and a certificate carried by the following ranking
 block brings the referenced transactions into the ledger.  Three of the nine
-parameters measure a Leios round in slots (header diffusion, voting, and the
-additional diffusion that follows voting), four bound an EB (its reference list,
-the transactions listed, their script execution, and their reference scripts),
-and two are fractions of the total active stake:
-`leiosCommitteeStakeCoverage`{.AgdaField} (`σ_c`) is the stake the committee is
-selected to cover, `leiosQuorumStakeThreshold`{.AgdaField} (`τ`) the stake a
-certificate's signers must carry.  The ranking block keeps its existing bound
-`maxBlockSize`{.AgdaField}, so Leios adds no field for it.
+parameters measure a Leios round in wall-clock time (header diffusion, voting,
+and the additional diffusion that follows voting), four bound an EB (its
+reference list, the transactions listed, their script execution, and their
+reference scripts), `leiosCommitteeSize`{.AgdaField} (`N_c`) is the number of
+committee seats — the committee being the `N_c` pools with the most active
+stake — and `leiosQuorumStakeThreshold`{.AgdaField} (`τ`) is the fraction of
+the total active stake a certificate's signers must carry.  The ranking block
+keeps its existing bound `maxBlockSize`{.AgdaField}, so Leios adds no field for
+it.  Zero-valued Leios parameters are meaningful: they are the protocol's
+disabled state during rollout.
 
 The field names are this specification's; the cardano-ledger proposal
 [#5965][cl-5965], which maps the same parameters onto the Haskell `PParams`,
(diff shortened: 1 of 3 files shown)


src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -246,7 +314,9 @@ module PParamsUpdate where
   paramsUpdateWellFormed ppu =
        just 0 ∉ fromList ( maxBlockSize ∷ maxTxSize ∷ maxHeaderSize ∷ maxValSize
                          ∷ coinsPerUTxOByte ∷ poolDeposit ∷ collateralPercentage ∷ ccMaxTermLength
-                         ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit ∷ [] )
+                         ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit
+                         ∷ leiosHeaderPeriod ∷ leiosVotingPeriod ∷ leiosDiffusionPeriod
+                         ∷ leiosMaxEBSize ∷ leiosMaxEBTxsSize ∷ leiosMaxRefScriptSizePerEB ∷ [] )
```
williamdemeo: This list is what makes the disabled state unreachable: governance touches parameters only through updates (`actionWellFormed` for `ChangePParams` is `ppdWellFormed`, whose first component is this predicate), so the very update that would zero `leiosVotingPeriod` is rejected as ill-formed, and the state the new prose promises can never be enacted.

Dropping the six keeps the `ppWF` bridge in `Ledger.Dijkstra.Foreign.ExternalStructures` justified: the fields this predicate then checks are exactly `positivePParams`, so a well-formed update still cannot zero anything `paramsWellFormed` inspects.

```suggestion
       just 0 ∉ fromList ( maxBlockSize ∷ maxTxSize ∷ maxHeaderSize ∷ maxValSize
                         ∷ coinsPerUTxOByte ∷ poolDeposit ∷ collateralPercentage ∷ ccMaxTermLength
                         ∷ govActionLifetime ∷ govActionDeposit ∷ drepDeposit ∷ [] )
```

williamdemeo (commented):
Thanks for carrying M1-4 the rest of the way.  The alignment tracks the CIP update faithfully (wall-clock periods, the seat-count committee parameter, `S_EB-ref` retained now that Table 3 owns it, still no field for `S_RB`) and dropping `τ < σ_c` from `paramsWellFormed` has a welcome side effect: the `ppWF` bridge (the `trustMe` postulate) in `Ledger.Dijkstra.Foreign.ExternalStructures` is justified again, which resolves the soundness caveat flagged in the description of williamdemeo/formal-ledger-specifications#20.

One blocker noted below: the update predicate contradicts the disabled-state story; the rest is minor prose accuracy and polish.

@ch1bo please note: Claude helped me identify potential issues and come up with the suggestions below, so take them with a grain of salt.

🤖 This PR review was AI-assisted: Claude Fable 5 (Anthropic)

src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -151,17 +162,65 @@ record PParams : Type where
...
+uses different ones.  Its periods are `SlotInterval` lenses suffixed `Length`,
+the diffusion one further prefixed `Additional`.  Its two size bounds are
+`Word32` lenses named for the endorser-block *header* and *body*, its terms for
+the reference list and the transactions listed; those two bounds are
+`leiosMaxEBSize`{.AgdaField} and `leiosMaxEBTxsSize`{.AgdaField} here.  Its
+`OrdExUnits` lens is `leiosMaxEBExUnits`{.AgdaField}, one field for
+[CIP-164][cip-164]'s separate per-EB steps and memory budgets.  The remaining
+field, `leiosMaxRefScriptSizePerEB`{.AgdaField}, has no CIP-164 row at all; it
+is the proposal's own addition, the per-EB analogue of
+`maxRefScriptSizePerBlock`{.AgdaField}.
```
williamdemeo, line 194: Two statements here went stale against the revision this PR aligns to: `leiosMaxRefScriptSizePerEB` now *does* have a CIP-164 row (`S_EB-ref` joined Table 3 in the same update), and "maps the same parameters" no longer holds (#5965 still carries the stake-coverage fraction and slot-denominated periods).  While touching it, this paragraph is also the natural home for the one sentence the type change needs: the CIP states the periods in seconds, the way the genesis `slotLength` is given, and `Milliseconds` is their fixed-precision representation.

```suggestion
The field names are this specification's; the cardano-ledger proposal
[#5965][cl-5965], which maps an earlier revision of the table onto the
Haskell `PParams`, differs in names and in types.  Its periods are
`SlotInterval` lenses suffixed `Length`, the diffusion one further prefixed
`Additional`, where the table now measures wall-clock seconds;
`Milliseconds`{.AgdaFunction} here is the fixed-precision form of those
fractional seconds.  Its committee parameter is the earlier revision's
stake-coverage fraction, not a seat count.  Its two size bounds are `Word32`
lenses named for the endorser-block *header* and *body*, its terms for the
reference list and the transactions listed; those two bounds are
`leiosMaxEBSize`{.AgdaField} and `leiosMaxEBTxsSize`{.AgdaField} here.  Its
`OrdExUnits` lens is `leiosMaxEBExUnits`{.AgdaField}, one field for
[CIP-164][cip-164]'s separate per-EB steps and memory budgets.  The last
field, `leiosMaxRefScriptSizePerEB`{.AgdaField}, began as the proposal's own
addition and is now the table's `S_EB-ref`, the per-EB analogue of
`maxRefScriptSizePerBlock`{.AgdaField}.
```

src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -217,6 +276,15 @@ module PParamsUpdate where
           maxCollateralInputs           : Maybe ℕ
           maxTxExUnits maxBlockExUnits  : Maybe ExUnits
           pv                            : Maybe ProtVer -- retired, keep for now
+          leiosHeaderPeriod             : Maybe ℕ
+          leiosVotingPeriod             : Maybe ℕ
+          leiosDiffusionPeriod          : Maybe ℕ
```
williamdemeo: Same unit as the fields they update:

```suggestion
          leiosHeaderPeriod             : Maybe Milliseconds
          leiosVotingPeriod             : Maybe Milliseconds
          leiosDiffusionPeriod          : Maybe Milliseconds
```

src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -151,17 +162,65 @@ record PParams : Type where
...
 
+The Leios parameters are deliberately absent from
+`positivePParams`{.AgdaFunction}, since governance must be able to reach the
+protocol's disabled state: `leiosHeaderPeriod`{.AgdaField} and
+`leiosVotingPeriod`{.AgdaField} must be able to reach zero — with no voting
+window no certificate can ever form — while `leiosDiffusionPeriod`{.AgdaField}
+disables the other way around, at the largest duration its wire format can
+carry (durations serialize as 32-bit unsigned integers, so `2^32 − 1`), which
+pushes the earliest slot admitting a certificate beyond any chain.  CIP-164's
+quorum constraint `τ < σ(N_c)` relates the threshold to the stake coverage of
```
williamdemeo: The CIP bounds τ from below as well ("Must satisfy 0.5 < τ < σ(N_c)"), and unlike the upper bound, `½ < τ` is a property of the parameters alone, so it could live in `paramsWellFormed`; the disabled state doesn't need `τ ≡ 0`, since the periods do the disabling.  Deliberate?  Either imposing it or adding the reason it's omitted would spare the next reader the same question.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
```diff
@@ -151,17 +162,65 @@ record PParams : Type where
...
 
 ## Protocol Parameter Well Formedness
 
+The Leios parameters are deliberately absent from
+`positivePParams`{.AgdaFunction}, since governance must be able to reach the
+protocol's disabled state: `leiosHeaderPeriod`{.AgdaField} and
+`leiosVotingPeriod`{.AgdaField} must be able to reach zero — with no voting
+window no certificate can ever form — while `leiosDiffusionPeriod`{.AgdaField}
+disables the other way around, at the largest duration its wire format can
+carry (durations serialize as 32-bit unsigned integers, so `2^32 − 1`), which
```
williamdemeo: Is the 32-bit encoding citable?  CIP-164 carries no pparams wire format, and the type here is unbounded `ℕ`, so the parenthetical asserts a bound nothing in the CIP or this spec pins; if it comes from a cardano-ledger draft encoding, a reference would anchor it, and otherwise "at the largest duration its wire encoding can carry", without the number, keeps the claim safe.

A nuance on the mechanism: 2^32 − 1 ms is roughly fifty days, which a chain comfortably outlives, so what actually forecloses certificates is the direct-successor rule — with only the announcing block's immediate successor eligible to certify, any required gap beyond a realistic inter-block gap already disables.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
williamdemeo: Still present at 222456de3: the six fields remain in this list while the prose above keeps the promise that governance can reach the disabled state, so the disabling update is still rejected as ill-formed.  The suggestion above applies as-is, and it keeps the `ppWF` bridge justified: the eleven remaining fields are exactly `positivePParams`.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
williamdemeo: Dropping the alias resolves the record-vs-update inconsistency the other way, but it leaves the periods without a stated unit: `leiosHeaderPeriod : ℕ` plus "wall-clock time" doesn't say what one unit of ℕ is, and the certification window's eventual `⌈…/slotLength⌉` needs to know.  The CIP states the periods in seconds with sub-second precision, so the ℕ must carry some fixed-precision unit; one prose sentence naming it (or the alias back) would close this.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
williamdemeo: Unchanged at 222456de3: the paragraph still says `leiosMaxRefScriptSizePerEB` "has no CIP-164 row at all".  The suggested rewrite above still applies; if the `Milliseconds` alias stays out, drop that clause from it.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
ch1bo: I agree and thought I had dropped it. Doing it now.

src/Ledger/Dijkstra/Specification/PParams.lagda.md
ch1bo: Not done and dusted yet, but could we maybe specify that it should be at least `epochLength * slotLength` somehow?

PR #1308, opened 2026-09-03
[Dijkstra] Add starting account balance intervals

carlostome:
# Description

Following the changes in cardano-ledger (https://github.com/IntersectMBO/cardano-ledger/pull/6001) and the amendments to the CIP (https://github.com/cardano-foundation/CIPs/pull/1243)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/src/Ledger/Dijkstra/Foreign/Transaction.agda
+++ b/src/Ledger/Dijkstra/Foreign/Transaction.agda
@@ -136,6 +136,7 @@ record TxBodyTop : Type where
     txGuards                 : ℙ Credential
     txDirectDeposits         : DirectDeposits
     txBalanceIntervals       : AccountBalanceIntervals
+    txStartingBalanceIntervals : AccountBalanceIntervals
 
 instance
   convTxBodyTop : Convertible (TxBody TxLevelTop) TxBodyTop
--- a/src/Ledger/Dijkstra/Specification/Entities.lagda.md
+++ b/src/Ledger/Dijkstra/Specification/Entities.lagda.md
@@ -175,12 +175,13 @@ data _⊢_⇀⦇_,ENTITIES⦈_ : EntitiesEnv → CertState → TopLevelTx → Ce
     let refresh         = mapPartial (isGovVoterDRep ∘ voter) (fromList (ListOfGovVotesOf txTop))
         refreshedDReps  = mapValueRestricted (const (e + pp .drepActivity)) dReps refresh
 
-        withdrawalsSubTxs          = foldl (λ acc txSub → acc ∪⁺ WithdrawalsOf txSub) ∅ (SubTransactionsOf txTop)
-        withdrawals                = WithdrawalsOf txTop
-        withdrawalsCredentials     = mapˢ stake (dom withdrawals)
-        accountBalanceIntervals    = BalanceIntervalsOf txTop
-        directDeposits             = DirectDepositsOf txTop
-        directDepositsCredentials  = mapˢ stake (dom directDeposits)
+        withdrawalsSubTxs               = foldl (λ acc txSub → acc ∪⁺ WithdrawalsOf txSub) ∅ (SubTransactionsOf txTop)
+        withdrawals                     = WithdrawalsOf txTop
+        withdrawalsCredentials          = mapˢ stake (dom withdrawals)
+        accountBalanceIntervals         = BalanceIntervalsOf txTop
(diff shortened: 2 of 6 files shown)


src/Ledger/Dijkstra/Specification/Entities.lagda.md
williamdemeo: This paragraph is stale; it still says two fields instead of three.  I suggest replacing it with the following:

```suggestion
CIP-159 introduces three new transaction fields: `txDirectDeposits`{.AgdaField},
`txBalanceIntervals`{.AgdaField} and, at the top level only,
`txStartingBalanceIntervals`{.AgdaField}.  Direct deposits represent value that
flows from the transaction into account addresses.  The two interval fields let
a transaction assert bounds on account balances: `txBalanceIntervals`{.AgdaField}
against the balances this rule sees, and `txStartingBalanceIntervals`{.AgdaField}
against the balances at the start of the whole batch (`rewards₀`{.AgdaField}).
```

src/Ledger/Dijkstra/Foreign/Transaction.agda
```diff
@@ -136,6 +136,7 @@ record TxBodyTop : Type where
     txGuards                 : ℙ Credential
     txDirectDeposits         : DirectDeposits
     txBalanceIntervals       : AccountBalanceIntervals
+    txStartingBalanceIntervals : AccountBalanceIntervals
```
williamdemeo, line 139: It's annoying, but all of these fields should probably be realigned.

src/Ledger/Dijkstra/Specification/Entities/Properties/Computational.lagda.md
```diff
@@ -103,6 +103,7 @@ instance
                      withdrawals               = WithdrawalsOf txTop
                      withdrawalsCredentials    = mapˢ stake (dom withdrawals)
                      accountBalanceIntervals   = BalanceIntervalsOf txTop
+                     startingAccountBalanceIntervals = StartingBalanceIntervalsOf txTop
                      directDeposits            = DirectDepositsOf txTop
                      directDepositsCredentials = mapˢ stake (dom directDeposits)
```
williamdemeo, line 108: Annoying nit, but we should either align all of these `let`s, or none of them.  I don't have a strong preference, though most-but-not-all alignment is my least favorite.

williamdemeo (approved):
**Approved** with some minor, non-blocking comments.

**Question** (not a change request).  A non-empty `txBalanceIntervals` makes `UsesV4Features tx` hold (`hasBalanceIntervals` in `Utxow`), which makes `allowedLanguagesLegacy` exclude V1--V3 scripts, so a transaction carrying a non-empty `txBalanceIntervals` cannot run V1--V3 scripts.  Otoh, a non-empty `txStartingBalanceIntervals` has no such restriction, since it is not a `UsesV4Features` constructor.

Is that intentional?  The restriction to V4 for non-empty `txBalanceIntervals` seems reasonable.  But I don't see why the same restriction shouldn't hold for a non-empty `txStartingBalanceIntervals`.

**Follow-up suggestion**. This PR shows once again how brittle the sts rule patterns are.  In this case, in `ENTITIES`, one premise insertion touched five pattern sites across Computational and PoV, and the eleven-tuple is mostly underscores.

(In my recent properties PRs, I battled with the same issue **again**.)  

`Utxow` solves this with `genPremises` and a `⋯` pattern synonym (see Utxow.lagda.md:539-542).  A follow-up doing the same for `ENTITIES` and `SUBENTITIES` would let the PoV proofs name `certsStep` and `ddCreds⊆` instead of counting positions.  I think you mentioned you planned to tackle the `genPremises` problems soon.  Maybe this can be part of that work.

carlostome:
> **Follow-up suggestion**. This PR shows once again how brittle the sts rule patterns are. In this case, in `ENTITIES`, one premise insertion touched five pattern sites across Computational and PoV, and the eleven-tuple is mostly underscores.
> 
> (In my recent properties PRs, I battled with the same issue **again**.)
> 
> `Utxow` solves this with `genPremises` and a `⋯` pattern synonym (see Utxow.lagda.md:539-542). A follow-up doing the same for `ENTITIES` and `SUBENTITIES` would let the PoV proofs name `certsStep` and `ddCreds⊆` instead of counting positions. I think you mentioned you planned to tackle the `genPremises` problems soon. Maybe this can be part of that work.

I agree but atm `genPremises` doesn't work for `ENTITIES`.

PR #1316, opened 2026-09-08
[doc/infra] update `CONTRIBUTING`; add devShell for `formal-ledger-test` (#1315)

### LINKED ISSUE #1315: No documented devShell type-checks `formal-ledger-test`
## Problem

In the default dev shell every module under `formal-ledger-test` fails to load, while modules under `src` load normally.

```
$ nix develop
$ cd formal-ledger-test && agda src/Test/Prelude.lagda.md
error: [LibraryError]
Library 'formal-ledger' not found.
Add the path to its .agda-lib file to
  '/nix/store/…-libraries'
to install.
Installed libraries:
  standard-library-2.3
  standard-library-classes
  standard-library-meta
  abstract-set-theory
  iog-prelude
```

It is the same in Emacs, which is how this surfaced: `C-c C-l` on any file under `formal-ledger-test` reports the error above.

Thanks to Tudor Ferariu (@tferariu) for reporting this issue.

## Why

`formal-ledger-test/formal-ledger-test.agda-lib` declares `depend: … formal-ledger`, and the default shell registers only the five upstream libraries, not `formal-ledger` itself.  Two changes combined to produce this.

+  [#1175](https://github.com/IntersectMBO/formal-ledger-specifications/pull/1175) moved `src/Test/*` into the separate `formal-ledger-test` library. Before that the test modules were part of `formal-ledger`, and the default shell served them.
+  [#1177](https://github.com/IntersectMBO/formal-ledger-specifications/pull/1177) removed `formal-ledger-test` from `devShells.default`'s `inputsFrom`; that is right on its own (otherwise entering the default shell first waits for the fully type-checked ledger) but it left the test library with no shell.

Two things a contributor is likely to try cannot work, for reasons worth recording.

+  Adding a line to `~/.config/agda/libraries` has no effect.  The shell's `agda` is a wrapper that passes `--library-file=/nix/store/…-libraries` on every invocation, which overrides the user's own library file entirely.
+  Running `nix develop` from inside `formal-ledger-test` gives the identical shell, since Nix searches upward for `flake.nix` and the shell does not depend on the working directory.

## What to do

+  Expose a dev shell for the test library (`devShells.formal-ledger-test`).  Entering that environment already works by package attribute, since `formal-ledger-test` is a flake package and `nix develop` accepts one, but nothing advertises it: `nix flake show` lists only `default`, `fls-agda`, `fls-shake`, and `fls-shake-agdaWithPackages`.
+  Document the split in `CONTRIBUTING.md`, together with two facts that cost time to rediscover: Agda selects the governing `.agda-lib` by searching the current directory and its parents, so the test library must be type-checked from inside its own directory (from the repository root Agda picks the root library and reports `ModuleNameDoesntMatchFileName`); and a command-line `--library-file` overrides the wrapper's, which is how to check the test library against a working copy of `src` rather than against the built ledger.

## Adjacent staleness in CONTRIBUTING

These were found while diagnosing this.

+  The Agda pin moved to 2.8.0 in d8eb89d1a, a year ago; §Setup Without Nix still says "Install Agda 2.7.0.1", and the Emacs snippet still labels `~/ledger-agda` as 2.7.0.1.
+  §Using `update-alternatives` tells the reader to register `~/ledger-agda/bin/agda-mode`, which does not exist.  Our `agda` provides no `agda-mode`, and that executable is deprecated as of Agda 2.8.0 in favour of `agda --emacs-mode`.  An Emacs whose `agda2-mode` elisp comes from a different Agda then fails with `The Agda mode's version (X) does not match that of agda (2.8.0).`, and the guide says nothing about where matching elisp comes from.
+  §Setup Without Nix pins agda-stdlib at v2.2 (we use 2.3) and gives four commit hashes that all differ from `flake.lock`, referring the reader to a `build-tools/nix/sources.json` that no longer exists.
+  The same section type-checks `src/Everything.agda`, which does not exist.
+  `scripts/plot_typecheck_time.py` moved to `build-tools/scripts/plot_typecheck_time.py`.

## Not covered here

`TROUBLESHOOTING.md` is still pre-flake: it recommends `nix-shell -A devShells.ci` and `nix-build -A formalLedger`, neither of which exists.

williamdemeo:
## Description

Closes #1315.

### `flake.nix`: add dev shell for test library

Add `devShells.formal-ledger-test`, with an Agda that registers `formal-ledger` alongside the five upstream libraries, so modules under `formal-ledger-test` type-check.

The default shell is untouched; its derivation path is identical to master's.  The new shell joins `hydraJobs`, so its inputs stay in the binary cache.

`nix develop .#formal-ledger-test` already worked, since `formal-ledger-test` is a flake package and `nix develop` accepts one.  Naming it as a devShell makes it visible in `nix flake show` and gives the guide something stable to point at.

### `CONTRIBUTING`: document Agda libraries and refresh stale facts

A new section, "Agda libraries", states which shell serves which library, that Agda finds the governing `.agda-lib` by searching the current directory and its parents (so the test library must be type-checked from inside `formal-ledger-test`), and how to type-check the test library against a working copy of `src` rather than against the pre-built ledger.

The test library shell is listed beside the default one in the "Development Environment Setup" section, and the Emacs instructions point at both.

The five stale items named in #1315 are fixed, as follows:

+  Agda 2.7.0.1 becomes 2.8.0, in the "Setup Without Nix" section and in the Emacs snippet.
+  The `agda-mode` guidance becomes `agda --emacs-mode {locate,setup,compile}`, which replaces the executable that Agda 2.8.0 deprecated and which our `agda` supports.  A new Emacs step 1 explains the version-mismatch error and where matching elisp comes from.
+  §Setup Without Nix reads the pinned revisions out of `flake.lock` instead of listing hashes that drift, and agda-stdlib is v2.3.  The dead `sources.json` link definition is gone.
+  The non-Nix type-check target is `src/Ledger.lagda.md`, not `src/Everything.agda`.
+  `plot_typecheck_time.py` is under `build-tools/scripts`.

### Verification

+  `nix develop .#formal-ledger-test`, then `cd formal-ledger-test && agda src/Test/Prelude.lagda.md`: succeeds in about 20 s, after a one-time 127 MiB fetch.  One `agda` on `PATH`, and its library file lists the built `formal-ledger`.
+  `nix eval .#devShells.x86_64-linux.default.drvPath` gives the same path on this branch and on master.  The new shell also evaluates for `aarch64-darwin`, and `hydraJobs.devShells` gains `formal-ledger-test`.
+  Modules under `src` still resolve in the new shell, so it serves work spanning both libraries.
+  Every command added to the guide was run as written, in both shells.
+  All reference links and in-page anchors in `CONTRIBUTING.md` resolve.  The file is not part of the mkdocs site.
+  `scan_properties.py --check` passes with the dashboard up to date.

`TROUBLESHOOTING.md` is still pre-flake, as #1315 notes; that is out of scope here.

## Checklist

- [x] Self-reviewed the diff


flake.nix
```diff
@@ -126,6 +126,13 @@
                   ];
                 });
 
+            # The test library is a separate Agda package that depends on
+            # `formal-ledger`, which the default shell does not register.  This
+            # shell does, so modules under `formal-ledger-test` type-check.
```
carlostome:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -590,17 +689,33 @@ prefer not to use Nix.
...
+    jq -r '.nodes | to_entries[]
+           | select(.value.locked.type == "github")
+           | "\(.key): \(.value.locked.owner)/\(.value.locked.repo) \(.value.locked.rev)"' \
+      flake.lock
+    ```
+
+    The `abstract-set-theory`, `iog-prelude`, `standard-library-classes`, and
+    `standard-library-meta` lines are the four libraries you need to clone.  The
+    standard library itself comes from the pinned `nixpkgs` rather than from a flake
+    input of its own; it is version 2.3 as of this writing.
```
carlostome: This section is about titled "Setup without Nix", thus,
```suggestion
```

CONTRIBUTING.md
```diff
@@ -755,19 +873,21 @@ available, on your system. For example,
 sudo update-alternatives --install /usr/bin/agda agda ~/.cabal/bin/agda-2.8.0 10
 ```
 
-Install the associated version of `agda-mode`, which is required for using Agda
-(versions < 2.8.0) in Emacs.
+The separate `agda-mode` executable is deprecated as of Agda 2.8.0, and
+`~/ledger-agda/bin` contains only `agda`, so there is nothing to register alongside it;
+use `agda --emacs-mode` instead, as described in step 1 of the [Emacs](#emacs)
+instructions above.  If you also manage an Agda older than 2.8.0, register its
+`agda-mode` as well, so that you can select the pair together.
```
carlostome: I don't think we support older versions than 2.8.0 so I'd remove this paragraph
```suggestion
```

CONTRIBUTING.md
```diff
@@ -755,19 +873,21 @@ available, on your system. For example,
...
 
 ```bash
-sudo update-alternatives --install /usr/bin/agda-mode agda-mode ~/ledger-agda/bin/agda-mode 1
+sudo update-alternatives --install /usr/bin/agda-mode agda-mode ~/.cabal/bin/agda-mode-2.7.0.1 10
```
carlostome: ditto
```suggestion
```

CONTRIBUTING.md
```diff
@@ -755,19 +873,21 @@ available, on your system. For example,
...
 ```
 
-Choose which Agda version you want to use; if that version below 2.8.0, select the
-appropriate `agda-mode` version to accompany it!
+Then choose which Agda version you want to use.
 
 ```bash
 sudo update-alternatives --config agda
-sudo update-alternatives --config agda-mode
+sudo update-alternatives --config agda-mode   # only if you registered an older agda-mode
```
carlostome: ditto
```suggestion
```

CONTRIBUTING.md
```diff
@@ -399,6 +438,66 @@ a module without staying in the shell with the command
...
+The main specification and test suite declare two Agda libraries, and each has its own development shell, as
+follows:
+
++  `formal-ledger`, declared by `formal-ledger.agda-lib` in the repository root, whose
+   sources are `src` and `src-lib-exts`.  The default shell serves this library.
++  `formal-ledger-test`, declared by `formal-ledger-test/formal-ledger-test.agda-lib`,
+   whose sources are under `formal-ledger-test/src`.  This library depends on
+   `formal-ledger`, which the default shell does not register, so a module under
+   `formal-ledger-test` fails there with
+
```
carlostome, line 452: I feel a version of this bit should be part of `Project Overview` instead of a standalone subsection. Also it contains repeated information about the nix-shells.

CONTRIBUTING.md
```diff
@@ -399,6 +438,66 @@ a module without staying in the shell with the command
...
+   Library 'formal-ledger' not found.
+   ```
+
+Type-check the test library from that package's own environment, as follows:
+
+```bash
+nix develop .#formal-ledger-test
+cd formal-ledger-test
+agda src/Test/Prelude.lagda.md
+```
```
carlostome, line 464: Repeated from above.

CONTRIBUTING.md
```diff
@@ -399,6 +438,66 @@ a module without staying in the shell with the command
...
+
+Entering that shell the first time fetches the type-checked ledger from the binary cache
+(about 130 MiB), after which a single test module type-checks in seconds.  The shell
+resolves modules under `src` as well, so it is the one to launch an editor from when
+your work spans both libraries.  Two things to keep in mind: the ledger it registers is
+the *built* library, so local edits under `src` are invisible to it; and it provides
+neither `fls-shake` nor `mkdocs`, so keep the default shell for build-tool work.
+Type-check the whole test library with `nix build .#formal-ledger-test`.
```
carlostome, line 472: Maybe this bit should be a comment when the `formal-ledger-test` nix devShell is introduced.

CONTRIBUTING.md
```diff
@@ -399,6 +438,66 @@ a module without staying in the shell with the command
...
+agda --library-file="$FLS_LIBS" src/Test/Prelude.lagda.md
+```
+
+A `--library-file` given on the command line overrides the one the shell's `agda`
+wrapper supplies; that hardcoded wrapper flag is also why adding a line to
+`~/.config/agda/libraries` has no effect in these shells.  In Emacs, pass the same flag
+with `(setq agda2-program-args '("--library-file" "/path/to/fls-libraries"))`.  Expect a
+long first run, since the ledger is then type-checked from source rather than taken from
+the cache.
+
```
carlostome, line 500: I'd keep this part out or move it to workflows as an example.

CONTRIBUTING.md
```diff
@@ -399,6 +438,66 @@ a module without staying in the shell with the command
...
+the *built* library, so local edits under `src` are invisible to it; and it provides
+neither `fls-shake` nor `mkdocs`, so keep the default shell for build-tool work.
+Type-check the whole test library with `nix build .#formal-ledger-test`.
+
+**The current directory matters**.  Agda finds the `.agda-lib` file that governs a
+module by searching the current directory and its parents, not the directory of the file
+named on the command line.  Type-check the test library from inside
+`formal-ledger-test`, as above; from the repository root the *root* library governs, and
+Agda reports `ModuleNameDoesntMatchFileName` instead.  Emacs is unaffected, since
+`agda-mode` runs Agda in the directory of the file being loaded.
```
carlostome, line 479: This is not particular to the formal-ledger-test so I'd remove it.

PR #1319, opened 2026-09-10
Close 1314 by renaming type in examples

### LINKED ISSUE #1314: Renaming formal-ledger-test example contract Types
Rewriting the example smart contracts and their traces so that:
`Label` -> `Datum`
`Input` -> `Redeemer`
This more clearly matches the expected names for those types and reflects the most recent version of my work that references them.

tferariu:
# Description

Performed the desired renaming. All changes are confined to the formal-ledget-test folder, so they should be self-contained and not affect anything else in the repo.

# Checklist

- [x] Self-reviewed the diff


formal-ledger-test/src/Test/Examples/MultiSigV2/Validator.lagda.md
```diff
@@ -187,26 +187,26 @@ notTooLate par deadline (txinfo , _) = maybe (λ now →  ⌊ (_+_ {{addNat}} no
...
+      geq (length sigs) (minNumSignatures param) && continuing ctx && (case (newDatum ctx) of λ where
         (just Holding) -> (checkPayment pkh v ctx) && (oldValue ctx == (maybeMap (_+_ {{addValue}} v) (newValue ctx)))
         _ -> false)
     ((Collecting v pkh d sigs) , Cancel) ->
       (newValue ctx == oldValue ctx) && continuing ctx &&
-      (case (newLabel ctx) of λ where
+      (case (newDatum ctx) of λ where
         (just Holding) -> expired d ctx 
         _ -> false) 
     (Holding , Cleanup) -> gt minValue (fromMaybe 0 (oldValue ctx)) && not (continuing ctx)
```
williamdemeo, line 212:
```suggestion
    (Holding , Stop) -> gt minValue (fromMaybe 0 (oldValue ctx)) && not (continuing ctx)
```
`Cleanup` is no longer a constructor of `Redeemer` (it became `Stop` in `MultiSigV2/Datum.lagda.md`), so this is not a constructor pattern any more: Agda reads it as a fresh pattern variable.  The clause is now effectively `(Holding , _) -> …`, which silently widens the validator rather than failing to compile.

Copilot flagged this line but gave the wrong reason; it says the module "cannot type-check."  It does type-check.  But it should be fixed despite a green CI pass.

What actually changed: with datum `Holding`, the redeemers `Add sig`, `Pay` and `Cancel` used to fall through to `_ -> false`.  They now take this clause and return `gt minValue (fromMaybe 0 (oldValue ctx)) && not (continuing ctx)`, so the contract can be terminated through three redeemers that should reject.

I checked this by appending the following to the module:

```agda
probeA : (p : Params) (ctx : ScriptContext) →
  agdaValidator p Holding Pay ctx ≡ (gt minValue (fromMaybe 0 (oldValue ctx)) && not (continuing ctx))
probeA p ctx = refl
```

which typechecks, and the following statement (which *ought* to hold) fails, with `agdaValidator p Holding Pay ctx != false of type Bool`:

```agda
probeB : (p : Params) (ctx : ScriptContext) → agdaValidator p Holding Pay ctx ≡ false
probeB p ctx = refl
```

If you make the suggested change, `Cleanup` -> `Stop`, then `agda src/Test/Examples.lagda.md` exits 0 and all five trace assertions in `MultiSigV2/Test/Trace.lagda.md` still hold (verified).

formal-ledger-test/src/Test/Examples/AccountSim/OffChain/Stop.lagda.md
williamdemeo, line 3:
```suggestion
source_path: src/Test/Examples/AccountSim/OffChain/Stop.lagda.md
```

formal-ledger-test/src/Test/Examples/DEx/OffChain/Stop.lagda.md
williamdemeo, line 3:
```suggestion
source_path: src/Test/Examples/DEx/OffChain/Stop.lagda.md
```

formal-ledger-test/src/Test/Examples/MultiSigV2/OffChain/Stop.lagda.md
williamdemeo, line 3:
```suggestion
source_path: src/Test/Examples/MultiSigV2/OffChain/Stop.lagda.md
```

williamdemeo (changes requested):
Nice work, Tudor.  Just a few change requests, noted in-line.

Plus, here are a few minor notes that you can take or leave:

+  `SCDatum` is used in only four of the 23 modules that introduce it, namely the four `Validator.lagda.md` files.  In the other 19 the `renaming` is doing a `hiding`'s job, which works but says less than it could.  `hiding (Datum; Redeemer)` states the intent and avoids introducing an alias.  I checked that the substitution compiles both where `Datum` is used and where it is not.  That gives one clause for the four validators and one for everything else.  In the validators the `renaming` is genuinely needed, and `Data` is not a shortcut around it.

+  Renaming `getLabel` to `getDatum` collides with `Ledger.Conway.Specification.Script.Validation.getDatum`, a different
   function: `Tx → UTxO → ScriptPurpose → Maybe Datum` against the example's `TxOut → Maybe Datum`.

   It is a hard clash at the definition site, not merely on use; restoring that import to `MultiSigV2/OffChain/Lib.lagda.md` gives `[ClashingDefinition] Multiple definitions of getDatum` (checked).

   That is why this PR has to drop `open import Ledger.Conway.Specification.Script.Validation …` from five MultiSigV2 off-chain modules (`AddSig`, `Cancel`, `Lib`, `Pay`, `Propose`).  The other three contracts escape only because their off-chain modules never imported it, and all four `Lib.lagda.md` files are now permanently unable to.

   Worth a name that does not collide, so the import can stay and so two different `getDatum`s are not in play while reading.  `txOutToDatum` has the merit of already being the spec's own name for exactly this function (`src/Ledger/Dijkstra/Specification/Transaction.lagda.md:588`).

+  Three small things about the PR description:

   1.  "formal-ledget-test"  ->  "formal-ledger-test".
   2.  "Performed the desired renaming" undersells the PR.  Beyond the `Label`/`Input` rename that [#1314] asks for, this PR

       + renames the terminal action in three contracts (`Cleanup` and `Close` both become `Stop`),
       + renames three modules, and
       + drops five imports.

       Those are good changes; add these bullet points to the PR description.

   3.  Add a `Closes #1314` line to the description so merging the PR will close the issue.


+  Naming symmetry: `AccountSim` and `DEx` now pair `Start.lagda.md`/`startTx` with `Stop.lagda.md`/`makeStopTx`, while MultiSigV2 pairs `Open.lagda.md`/`openTx`/`openContract` with the new `Stop`.  If the point of the PR is a shared vocabulary, `Start`/`Stop` in MultiSigV2 would finish the job.

